PR #6, opened 2021-03-10
Remark input

hbulgarini:
Remark.ts is not the definitive component per se , is kind of testing component for sending remarks.
My idea is to have a generic component that handles "actions" and then different components/hooks/functions that will handle the specific use case logic. 

Known issue:
Messages from millau to rialto are not currently working because of Bad Signature issues. I suspect this comes because of the custom hasher (H512) needs to be set up  for millau.

All statements like `sourceApi.registry.createType` were ts ignored because TS parser is not recognizing the custom types:
https://polkadot.js.org/docs/api/start/typescript.user


README.md
```diff
@@ -1,6 +1,6 @@
 # UI Bridge for Substrate chains.
 
-This is a UI proposal for querying , operating  and testing the bridge.
+This is a UI proposal for querying , operating and testing the bridge.
```
tomusdrw, line 3:
```suggestion
This is a UI proposal for querying, operating and testing the bridge.
```

README.md
```diff
@@ -13,6 +13,25 @@ This will install all the dependencies for the project.
...
+# In case these values are not provided the default chains will be rialto and millau chains.
+
+### `yarn test`
+
+Launches the test runner in the interactive watch mode.\
+See the section about [running tests](https://facebook.github.io/create-react-app/docs/running-tests) for more information.
+
+### Configure custom substrate providers
+
+In the `.env` file at root project directory, you
```
tomusdrw, line 34:
```suggestion
```

src/components/Remark.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  const onChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setRemarkInput(event.target.value);
+  };
+
+  async function sendMessageRemark() {
+    const keyring = new Keyring({ type: 'sr25519' });
+    const account = keyring.addFromUri('//Alice');
+
+    const remarkCall = await targetApi.tx.system.remark(remarkInput);
```
tomusdrw, line 34: The expectation seems to be that `remarkInput` is a `0x`-prefixed hex bytes string, what if it's not correct? Will it fail here or be converted?

src/components/Remark.tsx
```diff
@@ -0,0 +1,91 @@
...
+      weight
+    };
+
+    // Ignoring custom types missed for TS for now.
+    // Need to apply: https://polkadot.js.org/docs/api/start/typescript.user
+    // @ts-ignore
+    const payloadType = sourceApi.registry.createType('OutboundPayload', payload);
+    // @ts-ignore
+    const messageFeeType = sourceApi.registry.createType('MessageFeeData', {
+      lane_id: '0x00000000',
```
tomusdrw, line 56: Please move `lane_id` to some global constant instead of repeating it.

src/components/Remark.tsx
```diff
@@ -0,0 +1,91 @@
...
+      payload: u8aToHex(payloadType.toU8a())
+    });
+
+    const estimatedFeeCall = await sourceApi.rpc.state.call<Codec>(
```
tomusdrw, line 60: I'd enclose the entire `estimate_message_delivery_and_dispatch_fee` workflow into a separate function, cause it most likely is going to be useful outside of `Remark` component.

src/components/Remark.tsx
```diff
@@ -0,0 +1,91 @@
...
+    );
+
+    // @ts-ignore
+    const estimatedFeeType = sourceApi.registry.createType('Option<Balance>', estimatedFeeCall);
+    const estimatedFee = estimatedFeeType.toString();
+
+    const bridgeMessage = sourceApi.tx[`bridge${targetChain}MessageLane`].sendMessage(lane_id, payload, estimatedFee);
+
+    const nonceCall = await sourceApi.rpc.system.accountNextIndex(account.address);
+    const nonce = nonceCall.toNumber();
```
tomusdrw, line 72: Why do you query the nonce manually? That's something that the api will do for you, no?

src/components/Remark.tsx
```diff
@@ -0,0 +1,91 @@
...
+    await bridgeMessage.signAndSend(account, { nonce });
+  }
+
+  if (!areApiLoading) {
+    return null;
```
tomusdrw, line 77: This doesn't make much sense. "If the API is NOT loading display nothing". I'm assuming  it works, so the variable name should be changed. Also `useLoadingApi` could get a rename to make it easy to understand which value you are getting (is it loading or is it NOT loading).

src/components/Remark.tsx
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  return (
+    <Container className={className}>
+      <Input onChange={onChange} value={remarkInput} />
+      <Button onClick={sendMessageRemark}>Message Remark</Button>
```
tomusdrw, line 83:
```suggestion
      <Button onClick={sendMessageRemark}>Send Remark</Button>
```

README.md
hbulgarini: i don't understand what changes you are proposing here ... diff looks quite similar

README.md
hbulgarini: why do you want to remove this?

README.md
tomusdrw: the space before comma.

README.md
tomusdrw: It's duplicated here, two paragraphs up you have exactly the copy and this part is simply unfinished.

src/components/Remark.tsx
hbulgarini: actually i think that if it's not a `0x` prefixed hex , the input will be consider as ascii right? the call does not fail if you provide a not 0x fyi

src/components/Remark.tsx
hbulgarini: this is why i mentioned that remark.tsx is kind of testing component, all those functions will not remain in the component logic. My plan is to have the end to end working as it is now and then abstract the code accordingly.

src/components/Remark.tsx
hbulgarini: not really, i had to specify the nonce in the call otherwise it does not get refreshed ( we discussed this in the chat )

src/components/Remark.tsx
hbulgarini: I understand that the UX is not the best one, because i was looking for a quick hack to prevent the message call to be executed if the both source and target connection are not ready.  As i mentioned this is just to test the api calls and set up ... for sure UX will not have nothing to do with the current behavior.

Anyway ... i'm not getting what you are referring when you say : ` could get a rename to make it easy to understand which value you are getting)` ... i think the name `areApiLoading` is quite descriptive isn't it?

src/components/Remark.tsx
hbulgarini: btw ... would you prefer to set LaneId as part of a variable in the .env file?

src/components/Remark.tsx
tomusdrw: Yeah, that was my expectation as well, but just wanted to confirm that.

src/components/Remark.tsx
tomusdrw: I think the safest way is that it should be coming from a hook (like `useLoadingApi`). Initially it will be constant but in the future we might want to introduce some selection.

src/components/Remark.tsx
tomusdrw: This doesn't solve the "refreshing" problem really. What I meant is that you need to wait for submission of previous transaction before submitting a new one so that two paralel calls to `accountNextIndex` don't end up returning the same nonce.

src/components/Remark.tsx
tomusdrw: I'm not talking about UX, only about the code and variable names. You have a variable called `areApiLoading` which I presume should be `true` if the API is loading and is `false` if the API is already connected.

In this condition you are flipping the value, so in the case of `areApiLoading == false` you are returning `null` from render. That's what doesn't make sense to me - the variables suggest that the API is no longer loading, so we can safely display something, while you are doing the opposite.

Hence I think the variable names should be renamed to explain what they are for to avoid confusion.

hbulgarini: ahhh yes ... sorry. I know what you mean , originally the the logic was right and after a change it got mixed. You are right

hbulgarini:
@tomusdrw  i think that all the comments were address/replied

src/components/Remark.tsx
```diff
@@ -67,20 +73,20 @@ const Remark = ({ className, targetChain }: Props) => {
     const estimatedFee = estimatedFeeType.toString();
 
     const bridgeMessage = sourceApi.tx[`bridge${targetChain}MessageLane`].sendMessage(lane_id, payload, estimatedFee);
-
-    const nonceCall = await sourceApi.rpc.system.accountNextIndex(account.address);
-    const nonce = nonceCall.toNumber();
-    await bridgeMessage.signAndSend(account, { nonce });
+    await bridgeMessage.signAndSend(account, { nonce: -1 });
+    setIsExecuting(false);
```
tomusdrw, line 77: `try-catch-finally`?

tomusdrw (approved):
:ok_hand: Awesome!

src/components/Remark.tsx
hbulgarini: oh i forgot this one... committing it soon

PR #8, opened 2021-03-17
Accounts selector + keyring + derived Account.

hbulgarini:
This PR contains:

- Implemented Keyring so now accounts are read from the extension + test accounts from the chain in case .env variable is set.
- Account selector: When an account is selected the account gets stored in the global state provided by a hook.
- DerivedAccount function and hook: Derives the account from the current account selected provided by a hook as well.
- Remark test component update: now it executes the remark using the account selected.

Disclaimer: Similar to previous PR, there are some UX basic behaviors not implemented on remark component ... for example you can trigger a remark without an account selected making the process to fail. All these will be implemented in next PR when transfer component is ready and i will have the necessary experience to abstract components/functions.
Disclaimer: All code related to custom hooks / reducers / context can be considered as definitive .


hbulgarini:
@tomusdrw , ready to review. 
Issue #11  is fixed in commit #fa7e129

src/components/Accounts.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+interface Props {
+  className?: string;
+}
+
+const formatOptions = (accounts: Array<KeyringPair>) =>
+  accounts.map(({ meta, address }) => ({
+    icon: 'user',
+    key: address,
+    text: (meta.name as string).toLocaleUpperCase(),
```
tomusdrw, line 36: I guess uppercasing is what PJS apps is doing as well, right? IF that's the standard already, let's do that but IMHO we shouldn't be formatting how users named their accounts.

src/components/Accounts.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const { dispatchAccount } = useUpdateAccountContext();
+  const { sourceChain, targetChain } = useSourceTarget();
+  const { account: currentAccount } = useAccountContext();
+
+  const value = currentAccount?.address || '';
+
+  const onChange = (evt: any, { value }: any) => {
+    const account = accounts.find(({ address }) => address === value);
+    if (account) {
+      dispatchAccount({ payload: { account }, type: AccountActions.SET_ACCOUNT });
```
tomusdrw, line 52: Is it really necessary to complicate this so much (i.e. payload & action)? Why not simply have `setAccount` function coming from `useAccounts`?

src/components/Accounts.tsx
```diff
@@ -0,0 +1,76 @@
...
+    text: (meta.name as string).toLocaleUpperCase(),
+    value: address
+  }));
+
+const Accounts = ({ className }: Props) => {
+  const accounts = useAccounts();
+  const derivedAccount = useDerivedAccount();
+  const { dispatchAccount } = useUpdateAccountContext();
+  const { sourceChain, targetChain } = useSourceTarget();
+  const { account: currentAccount } = useAccountContext();
```
tomusdrw, line 45: I'm not a fan of 4 seemingly separate hooks that are basically super tightly coupled. Why can't we simply have:
```
const { accounts, setCurrentAccount, currentAccount, derivedAccount } = useAccounts();
```
?

If you don't want to bundle all things together then let's at least make this similar to the state hooks and merge getter and a function to alter the state.
```
const [account, setAccount] = useCurrentAccount();
```

src/components/Accounts.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <p>{sourceChain}</p>
+      <Dropdown
+        value={value}
+        onChange={onChange}
+        placeholder="Select Account"
+        fluid
+        selection
+        options={formatOptions(accounts)}
+      />
+      {derivedAccount && (
```
tomusdrw, line 67: This will cause UI elements to move around after `derivedAccount` is set. I think the `<p>` element should always be there and we should just be changing the content conditionally.

src/components/Remark.tsx
```diff
@@ -39,6 +44,7 @@ const Remark = ({ className, targetChain }: Props) => {
   const [remarkInput, setRemarkInput] = useState('0x');
   const lane_id = useLaneId();
   const [executionStatus, setExecutionStatus] = useState('');
+  const { account: currentAccount } = useAccountContext();
```
tomusdrw, line 47:
```suggestion
  const { account } = useAccountContext();
```

src/components/Remark.tsx
```diff
@@ -52,13 +58,14 @@ const Remark = ({ className, targetChain }: Props) => {
     setIsExecuting(true);
 
     try {
-      const keyring = new Keyring({ type: 'sr25519' });
-      const account = keyring.addFromUri('//Alice');
-
+      if (!currentAccount) {
+        return;
+      }
+      const account = currentAccount;
```
tomusdrw, line 64:
```suggestion
      if (!account) {
        return;
      }
```

src/contexts/AccountContextProvider.tsx
```diff
@@ -0,0 +1,56 @@
...
+//
+// Parity Bridges UI is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+
+//copied over from @substrate/context This needs to be updated.
```
tomusdrw, line 17: Which file is it copied from exactly? We should make sure that the license is not violated here.

src/contexts/KeyringContextProvider.tsx
```diff
@@ -0,0 +1,106 @@
...
+//
+// Parity Bridges UI is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+
+//copied over from @substrate/context This needs to be updated.
```
tomusdrw, line 17: same here.

src/contexts/KeyringContextProvider.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+export function KeyringContextProvider(props: KeyringContextProviderProps): React.ReactElement {
+  const { children = null } = props;
+  const { sourceChain } = useSourceTarget();
+  const { ss58Formats } = chainsConfigs;
+  const sourceChainSS58Format = ss58Formats[sourceChain];
+  const initState = { ...INIT_STATE };
+  const [state, dispatch] = useReducer(keyringReducer, initState);
+  const [keyringPairs, setKeyringPairs] = useState<Array<KeyringPair>>([]);
+  const [keyringPairsReady, setkeyringPairsReady] = useState(false);
```
tomusdrw, line 53: Why do we mix both the `useReducer` and `useState` here? It seems that information stored in both is tightly coupled. I think we should settle on either `useState` if the state is simple (and I think it is) or `useReducer` in case the state management is complicated (I think it isn't since we basically have one value & it's readiness).

src/util/substrateProviders.ts
```diff
@@ -52,8 +65,32 @@ export const customTypes: CustomTypes = {
...
 
+export const ss58Formats: GenericKeyNumberConfigs = {
+  [CHAIN_1]: 48,
+  [CHAIN_2]: 60
+};
+
+export const accountDerivations: GenericKeyStringConfigs = {
+  [CHAIN_1]: 'pallet-bridge/account-derivation/account',
+  [CHAIN_2]: 'pallet-bridge/account-derivation/account'
+};
```
tomusdrw, line 85: This is going to be always constant, so no need to have per-chain configuration.

src/util/substrateProviders.ts
```diff
@@ -52,8 +65,32 @@ export const customTypes: CustomTypes = {
   [CHAIN_2]: types[CHAIN_2]
 };
 
+export const bridgeIds: GenericKeyStringConfigs = {
+  [CHAIN_1]: 'rlto',
```
tomusdrw, line 69: This doesn't look very nice with `CHAIN_1` & `CHAIN_2` environment varaiables. It's assuming `CHAIN_1` is Rialto anyway, so it doesn't make sense to pretend it can be generic.

src/util/substrateProviders.ts
```diff
@@ -52,8 +65,32 @@ export const customTypes: CustomTypes = {
...
+};
+
 export const customHashers: CustomHashers = {
   [CHAIN_2]: hasherH512
 };
 
+export const ss58Formats: GenericKeyNumberConfigs = {
+  [CHAIN_1]: 48,
+  [CHAIN_2]: 60
+};
```
tomusdrw, line 80: I'm not sure if this is the most convenient way of providing chain-specific configuration, why not group all parameters per chain like so:
```
export const knownChains = {
  'Rialto': {
     ss58Prefix: 48,
     customHasher: null,
     bridgeId: 'rlto',
     customTypes: types.Rialto,
     provider: 'wss://wss.rialto.brucke.link',
  },
  'Millau': {
     ...
  },
};
```

We can introduce parametrising (via env-variables) this later on when we know all of the parameters, because now it becomes unmaintainable.

src/components/Accounts.tsx
hbulgarini: The reason why i'm using a global variable for account is because the account object is ( and will be ) used in a lot of components across the UI. If i just rely on a local state for that , then i would have to pass the selected account in a props chain hell.

src/components/Accounts.tsx
hbulgarini: it's the same for me, i formatted the accounts in uppercase because the substrate template project was doing that as well.  I can easily remove it if you prefer to

src/components/Remark.tsx
hbulgarini: This was addressed already in a new PR... but yes ... that does not makes sense.

src/contexts/AccountContextProvider.tsx
hbulgarini: hmm that comment should not be there... i will remove it

src/util/substrateProviders.ts
hbulgarini: cool ...

src/contexts/KeyringContextProvider.tsx
hbulgarini: yes, i think the keyring reducer can be removed here. Actually my plan was to use this reducer as the account management reducer in general but i preferred to use the Account reducer that makes the keyring reducer to not have too much sense since it is just handling connection status updates that can live in a local state.

src/components/Accounts.tsx
hbulgarini: ok, that would add another hook to collect them all , but i see your point.

src/components/Accounts.tsx
hbulgarini: you are right ... to be honest i was not focus on these things at all since i just wanted to make sure that the derive account is the right one once the account gets selected.

src/components/Remark.tsx
hbulgarini: This was addressed already in a new PR... but yes ... that does not makes sense.

src/components/Accounts.tsx
tomusdrw: No, let's keep it like that. I think it's better if it's uniform.

src/components/Accounts.tsx
tomusdrw: You're missing the point here. I'm fine with using context for storing this, don't want this do be pushed through the components.
My point was that I think that the reducer API (and action construction) is pretty low-level API and shouldn't be used in the high-level component (`Accounts.tsx`). What I proposed in some other comment is to create account-related hook and only expose `setCurrentAccount` function - it's fine to use the reducer under the hood, but the point is to expose simple API for high-level components.

src/components/Accounts.tsx
tomusdrw: But why another hook? This is account-related funcitonality, I think it can easily stay together in one file without getting overwhelmingly big? But even one more hook is better than too fine-grained and low-level API expose to high-level components IMHO.

src/components/Accounts.tsx
tomusdrw: Sure, but it's really a trivial change :)

src/contexts/KeyringContextProvider.tsx
```diff
@@ -67,40 +55,32 @@ export function KeyringContextProvider(props: KeyringContextProviderProps): Reac
...
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [state]);
+  }, [keyringStatus, loadAccounts]);
 
   useEffect(() => {
-    const { keyringStatus } = state;
     if (keyringStatus === KeyringStatuses.READY) {
       const keyringOptions = keyring.getPairs();
+      console.log('keyringOptions', keyringOptions);
```
tomusdrw, line 79: stray `console.log`

src/components/Accounts.tsx
```diff
@@ -38,19 +35,13 @@ const formatOptions = (accounts: Array<KeyringPair>) =>
...
 
-  const value = currentAccount?.address || '';
+  const value = account?.address || '';
 
   const onChange = (evt: any, { value }: any) => {
-    const account = accounts.find(({ address }) => address === value);
-    if (account) {
-      dispatchAccount({ payload: { account }, type: AccountActions.SET_ACCOUNT });
-    }
+    onChangeAccount(value);
```
tomusdrw, line 44: IMHO this should be called `setAccount` or `changeAccount`. `on*` prefix suggests that this is an event handler and not something that triggers the action.

src/configs/substrateProviders.ts
```diff
@@ -0,0 +1,76 @@
...
+  if (envPrefix) {
+    return parseInt(envPrefix);
+  }
+  return defaultValue;
+};
+
+export const chainsConfigs: ChainConfigs = {
+  Millau: {
+    SS58Format: getSS58Format(process.env.REACT_APP_MILLAU_SS58_PREFIX, 60),
+    bridgeId: process.env.REACT_APP_MILLAU_BRIDGE_ID || 'mlau',
```
tomusdrw, line 65: I think `.env.example` might need an update as well.

tomusdrw (approved):
Awesome! Couple of other tiny things, but after fixing ready to be merged.

Issue #13, opened 2021-03-18
Missing LICENSE file


HCastano:
The code has a header which refers to a `LICENSE` file which I haven't been able to find.

We should also double check what license to use for this repo (is Apache fine, or should we be more conservative and go with GPL3 for example).

tomusdrw:
Good catch, @hbulgarini  please switch to GPL3 - that's the default license we use, I see no reason to change this to Apache.

hbulgarini:
Actually that header was inherited from the previous bridge-ui project. As the header mentioned paritytech i thought that was the right license. I will switch to use the same license header as the common repository

hbulgarini:
closed #14

PR #14, opened 2021-03-18
License change

hbulgarini:
In regards to the request in #13

### DIFF
--- a/src/util/useDidUpdateEffect.ts
+++ b/src/util/useDidUpdateEffect.ts
@@ -1,4 +1,18 @@
-/* eslint-disable header/header */
+// Copyright 2021 Parity Technologies (UK) Ltd.
+// This file is part of Parity Bridges UI.
+//
+// Parity Bridges UI is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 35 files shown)


src/util/useDidUpdateEffect.ts
```diff
@@ -1,9 +1,5 @@
 /* eslint-disable header/header */
```
HCastano, line 1: This is missing a header

HCastano (approved):
You're missing a top level `LICENSE` file. Add one before merging (you can just copy the one in the [bridges repo](https://github.com/paritytech/parity-bridges-common/blob/master/LICENSE)).

src/util/useDidUpdateEffect.ts
hbulgarini: Actually Thibaut disabled the header on purpose since that was taken from stackoverflow ( he copied the link ). Do you think that i should put the header anyway ?

hbulgarini:
> You're missing a top level `LICENSE` file. Add one before merging (you can just copy the one in the [bridges repo](https://github.com/paritytech/parity-bridges-common/blob/master/LICENSE)).

Thanks! It was added.

PR #23, opened 2021-03-24
Update README.md

tomusdrw:
[Rendered Version](https://github.com/paritytech/parity-bridges-ui/blob/82f844a26b0475e5027272f49362ab9341d22e74/README.md)


hbulgarini:
I think we should add more env variables that got in since i wrote this readme file.

hbulgarini (approved):

Issue #12, opened 2021-03-18
Investigate node's abnormal closure connect issues.


hbulgarini:
In the developer console after some time the following errors start appearing:

index.js:1 2021-03-18 13:10:15          API-WS: disconnected from ws://127.0.0.1:19944: 1006:: Abnormal Closure

PR #28, opened 2021-03-24
Swap chains bug - disconnection issues.

hbulgarini:
Solves issue in #25. 
Possibly solves issue described in #12

### DIFF
--- a/src/contexts/ApiPromiseContextProvider.tsx
+++ b/src/contexts/ApiPromiseContextProvider.tsx
@@ -14,6 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
 
+/* eslint-disable @typescript-eslint/no-unused-vars */
 import { ApiPromise } from '@polkadot/api';
 import { ApiOptions } from '@polkadot/api/types';
 import { ProviderInterface } from '@polkadot/rpc-provider/types';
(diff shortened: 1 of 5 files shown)


src/contexts/ApiPromiseContextProvider.tsx
```diff
@@ -42,29 +42,48 @@ export function ApiPromiseContextProvider(props: ApiRxContextProviderProps): Rea
...
-      setApiPromise(_api);
-    });
-  }, [provider]);
+    if (!isReady && !apiPromise.isConnected) {
+      apiPromise.disconnect().then(() => {
+        setDisconnected(true);
+        logger.info(`${contextType} Resetting connection`);
+      });
+    }
+  }, [isReady]);
```
tomusdrw, line 63:
```suggestion
  }, [isReady, apiPromise]);
```

shouldn't this depend on `apiPromise`?

src/contexts/ApiPromiseContextProvider.tsx
```diff
@@ -42,29 +42,48 @@ export function ApiPromiseContextProvider(props: ApiRxContextProviderProps): Rea
...
+      apiPromise.isReady
+        .then(() => {
+          if (types) {
+            registry.register(types);
+          }
+          logger.info(`${contextType} type registration ready`);
+          setIsReady(true);
+        })
+        .catch((e) => logger.error(`${contextType} error in registration`, e));
+    }
```
tomusdrw, line 86: TBH I have no idea what is happening in all of these `use*Effect` handlers and why we need so many, but if it's necessary - so be it.

tomusdrw (approved):

PR #31, opened 2021-03-26
Changing pallet name to grandpa

hbulgarini:


src/hooks/useBridgedBlocks.ts
```diff
@@ -37,8 +37,7 @@ interface Props {
 const useBridgedBlocks = ({ isApiReady, api, chain }: Props) => {
   const [importedHeaders, setImportedHeaders] = useState('');
 
-  //const bridgedChain = `bridge${chain}`;
-  const bridgedChain = `bridge${chain}`;
+  const bridgedChain = `bridge${chain}Grandpa`;
```
tomusdrw, line 40: I'd if all chain-related names would be grouped in a single file (like constants). This may change in the future, and I'd rather not look for string concats in the entire codebase.
```js
function getGrandpaPalletName(chain) {
  return `bridge${chain}Grandpa`;
}
```

tomusdrw (approved):
lgtm, although it would be good to have all string concats like that in a single js file (kind of like constants)

src/hooks/useBridgedBlocks.ts
hbulgarini: i know what you mean and i agree... but this concatenation is used only once across the application and only in this file.
I feel that now is simpler having the constant in the file and going to this file specifically and change the name in case it is needed. But of course that i'm opened to implement that function if it worths it.

PR #32, opened 2021-03-26
Adding specific env files

hbulgarini:


.env.production
```diff
@@ -0,0 +1,11 @@
+REACT_APP_CHAIN_1=Rialto
+REACT_APP_CHAIN_2=Millau
```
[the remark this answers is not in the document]
hbulgarini: It is used and it has  a valid purpose: https://github.com/paritytech/parity-bridges-ui/blob/master/src/contexts/SourceTargetContextProvider.tsx#L19

Let's suppose that we have the bridge ready for polkadot and kusama, this production file will look like:
```
REACT_APP_CHAIN_1=Rialto
REACT_APP_CHAIN_2=Millau
```

.env.production
```diff
@@ -0,0 +1,11 @@
...
+REACT_APP_RIALTO_SUBSTRATE_PROVIDER=wss://wss.rialto.brucke.link
```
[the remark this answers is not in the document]
hbulgarini: agree ... i will remove the defaults in the JS code.

.env.test
```diff
@@ -0,0 +1,11 @@
+REACT_APP_CHAIN_1=Rialto
```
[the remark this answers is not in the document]
hbulgarini: I think this was explained across the PR comments. 
I feel it does not heart to keep the .env.example with generic values for documentation purposes.

src/configs/substrateProviders.ts
```diff
@@ -37,40 +35,35 @@ interface ChainConfigs {
   [chain: string]: ChainConfig;
 }
 
-// create a custom hasher (512 bits, combo of blake2 and keccak)
-function hasherH512(data: any) {
-  return u8aConcat(blake2AsU8a(data), keccakAsU8a(data));
-}
+const getChainNames = () => {
+  const chainVarName = 'REACT_APP_CHAIN_';
+  return [checkEnvVariable(`${chainVarName}1`), checkEnvVariable(`${chainVarName}2`)];
```
tomusdrw, line 40:
```suggestion
  return [
    checkEnvVariable('REACT_APP_CHAIN_1'),
    checkEnvVariable('REACT_APP_CHAIN_2')
  ];
```
This is overkill :) Having variables names explicit makes it much easier to look for them.

src/util/checkEnvVariables.ts
```diff
@@ -0,0 +1,54 @@
...
+  expectedCommonVariables.map((chainVariable) => {
+    const envVariable: ProcessEnv = process.env[`REACT_APP_${chainVariable}`];
+    if (envVariable) {
+      found = true;
+    }
+  });
+
+  if (!found) {
+    throw new Error(`Env Variable ${variable} was not defined`);
+  }
```
tomusdrw, line 50: Wait, this thing is mixing functionality quite heavily. We do two things here:

1. Extract a requested environment variable (`variable` argument) and fail if it's not set.
2. Validate that some more expected variables are being set.

That should be split into two separate functions, where (2) is simply being called at the application startup and not for every unrelated usage of `checkEvnVariable` function.

src/util/getDeriveAccount.ts
```diff
@@ -17,14 +17,14 @@
 import { compactAddLength, stringToU8a } from '@polkadot/util';
 import { blake2AsHex, decodeAddress, encodeAddress } from '@polkadot/util-crypto';
 
+const accountDerivation = 'pallet-bridge/account-derivation/account';
```
tomusdrw, line 20: So since the variable is not used, why do we have tests for it? (see below)

src/util/tests/getDeriveAccount.test.ts
```diff
@@ -17,6 +17,8 @@
 import getDeriveAccount from '../getDeriveAccount';
 
 describe('getDeriveAccount', () => {
+  process.env.REACT_APP_ACCOUNT_DERIVATION = 'pallet-bridge/account-derivation/account';
```
tomusdrw, line 20: This variable is not used any more - and that's correct, cause that's not something that can be configured. Please remove it from `.env`, `.env.example` and here.

tomusdrw (changes requested):
The `checkEnvVariable` is a bit overcomplicated, a function to retrieve arbitrary environment variable should not have side effects of checking existence of some other env variables.

hbulgarini (commented):
> The `checkEnvVariable` is a bit overcomplicated, a function to retrieve arbitrary environment variable should not have side effects of checking existence of some other env variables.

Addressed.

src/configs/substrateProviders.ts
```diff
@@ -37,40 +35,35 @@ interface ChainConfigs {
...
-    types: customTypesMillau
-  },
-  Rialto: {
-    SS58Format: getSS58Format(process.env.REACT_APP_RIALTO_SS58_PREFIX, 48),
-    bridgeId: process.env.REACT_APP_RIALTO_BRIDGE_ID || 'rlto',
-    provider: getProvider(RIALTO_SUBSTRATE_PROVIDER),
-    types: customTypesRialto
-  }
+export const getChainConfigs = (): ChainConfigs => {
+  checkExpectedVariables();
```
tomusdrw, line 61: Can't we call this function even before we start `React`? IMHO it's better to fail as soon as possible and not wait for some components to render.

tomusdrw (approved):
lgtm!

src/configs/substrateProviders.ts
hbulgarini: yes definitely ... i cannot agree more on this

hbulgarini:
closes #30

PR #39, opened 2021-04-01
Hb execution steps

hbulgarini:


src/configs/substrateCustomTypes/customTypesMillau.json
```diff
@@ -117,7 +112,7 @@
       "Ecdsa": "[u8;33]"
     }
   },
-  "MesssagesProofOf": {
+  "MessagesProofOf": {
```
tomusdrw, line 115: This should be gone after #40, right?

src/hooks/useCurrentExecutionStatus.ts
```diff
@@ -0,0 +1,142 @@
...
+import { TransactionStatusEnum } from '../types/transactionTypes';
+import getSubstrateDynamicNames from '../util/getSubstrateDynamicNames';
+
+interface Step {
+  chainType: string;
+  label: string;
+  status: string;
+}
+
+export default function useCurrentExecutionStatus() {
```
tomusdrw, line 37: Why is this a hook instead of being a component? If we want to re-use this for multiple transactions the transaction has to be coming as a parameter.

src/hooks/useCurrentExecutionStatus.ts
```diff
@@ -0,0 +1,142 @@
...
+      const latestReceivedNonceCall = await targetApi.rpc.state.call<Codec>(
+        latestReceivedNonceMethodName,
+        laneId,
+        blockHash.toJSON()
+      );
+
+      // @ts-ignore
+      const latestReceivedNonceCallType = targetApi.registry.createType('MessageNonce', latestReceivedNonceCall);
+      const latestReceivedNonce = latestReceivedNonceCallType.toString();
+      setNonceOfTargetFinalizedBlock(parseInt(latestReceivedNonce));
```
tomusdrw, line 69: Why do we go from `MessageNonce -> String -> parseInt`? Isn't there a better way to get the numeric value directly from `MessageNonce`?

src/hooks/useCurrentExecutionStatus.ts
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    const stepEvaluator = (transactionValue: string | number | null, chainValue: number | null) => {
+      if (!transactionValue || !chainValue) {
+        return false;
+      }
+      return chainValue > transactionValue;
+    };
+
+    const isDone = (status: boolean | boolean) => (currentTransaction.block && status ? 'DONE' : 'RUNNING');
```
tomusdrw, line 104: `is*` functions should return boolean values, while this one returns a string. I think it's misleading and maybe better to give the function more descriptive name, like `doneStatus` or `completionStatus`?

tomusdrw, line 104:
```suggestion
    const isDone = (status: boolean) => (currentTransaction.block && status ? 'DONE' : 'RUNNING');
```

src/hooks/useCurrentExecutionStatus.ts
```diff
@@ -0,0 +1,142 @@
...
+    const step6 = stepEvaluator(currentTransaction.messageNonce, latestReceivedNonceOnSource);
+
+    const steps = [
+      { chainType: sourceChain, label: 'Including message in block', status: step1 },
+      { chainType: sourceChain, label: 'Waiting for block finality on source', status: isDone(step2) },
+      { chainType: targetChain, label: 'Source Chain block finality received on target', status: isDone(step3) },
+      { chainType: targetChain, label: 'Message delivered', status: isDone(step4) },
+      { chainType: targetChain, label: 'Message finality', status: isDone(step5) },
+      { chainType: sourceChain, label: 'Delivery confirmation on source', status: isDone(step6) }
+    ];
```
tomusdrw, line 119: If this was made a component this would simply be a list of components as well, I don't feel like abstracting this is required.

src/hooks/useCurrentExecutionStatus.ts
```diff
@@ -0,0 +1,142 @@
...
+      return chainValue > transactionValue;
+    };
+
+    const isDone = (status: boolean | boolean) => (currentTransaction.block && status ? 'DONE' : 'RUNNING');
+    const step1 = currentTransaction.block ? `included in block ${currentTransaction.block}` : isDone(false);
+    const step2 = stepEvaluator(currentTransaction.block, parseInt(bestBlockFinalized));
+    const step3 = stepEvaluator(currentTransaction.block, parseInt(importedHeaders));
+    const step4 = stepEvaluator(currentTransaction.messageNonce, latestReceivedNonceRuntimeApi);
+    const step5 = stepEvaluator(currentTransaction.messageNonce, nonceOfTargetFinalizedBlock);
+    const step6 = stepEvaluator(currentTransaction.messageNonce, latestReceivedNonceOnSource);
```
tomusdrw, line 110: Calling the variables `step*` is not super informative. Like it's really hard right now to figure out what the step actually is without external context, I'd be more in favour in naming them properly:

```suggestion
    const sourceTransactionIncluded = currentTransaction.block ? `included in block ${currentTransaction.block}` : isDone(false);
    const sourceTransactionFinalized = stepEvaluator(currentTransaction.block, parseInt(bestBlockFinalized));
    const blockFinalityRelayed = stepEvaluator(currentTransaction.block, parseInt(importedHeaders));
    const messageDelivered = stepEvaluator(currentTransaction.messageNonce, latestReceivedNonceRuntimeApi);
    const messageFinalizedOnTarget = stepEvaluator(currentTransaction.messageNonce, nonceOfTargetFinalizedBlock);
    const sourceConfirmationReceived = stepEvaluator(currentTransaction.messageNonce, latestReceivedNonceOnSource);
```

src/hooks/useSendMessage.ts
```diff
@@ -64,8 +90,31 @@ function useSendMessage({ isRunning, setIsRunning, setExecutionStatus, message,
...
       }
-      await bridgeMessage.signAndSend(sourceAccount, { ...options });
-      setExecutionStatus(message.successfull);
+
+      const unsub = await bridgeMessage.signAndSend(sourceAccount, { ...options }, ({ events = [], status }) => {
+        if (status.isInBlock) {
+          events.forEach(({ event: { data, method } }) => {
+            if (method.toString() === 'MessageAccepted') {
+              const messageNonce = data.toArray()[1].toString();
+              sourceApi.rpc.chain.getBlock(status.asInBlock).then((res) => {
```
tomusdrw, line 99: Error is not handled from this promise.

src/hooks/useSendMessage.ts
```diff
@@ -64,8 +90,31 @@ function useSendMessage({ isRunning, setIsRunning, setExecutionStatus, message,
...
+                const block = res.block.header.number.toNumber();
```
tomusdrw, line 100: Why does it have to be a number here? In theory block numbers can be `u128` which is way more than what JS number  can store. I'd prefer to defer any numeric conversions as long as possible (i.e. only convert if it's strictly required for comparison or arithmetic) - this will make it easier to introduce `BigNumber` in the future if needed.

src/reducers/transactionReducer.ts
```diff
@@ -15,14 +15,29 @@
...
-import type { TransactionsActionType, TransactionState } from '../types/transactionTypes';
+import type { TransactionsActionType, TransactionState, UpdatedTransanctionStatus } from '../types/transactionTypes';
+
+const updateTransactionObject = (state: TransactionState, updatedValues: UpdatedTransanctionStatus) => {
+  const newState = { ...state };
+  if (state.currentTransaction) {
+    Object.keys(updatedValues).map((key) => {
+      newState.currentTransaction![key] = updatedValues[key];
+    });
+  }
```
tomusdrw, line 26: This can be a single statement and IMHO it's mor readable than the loop. The fact that we receive `updatedValues` in case there is no `currentTransaction` is imo a bug, and we should prevent it in some other place.

```suggestion
  return {
     ...state,
     currentTransaction: {
        ...state.currentTransaction,
        updatedValues,
     }
  };
```

src/hooks/useCurrentExecutionStatus.ts
```diff
@@ -0,0 +1,142 @@
...
+      { chainType: targetChain, label: 'Message delivered', status: isDone(step4) },
+      { chainType: targetChain, label: 'Message finality', status: isDone(step5) },
+      { chainType: sourceChain, label: 'Delivery confirmation on source', status: isDone(step6) }
+    ];
+
+    if (step6) {
+      dispatchTransaction(
+        TransactionActionCreators.updateTransactionStatus({ status: TransactionStatusEnum.COMPLETED })
+      );
+    }
```
tomusdrw, line 125: Hook dispatching a transaction looks a bit weird to me, let me clarify what I mean.

Since you have split hooks and global state imho they should be independent from each other. However here you are introducing a tight coupling between the two. If I create a component that is only interested in transaction status (i.e. doesn't call `useCurrentExecutionStatus`) I'll never see the transaction being completed, which is not quite what I expect.

That makes me feel this is not the right abstraction you have here and having such important logic within a not-quite-related hook is not the right design choice. I think that status tracking should be isolated to one hook (and actually as mentioned some place else we should be able to track multiple transactions at once, but that can be done in the future).

src/util/transactionLocalStorage.ts
```diff
@@ -0,0 +1,30 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+
+import { TransanctionStatus } from '../types/transactionTypes';
+
+const includeNewTransaction = (transaction: TransanctionStatus) => {
+  const rawTransactions = localStorage.getItem('bridge-ui-transactions');
```
tomusdrw, line 20:
```suggestion
  const rawTransactions = localStorage.getItem(STORAGE_KEY);
```
This should be extracted to a file-local constant cause it's being repeated 3 times in that file.

src/util/transactionLocalStorage.ts
```diff
@@ -0,0 +1,30 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+
+import { TransanctionStatus } from '../types/transactionTypes';
+
+const includeNewTransaction = (transaction: TransanctionStatus) => {
```
tomusdrw, line 19:
```suggestion
const appendNewTransaction = (transaction: TransanctionStatus) => {
```

src/util/transactionLocalStorage.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (!rawTransactions || !rawTransactions.length) {
+    const first = [transaction];
+    return localStorage.setItem('bridge-ui-transactions', JSON.stringify(first));
+  }
+  const transactions = JSON.parse(rawTransactions);
+  transactions.push(transaction);
+  return localStorage.setItem('bridge-ui-transactions', JSON.stringify(transactions));
+};
+
+export { includeNewTransaction };
```
tomusdrw, line 30: The whole logic should be wrapped in `try-catch` in case `localStorage` is not available or the size is exceeded.

tomusdrw: Note that transactions are also chain-specific so we should have separate `STORAGE_KEYs` for every chain.

src/actions/transactionActions.ts
```diff
@@ -29,9 +35,26 @@ const setReceiverAddress = (receiverAddress: string) => ({
   type: TransactionActionTypes.SET_RECEIVER_ADDRESS
 });
 
+const createTransactionStatus = (initialTransaction: TransanctionStatus) => {
+  includeNewTransaction(initialTransaction);
```
tomusdrw, line 39: Seems that currently we only append this stuff and never read it, right? I'd suggest removing this for the current PR cause it adds no value at all and adding a separate issue to make it work properly.

src/util/transactionLocalStorage.ts
```diff
@@ -0,0 +1,39 @@
...
+    if (!rawTransactions || !rawTransactions.length) {
+      const first = [transaction];
+      return localStorage.setItem(storageKey, JSON.stringify(first));
+    }
+    const transactions = JSON.parse(rawTransactions);
+    transactions.push(transaction);
+    return localStorage.setItem(storageKey, JSON.stringify(transactions));
+  } catch (e) {
+    logger.error(e);
+    throw new Error('Issue appending new transaction to local storage');
```
tomusdrw, line 35: We shouldn't be throwing the error - it will crash the entire React UI in cases that user can't really control.

src/util/transactionLocalStorage.ts
```diff
@@ -0,0 +1,39 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+
+import { CHAIN_1, CHAIN_2 } from '../configs/substrateProviders';
+import { TransanctionStatus } from '../types/transactionTypes';
+import getSubstrateDynamicNames from './getSubstrateDynamicNames';
+import logger from './logger';
+
+const appendNewTransaction = (transaction: TransanctionStatus) => {
+  const { storageKey } = getSubstrateDynamicNames(`${CHAIN_1}-${CHAIN_2}`);
```
tomusdrw, line 23: I think it would be better off as a separate function. `getSubstrateDynamicNames` has so far always been used by passing one chain name, now we pass some concatentaion and basically all other values of the returned object are never going to be used cause they have no sense at all.

src/components/TransactionStatus.tsx
```diff
@@ -0,0 +1,166 @@
...
+        label: 'Delivery confirmation on source',
+        status: completionStatus(sourceConfirmationReceived)
+      }
+    ];
+
+    if (sourceConfirmationReceived) {
+      dispatchTransaction(
+        TransactionActionCreators.updateTransactionStatus({ status: TransactionStatusEnum.COMPLETED })
+      );
+    }
```
tomusdrw, line 146: I still don't like this as mentioned earlier. I think this is incorrect split between global (hook) and local state.

1. Transaction status is global (hook-local).
2. Tracking confirmations is local.

If I rely on the transaction status I HAVE TO include the transaction status component as well, cause otherwise I'll not receive `COMPLETED` status.

I see two possible solutions:
1. Make transactions status local (i.e. part of this component) and remove all actions and reducer for `currentTransaction`.
2. Make it possible to track MULTIPLE (ALL) transaction status globally, without including any components. I.e. this means that `COMPLETED` should be triggered by the hook and the state should not have `currentTransaction` but rather a map of `transactions: { [nonce]: status }` or sth.

src/util/transactionLocalStorage.ts
tomusdrw: Again, since we never read this values in the PR, please move the code into an issue / separate PR, cause now we waste time discussing code that has nothing to do with the PR.

tomusdrw (approved):

PR #40, opened 2021-04-01
Adding process for reading customTypes from common repository

hbulgarini:


package.json
```diff
@@ -37,11 +37,12 @@
     "winston": "^3.3.3"
   },
   "scripts": {
-    "start": "PORT=3001 react-app-rewired start",
+    "start": "yarn run downloadCustomTypes && PORT=3001 react-app-rewired start",
```
tomusdrw, line 40:
```suggestion
    "start": "yarn downloadCustomTypes && PORT=3001 react-app-rewired start",
```

.github/workflows/js.yml
```diff
@@ -43,6 +44,7 @@ jobs:
       with:
         node-version: '14'
     - run: yarn
+    - run: yarn run downloadCustomTypes
```
tomusdrw, line 47:
```suggestion
    - run: yarn downloadCustomTypes
```

.github/workflows/js.yml
```diff
@@ -25,6 +25,7 @@ jobs:
...
```
tomusdrw, line 28:
```suggestion
    - run: yarn downloadCustomTypes
```
I'm pretty sure that `run` is not needed for yarn.

package.json
tomusdrw: I wonder if we could make the script be run in `pre-` hooks, but it would have to be `pre-start`, `pre-lint` and `pre-build`, so maybe easier to be explicit on the call site.

Can you update the readme as well and mention that it's required to fetch the files? Cause the error messages might be quite confusing (i.e. missing file).

hbulgarini: i can add the prestart , which is supported in create-react-app , but for the build just having the step configured in the github action i think it does the job. Also i like be explicit about this dependency in the logs as a separate step.

tomusdrw (approved):

PR #42, opened 2021-04-06
Adding snackbar functionalities ( global Messages)

hbulgarini:


hbulgarini:
Once #39 is merged, i will apply the corresponding dispatch message actions

tomusdrw (approved):
I don't see `triggerMessage` being used anywhere, but I'm assuming it's for future use. Can you elaborate what kind of errors/messages you plan to display there?

hbulgarini:
> I don't see `triggerMessage` being used anywhere, but I'm assuming it's for future use. Can you elaborate what kind of errors/messages you plan to display there?

Sure....
This would replace that feedback test functionality based on local state `setExecutionStatus`  which was used for provide feedback when it was an error or when the transaction was sent successfully.
Anyway i have uploaded some messages proposals where you can validate the usage.

PR #43, opened 2021-04-09
Connections Refactor.

hbulgarini:
This PR introduces a considerable enhancement to the pattern implemented for handling for chain connections.
  
Here the list of improvements: 

- Removed repeated and non performing ApiPromise context components. Now everything is handled from the `useApiConnection` hook   from `App`  preventing unwanted re renders and spawn connections to the node.
- Because of this new design , now switching chains is immediate since connections are not reseted anymore ( there are assigned to source/target accordingly )
-  Solves node's abnormal closure connect issues ( #12)


src/util/shortenItem.ts
```diff
@@ -0,0 +1,23 @@
...
+//
+// Parity Bridges UI is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+
+export default (item: string) => {
```
tomusdrw, line 17: This does not seem to be used anywhere.

tomusdrw (approved):

src/util/shortenItem.ts
hbulgarini: It is now ...! thanks

PR #69, opened 2021-04-18
material UI migration + account selector base.

hbulgarini:


tomusdrw (commented):
It's a bit hard to review cause it contains #68 and some extra features, but looks good initially given all of the TODOs are addressed or logged into the issue tracker.

src/components/Accounts.tsx
```diff
@@ -14,13 +14,19 @@
 // You should have received a copy of the GNU General Public License
```
hbulgarini, line 14: @goldsteinsveta : This component contains the account selector dropdown. Please take into account that component [Account](https://github.com/paritytech/parity-bridges-ui/pull/69/files#diff-3ad4937f6f0f4d3aedd698a6d1d6547b848be2e418fcd11b13b9131529183991) contains the custom style for the dropdown element.

src/App.tsx
```diff
@@ -14,12 +14,12 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
 
+import Backdrop from '@material-ui/core/Backdrop';
+import CircularProgress from '@material-ui/core/CircularProgress';
```
wirednkod, line 18: You can one-line these 2 by:
```
import { CircularProgress,  Backdrop } from '@material-ui/core';
```

src/App.tsx
```diff
@@ -14,12 +14,12 @@
...
 import { SnackbarProvider } from 'notistack';
 import React from 'react';
 import { BrowserRouter, Route, Switch } from 'react-router-dom';
-import { Dimmer, Loader } from 'semantic-ui-react';
 
-import TopBar from './components/TopBar';
 import { getChainProviders } from './configs/substrateProviders';
 import { CHAIN_1, CHAIN_2 } from './configs/substrateProviders';
```
wirednkod, line 24: add these 2 in 1 line:
```
import { getChainProviders, CHAIN_1, CHAIN_2 } from './configs/substrateProviders';

File: src/App.tsx [hunk 4]
```diff
@@ -40,11 +40,13 @@ function App() {
     { apiConnection: apiChain2, chainName: CHAIN_2, polkadotjsUrl: providers[CHAIN_2].polkadotjsUrl }
   ];
 
-  if (!apiChain1.isApiReady || !apiChain2.isApiReady) {
+  const apiReady = apiChain1.isApiReady && apiChain2.isApiReady;
+
+  if (!apiReady) {
```
wirednkod, line 45: im not sure but I think you should put this in a use effect and the outcome inside a state in order to conditional render

src/components/Accounts.tsx
```diff
@@ -14,13 +14,19 @@
...
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
 
+import Container from '@material-ui/core/Container';
+import FormControl from '@material-ui/core/FormControl';
+import InputLabel from '@material-ui/core/InputLabel';
+import MenuItem from '@material-ui/core/MenuItem';
+import Select from '@material-ui/core/Select';
```
wirednkod, line 21: As before you can oneline these into:
```
import { Container, FormControl, InputLabel, MenuItem, Select } from '@material-ui/core';
```
make stuff clearer

src/components/CustomCall.tsx
```diff
@@ -14,8 +14,10 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
 
+import Button from '@material-ui/core/Button';
+import Container from '@material-ui/core/Container';
+import TextField from '@material-ui/core/TextField';
 import React, { useState } from 'react';
```
wirednkod, line 20: Always as a good practice start with import React.... and all others should follow

src/components/CustomCall.tsx
```diff
@@ -14,8 +14,10 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
 
+import Button from '@material-ui/core/Button';
+import Container from '@material-ui/core/Container';
+import TextField from '@material-ui/core/TextField';
```
wirednkod, line 19: Same as before:
```
import { Button, Container, TextField } from '@material-ui/core';
```

src/components/Remark.tsx
```diff
@@ -14,8 +14,10 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
 
+import Button from '@material-ui/core/Button';
```
wirednkod, line 17: same one-liner comment as before

src/components/TransactionDisplay.tsx
```diff
@@ -14,8 +14,11 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
 
+import Card from '@material-ui/core/Card';
```
wirednkod, line 17: same one-liner comment as before

src/components/Transfer.tsx
```diff
@@ -14,8 +14,10 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
 
+import Button from '@material-ui/core/Button';
```
wirednkod, line 17: same one-liner comment as before

src/screens/Main.tsx
```diff
@@ -14,9 +14,12 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
 
+import Backdrop from '@material-ui/core/Backdrop';
```
wirednkod, line 17: same one-liner comment as before

src/screens/Main.tsx
```diff
@@ -44,62 +47,57 @@ export function Main({ className }: Props) {
 
   if (!areApiLoading) {
```
wirednkod, line 48: state (and conditional rendering) should be in useEffect to let react to decide

wirednkod (changes requested):

src/App.tsx
hbulgarini: yes, but by doing we would be importing the whole material package and the bundle size would increase. With this such a big packages i prefer to be explicit and bring the specific packages. I know that some configuration can be done as far i could read but i like being explicit on this cases ( same with lodash )

src/App.tsx
hbulgarini: IMHO i didn't see the need to implement that pattern in this root component since it will not be reused across the application. This is is pretty straight forward and i would be avoiding  to add  effect + state + conditional rendering.

Anyway your comment is more than valid and for the rest of the application i have applied the exact pattern you have described by using the hook : [useLoadingApi](https://github.com/paritytech/parity-bridges-ui/blob/master/src/hooks/useLoadingApi.ts). Then the value pull from this hook is pretty used on different effects.

That pattern would imply to add an effect and state that purpose is already served this this code. Anyway please take into account that this pattern is not used like that anymore across the code base.

wirednkod: This pattern (useEffect) in react functional components is replacing the lifecycle of class components. So its kind of a "prerequisite" in react in order to allow react handle things (instead of doing it in a custom way).
Have a look here:
https://github.com/paritytech/substrate-connect/blob/a5f5d71e5fd59222f0c57e1e3f3b4e5c4d115586/projects/burnr/src/App.tsx#L42

wirednkod (approved):

src/components/Accounts.tsx
goldsteinsveta: I'll probably rewrite it in pure MUI and remove styled-components dependency when will be styling this. Don't worry about it!

goldsteinsveta (approved):

src/components/Accounts.tsx
```diff
@@ -37,30 +43,66 @@ const formatOptions = (accounts: Array<KeyringPair>) =>
...
+      <InputLabel id="demo-simple-select-label">{sourceChain} Account</InputLabel>
+      <FormControl variant="outlined" className="formControl">
+        <Select
+          value={value}
+          onChange={onChange}
+          name="name"
+          inputProps={{
+            id: 'name-native-error'
+          }}
+          renderValue={(): React.ReactNode => <Input />}
```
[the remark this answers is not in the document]
goldsteinsveta: nit: I'd probably name it to something like "AccountSelected", so you don't need to explain it :) But I got it no problem

PR #75, opened 2021-04-20
Account Selector display both chain accounts and switch source/target based on the account selected.

hbulgarini:
This PR includes two functionalities:

- Display both chains in the account selector with their respective derived accounts.
- Switch source target according to what it is selected.

As i pointed out in in the comment of the file `src/util/formatAccounts.ts`  this implementation does not support specific accounts per chain. That will be implemented on task  #74

### DIFF
--- a/src/components/Accounts.tsx
+++ b/src/components/Accounts.tsx
@@ -90,7 +90,7 @@ const Accounts = ({ className }: Props) => {
           }}
           renderValue={(): React.ReactNode => <AccountSelected />}
         >
-          {accounts.length && chains.map((chain) => renderAccounts(chain))}
+          {chains.map((chain) => renderAccounts(chain))}
         </Select>
       </FormControl>
       {derivedAccount && (
--- a/src/components/Accounts.tsx
+++ b/src/components/Accounts.tsx
@@ -13,6 +13,23 @@
 //
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+// Copyright 2021 Parity Technologies (UK) Ltd.
+// This file is part of Parity Bridges UI.
+//
+// Parity Bridges UI is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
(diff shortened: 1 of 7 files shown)


src/reducers/sourceTargetReducer.ts
```diff
@@ -20,18 +20,21 @@ import { ChainDetails, SourceTargetAction, SourceTargetState } from '../types/so
...
+            sourcePolkadotjsUrl: state[ChainDetails.TARGET].targetPolkadotjsUrl
+          },
+          [ChainDetails.TARGET]: {
+            targetApiConnection: state[ChainDetails.SOURCE].sourceApiConnection,
+            targetChain: state[ChainDetails.SOURCE].sourceChain,
+            targetPolkadotjsUrl: state[ChainDetails.SOURCE].sourcePolkadotjsUrl
+          }
+        };
+      }
+      return state;
```
wirednkod, line 37: if not then what? Im not sure that a switch is needed here

wirednkod (changes requested):

src/reducers/sourceTargetReducer.ts
hbulgarini: The logic behind this if is to validate if the chain that comes from the payload is different than the current chain. If this is the case then source and target values are flipped ( because an account from the target chain was selected ).
In case an account from the source is chosen then no changes are needed and the source/target remains the same.

Regarding the usage of the switch to determine the action is a pattern which mainly comes from redux reducers: https://reactjs.org/docs/hooks-reference.html#usereducer

At this point in time it does not makes too much sense but as soon as we need new actions on this reducer, the switch statement will have more sense.

wirednkod (approved):

src/components/Accounts.tsx
```diff
@@ -19,55 +19,58 @@ import FormControl from '@material-ui/core/FormControl';
...
+        }}
+      >
         <Account text={text} value={value} showDerivedBalance />
       </MenuItem>
     ));
     return [<SubHeader key={chain} chain={chain} />, items];
   };
 
-  const Input = () => {
+  const AccountSelected = () => {
```
hbulgarini, line 73: @goldsteinsveta : i addressed your suggestion from PR #76

PR #76, opened 2021-04-20
fix hook useBalance and replace hardcoded balances with actual balanc…

### LINKED ISSUE #49: Display balances for source and target accounts.

wirednkod:
…es from source and target chains; Fixes #49

### DIFF
--- a/src/components/Account.tsx
+++ b/src/components/Account.tsx
@@ -29,9 +29,7 @@ interface Props {
 }
 
 const Account = ({ className, text, value, showDerivedBalance = false }: Props) => {
-  const results = useBalance(value, true);
-  const source = results[0];
-  const target = results[1];
+  const [source, target] = useBalance(value, true);
 
   return (
     <Container className={className}>
@@ -49,11 +47,11 @@ const Account = ({ className, text, value, showDerivedBalance = false }: Props) 
 
       <div className="balances">
         <div className="balance">
-          <p>{source ? source[0] : '-'}</p>
+          <p>{source ? source.formattedBalance : '-'}</p>
         </div>
         {showDerivedBalance && (
           <div className="balance">
-            <p>{target ? target[0] : '-'}</p>
+            <p>{target ? target.formattedBalance : '-'}</p>
           </div>
         )}
       </div>
--- a/src/components/Account.tsx
+++ b/src/components/Account.tsx
@@ -25,11 +25,12 @@ interface Props {
   value: string;
   text: string;
   className?: string;
+  chain?: string | undefined;
   showDerivedBalance?: boolean;
 }
 
-const Account = ({ className, text, value, showDerivedBalance = false }: Props) => {
-  const [source, target] = useBalance(value, true);
+const Account = ({ className, text, value, chain, showDerivedBalance = false }: Props) => {
+  const [source, target] = useBalance(text, value, chain, true);
 
   return (
     <Container className={className}>
--- a/src/components/Account.tsx
+++ b/src/components/Account.tsx
@@ -19,6 +19,8 @@ import People from '@material-ui/icons/People';
 import React from 'react';
 import styled from 'styled-components';
 
+import useBalance from '../hooks/useBalance';
+
 interface Props {
   value: string;
   text: string;
@@ -26,33 +28,38 @@ interface Props {
   showDerivedBalance?: boolean;
 }
 
-// TO-DO: To replace fake balances by real balances
-const Account = ({ className, text, value, showDerivedBalance = false }: Props) => (
-  <Container className={className}>
-    <div className="address">
-      <div className="topAddress">
-        <People />
-        <div className="text">
-          <p>{text}</p>
-        </div>
-      </div>
-      <div className="bottomAddress">
-        <p>{value}</p>
-      </div>
-    </div>
+const Account = ({ className, text, value, showDerivedBalance = false }: Props) => {
+  const results = useBalance(value, true);
+  const source = results[0];
+  const target = results[1];
 
-    <div className="balances">
-      <div className="balance">
-        <p>0.1 RLTO</p>
-      </div>
-      {showDerivedBalance && (
-        <div className="balance">
-          <p>0.1 MLAU</p>
+  return (
+    <Container className={className}>
+      <div className="address">
+        <div className="topAddress">
+          <People />
+          <div className="text">
+            <p>{text}</p>
+          </div>
         </div>
-      )}
-    </div>
-  </Container>
-);
+        <div className="bottomAddress">
+          <p>{value}</p>
+        </div>
+      </div>
+
+      <div className="balances">
+        <div className="balance">
+          <p>{source ? source[0] : '-'}</p>
+        </div>
+        {showDerivedBalance && (
+          <div className="balance">
+            <p>{target ? target[0] : '-'}</p>
+          </div>
+        )}
+      </div>
+    </Container>
+  );
+};
 
 export default styled(Account)`
   margin: auto 0;
--- a/src/components/Accounts.tsx
+++ b/src/components/Accounts.tsx
@@ -54,7 +54,6 @@ const Accounts = ({ className }: Props) => {
     setCurrentAccount(value, chain);
   };
   const renderAccounts = (chain: string) => {
-    console.log('chain is:', chain);
     const formatedAccounts = formatAccounts(accounts, chain);
     const items = formatedAccounts.map(({ text, value, key }: any) => (
       <MenuItem
--- a/src/components/Accounts.tsx
+++ b/src/components/Accounts.tsx
@@ -53,8 +53,8 @@ const Accounts = ({ className }: Props) => {
   const onChange = (value: string, chain: string) => {
     setCurrentAccount(value, chain);
   };
-
   const renderAccounts = (chain: string) => {
+    console.log('chain is:', chain);
     const formatedAccounts = formatAccounts(accounts, chain);
     const items = formatedAccounts.map(({ text, value, key }: any) => (
       <MenuItem
@@ -64,7 +64,7 @@ const Accounts = ({ className }: Props) => {
           onChange(value, chain);
         }}
       >
-        <Account text={text} value={value} showDerivedBalance />
+        <Account text={text} value={value} showDerivedBalance chain={chain} />
       </MenuItem>
     ));
     return [<SubHeader key={chain} chain={chain} />, items];
--- a/src/hooks/useBalance.ts
+++ b/src/hooks/useBalance.ts
@@ -91,7 +91,6 @@ const useBalance = (name: string, address: string, chain: string | undefined, pr
     };
 
     if (address) {
-      console.log('name: ', name, 'address', address, ' => derievedAddress', derivedAddress);
       unsubscribeSource = getBalance(sourceApi, address, setSourceState);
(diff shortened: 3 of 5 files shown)


src/hooks/useBalance.ts
```diff
@@ -0,0 +1,89 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+import { Balance } from '@polkadot/types/interfaces';
+import { formatBalance } from '@polkadot/util';
+import BN from 'bn.js';
+import { useEffect, useState } from 'react';
+
+import { useSourceTarget } from '../contexts/SourceTargetContextProvider';
+
+type State = [string, Balance, boolean, string];
```
hbulgarini, line 23: why not creating and object instead of an array so we have a more explicit state like: 
`sourceState.balance` or `sourceState.free`

src/hooks/useBalance.ts
```diff
@@ -0,0 +1,89 @@
...
+            }),
+            data.free,
+            data.free.isZero(),
+            data.free.registry.chainTokens[0]
+          ]);
+        })
+        .then((u): void => {
+          unsubscribeSource = u;
+        })
+        .catch(console.error);
```
hbulgarini, line 60: We use a custom logger helper : https://github.com/paritytech/parity-bridges-ui/blob/master/src/util/logger.ts to avoid `console.log` or `console.error`.

Also in case you want to show the error to the user you can dispatch the message to the snackbar component to show the error/warning/info. Here you have an example of the ussage:

https://github.com/paritytech/parity-bridges-ui/blob/master/src/hooks/useSendMessage.ts#L137

I need to confirm this , but i think that `e.message` must be provided instead of `e`

src/hooks/useBalance.ts
```diff
@@ -0,0 +1,89 @@
...
+            }),
+            data.free,
+            data.free.isZero(),
+            data.free.registry.chainTokens[0]
+          ]);
+        })
+        .then((u): void => {
+          unsubscribeTarget = u;
+        })
+        .catch(console.error);
```
hbulgarini, line 78: same as above.

src/components/Account.tsx
```diff
@@ -19,40 +19,47 @@ import People from '@material-ui/icons/People';
...
 
-    <div className="balances">
-      <div className="balance">
-        <p>0.1 RLTO</p>
-      </div>
-      {showDerivedBalance && (
+      <div className="balances">
         <div className="balance">
-          <p>0.1 MLAU</p>
+          <p>{source ? source[0] : '-'}</p>
```
hbulgarini, line 52: i think that if the balance would be an object instead of an array this would be easier to read. like:
`<p>{source ? source.balance : '-'}</p>`

hbulgarini (changes requested):

src/components/Account.tsx
```diff
@@ -19,40 +19,47 @@ import People from '@material-ui/icons/People';
...
-  <Container className={className}>
-    <div className="address">
-      <div className="topAddress">
-        <People />
-        <div className="text">
-          <p>{text}</p>
+const Account = ({ className, text, value, showDerivedBalance = false }: Props) => {
+  const results = useBalance(value, true);
+  const source = results[0];
+  const target = results[1];
```
tomusdrw, line 34:
```suggestion
  const [source, target] = useBalance(value, true);
```

src/hooks/useBalance.ts
```diff
@@ -0,0 +1,89 @@
...
+      sourceApi.query.system
+        .account(address, ({ data }): void => {
+          setSourceState([
+            formatBalance(data.free, {
+              decimals: sourceApi.registry.chainDecimals[0],
+              forceUnit: '-',
+              withSi: providedSi
+            }),
+            data.free,
+            data.free.isZero(),
```
tomusdrw, line 53: Why is this useful? Can't you just do `source[1].isZero()` instead of `source[2]`?

src/hooks/useBalance.ts
```diff
@@ -0,0 +1,89 @@
...
+            data.free.registry.chainTokens[0]
+          ]);
+        })
+        .then((u): void => {
+          unsubscribeSource = u;
+        })
+        .catch(console.error);
+
+      targetApi.query.system
+        .account(address, ({ data }): void => {
```
tomusdrw, line 63: This is incorrect, we shouldn't be querying the same account on both source and target - rather we need to query "bridge account" aka "derived account" on the target chain.

src/hooks/useBalance.ts
```diff
@@ -0,0 +1,89 @@
...
+              decimals: sourceApi.registry.chainDecimals[0],
+              forceUnit: '-',
+              withSi: providedSi
+            }),
+            data.free,
+            data.free.isZero(),
+            data.free.registry.chainTokens[0]
+          ]);
+        })
+        .then((u): void => {
```
tomusdrw, line 57: Can't we use `async/await` to simplify the code?

src/hooks/useBalance.ts
```diff
@@ -0,0 +1,89 @@
...
+    let unsubscribeSource: null | (() => void) = null;
+    let unsubscribeTarget: null | (() => void) = null;
+    if (address) {
+      sourceApi.query.system
+        .account(address, ({ data }): void => {
+          setSourceState([
+            formatBalance(data.free, {
+              decimals: sourceApi.registry.chainDecimals[0],
+              forceUnit: '-',
+              withSi: providedSi
```
tomusdrw, line 50: We need proper units and human-readable formatting here.

src/hooks/useBalance.ts
```diff
@@ -0,0 +1,89 @@
...
+      sourceApiConnection: { api: sourceApi }
+    },
+    targetChainDetails: {
+      targetApiConnection: { api: targetApi }
+    }
+  } = useSourceTarget();
+
+  useEffect((): (() => void) => {
+    let unsubscribeSource: null | (() => void) = null;
+    let unsubscribeTarget: null | (() => void) = null;
```
tomusdrw, line 42:
```suggestion
    let unsubscribeSource = () => {};
    let unsubscribeTarget = () => {};
```
Wouldn't that work? Also if you do this, you can skip the `unsubscribeSource && ` check in the function returned from `useEffect`.

src/hooks/useBalance.ts
wirednkod: I would like to discuss this a bit more in order to understand

tomusdrw: So I'm not sure exactly how `formatBalance` works and perhaps we are already doing what  I want to achieve.
I'd like the values to be displayed like that:
```
1 TOK
1 kTOK
1 MTOK
1 mTOK
```
etc. So using SI prefixes and unit symbol - i.e avoiding a situation where full number of decimals is being displayed.
My suggestion would be also to display full balance (no SI; with thousand separator) in a tooltip (`title` attribute is good enough).

tomusdrw: Discussed this offband and the API is doing exactly what I mean, the unit should be loaded from the chainspec's `properties` section and currently it's missing for Rialto & Millau, hence it displays incorrectly, but we will add it shortly.

src/hooks/useBalance.ts
wirednkod: I had to follow different way due to the async/await and cause I wanted reusable functions :)

src/hooks/useBalance.ts
wirednkod: > display full balance (no SI; with thousand separator) in a tooltip (title attribute is good enough).

I'll make this in different PR though (keep scope of PRs small)

hbulgarini (changes requested):

src/hooks/useBalance.ts
```diff
@@ -0,0 +1,105 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
```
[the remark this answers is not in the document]
wirednkod: I'll fix it here on this PR

src/hooks/useBalance.ts
```diff
@@ -0,0 +1,89 @@
...
+import BN from 'bn.js';
+import { useEffect, useState } from 'react';
+
+import { useSourceTarget } from '../contexts/SourceTargetContextProvider';
+
+type State = [string, Balance, boolean, string];
+
+const ZERO = new BN(0);
+
+const useBalance = (address: string, providedSi: boolean = false): State[] => {
```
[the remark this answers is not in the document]
tomusdrw: > In this case the logic to show the balances is the following:

I'm pretty sure this is incorrect. The entire codebase should not care what kind of account is selected. There is always only:
- `source` chain - the one we are going to send the transaction to
- `target` chain - the one that will have a message delivered and will perform dispatch.

When user selects an account in the dropdown, the correct `source/target` variables should be set accordingly, so when user selects Rialto:Alice, the `source` chain should be pointing to `Rialto`. When the user selects `Millau:Alice` the `source` chain should be pointing to `Millau`. Selecting accounts from different networks should cause the UI to switch `source/target` pair correctly.

src/components/Accounts.tsx
```diff
@@ -53,8 +53,8 @@ const Accounts = ({ className }: Props) => {
   const onChange = (value: string, chain: string) => {
     setCurrentAccount(value, chain);
   };
-
   const renderAccounts = (chain: string) => {
+    console.log('chain is:', chain);
```
tomusdrw, line 57: stray log

src/hooks/useBalance.ts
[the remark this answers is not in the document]
tomusdrw: I think I've jumped to conclusion too early. So the code seems correct now and the logic described here seems correct too. I think the complication comes from the way we use hooks and how we treat accounts.

I wonder if we could always associate accounts with their chains. I.e. every time you receive `account` it's not only just account id, but it actually always have an `api` or at least `chainName` associated with it, so that if you use any util functions there is no confusion what api / chain to use.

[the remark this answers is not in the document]
hbulgarini: The logic i described is correct i think , because the purpose of this is to show as well the balances within the account select component. If you check the designs you will see that you have the accounts from both chains with their respective derived account balance. 
For the sake of consistency and how the code was built, when initially the select is open there is already a source and target defined.

src/components/Accounts.tsx
wirednkod: Its not stray - Its called Jack and its my console log - I pet it and feed it 😄

src/hooks/useBalance.ts
```diff
@@ -0,0 +1,107 @@
...
+        return Promise.resolve(u);
+      } catch (e) {
+        dispatchMessage(MessageActionsCreators.triggerErrorMessage({ message: e.message }));
+        logger.error(e.message);
+        return Promise.reject();
+      }
+    };
+
+    if (address) {
+      console.log('name: ', name, 'address', address, ' => derievedAddress', derivedAddress);
```
hbulgarini, line 94: If this is something you want to log, please use the `logger` helper

hbulgarini (approved):
LGTM !  There is one small comment to address but you can merge as soon as it is addressed.

hbulgarini (approved):
LGTM ! There is one small comment to address but you can merge as soon as it is addressed.

Issue #22, opened 2021-03-24
Account derivation viewing screen (paste address and see possible options).


tomusdrw:
From comment: https://github.com/paritytech/parity-bridges-ui/pull/20/files/92bab7b41ffd9ce2d2dd9ddf78d40d04cb0703e5..7cb96f6a75d266f5b2b211caaa713c83b379b7e8#r599534736

> We should be more informative with this. I.e. we should parse the SS58 prefixes and display things like: "This address looks like a Polkadot address, not Rialto's.".
Also notice there are at least 3 different cases here:
>
> - Target chain provided - all good (receiver_address === provided_address)
> - Source chain provided - derivation (receiver_addresss === derive(provided_adddress))
> - SS58 does not match neither Target nor Source chain - a REALLY FAT warning and we should set receiver_address = provided_address (optionally with an option to derive account).
>
> This is actually extremely crucial, so let's have this as a separate issue and make it work extremely well.

PR #79, opened 2021-04-21
Account derivation feedback

hbulgarini:
PR related to issue : #22


src/App.tsx
```diff
@@ -26,7 +26,7 @@ import { MessageContextProvider } from './contexts/MessageContext';
 import { SourceTargetContextProvider } from './contexts/SourceTargetContextProvider';
 import { TransactionContextProvider } from './contexts/TransactionContext';
 import { useApiConnection } from './hooks/useApiConnection';
-import Main from './screens/Main';
+import Main from './components/Main';
```
wirednkod, line 29: I would not move the Main page inside the components. I normally prefer a distinguish between reusable components and page-components.

src/actions/accountActions.ts
```diff
@@ -17,16 +17,30 @@
 import type { Account } from '../types/accountTypes';
 
 enum AccountActionsTypes {
-  SET_ACCOUNT = 'SET_ACCOUNT'
+  SET_ACCOUNT = 'SET_ACCOUNT',
+  SET_DERIVED_ACCOUNT = 'SET_DERIVED_ACCOUNT ',
```
wirednkod, line 21: are the spaces at the end of each string (SET_DERIVED_ACCOUNT and SET_GENERIC_ACCOUNT intentional? looks like a typo

src/actions/transactionActions.ts
```diff
@@ -21,15 +21,18 @@ enum TransactionActionTypes {
   SET_RECEIVER_ADDRESS = 'SET_RECEIVER_ADDRESS',
   CREATE_TRANSACTION_STATUS = 'CREATE_TRANSACTION_STATUS',
   UPDATE_CURRENT_TRANSACTION_STATUS = 'UPDATE_CURRENT_TRANSACTION_STATUS',
-  SET_TRANSACTION_COMPLETED = 'SET_TRANSACTION_COMPLETED'
+  SET_TRANSACTION_COMPLETED = 'SET_TRANSACTION_COMPLETED',
+  SET_RECEIVER_VALIDATION = 'SET_RECEIVER_VALIDATION',
+  SET_DERIVED_RECEIVER_ACCOUNT = 'SET_DERIVED_RECEIVER_ACCOUNT ',
```
wirednkod, line 26: Again some spaces at the end of each string. Typo?

src/components/AccountDisplay.tsx
```diff
@@ -0,0 +1,103 @@
...
+    if (!accountName) {
+      return shorterItem(address);
+    }
+    return accountName;
+  };
+
+  if (fromSender) {
+    return (
+      <Container onClick={onClick} className={classes.onlyBalance}>
+        <p>{balance ? balance : '-'}</p>
```
wirednkod, line 83: I'd go with this:
```
<p>{balance || '-'}</p>
```

src/components/AccountDisplay.tsx
```diff
@@ -0,0 +1,103 @@
...
+  return (
+    <Container onClick={onClick} className={classes.container}>
+      <div className={classes.icon}>
+        <AccountIdenticon address={address} />
+      </div>
+      <div className={classes.address}>
+        <p>{displayText()}</p>
+      </div>
+      <div className={classes.balances}>
+        <p>{balance ? balance : '-'}</p>
```
wirednkod, line 97: I'd go with this:
```
<p>{balance || '-'}</p>
```

src/hooks/useBalance.ts
```diff
@@ -41,33 +39,12 @@ const initValues = {
   free: new BN(ZERO) as Balance
 };
 
-const useBalance = (name: string, address: string, chain: string | undefined, providedSi: boolean = false): State[] => {
+const useBalance = (api: ApiPromise, address: string, providedSi: boolean = false): State => {
```
wirednkod, line 42: I do not understand why to remove the double functionality of useBalance that can be very helpful in a project like bridges. Its more convenient to receive both amounts and chose which one to use than calling it twice. Or am I missing something?

src/types/accountTypes.ts
```diff
@@ -25,7 +25,7 @@ export interface AccountContextType {
 }
 
 interface Payload {
-  [propName: string]: Account;
+  [propName: string]: any;
```
wirednkod, line 28: I would avoid using any if possible

src/util/getReceiverAddress.ts
```diff
@@ -25,21 +26,38 @@ interface Props {
...
-      address = formattedAccount;
+      return { address: formattedAccount, formatFound };
     }
-    return address;
+    return { address, formatFound };
   } catch (e) {
-    throw new Error('INCORRECT-FORMAT');
+    if (receiverAddress) {
+      throw new Error(INCORRECT_FORMAT);
+    }
```
wirednkod, line 59: You should have also a GENERIC error (in case if  (!receiverAddress)

wirednkod (changes requested):
I would prefer to merge this PR **after** PR #96 if possible

src/hooks/useBalance.ts
hbulgarini: the reason why i dissected the logic to have specific calls per chain is because this functionality was quite tight to the select account component use case when you have the input  [addrress, chain], and you get the balances for the account and derived account.  In the `Transfer` component i realized that  i needed this logic extracted so i only have one account with one balance. 
In the previous version `Account` was retrieving always two rows: Account and derived  and in transfers sometimes i just need only one.

wirednkod (approved):

src/hooks/usePrevious.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { useRef, useEffect } from 'react';
+
+export default function usePrevious(value: any) {
+  const ref = useRef();
+  useEffect(() => {
+    ref.current = value;
+  });
+  return ref.current;
+}
```
tomusdrw, line 25: This thing looks super hacky, why do we `useRef`? Can't we have an explicit state with previous value (cause I believe that's what this thing is supposed to do).

src/util/getReceiverAddress.ts
```diff
@@ -25,21 +26,38 @@ interface Props {
 const getReceiverAddress = ({ receiverAddress, chain }: Props) => {
   const chainsConfigs = getChainConfigs();
   const { SS58Format, bridgeId } = chainsConfigs[chain];
+
   try {
-    const [validatedDerivedAcccount] = checkAddress(receiverAddress, SS58Format);
-    let address = receiverAddress;
+    const [validatedDerivedAcccount, rest] = checkAddress(receiverAddress, SS58Format);
```
tomusdrw, line 31:
```suggestion
 const [validatedDerivedAcccount, rest] = checkAddress(receiverAddress, SS58Format);
    const address = receiverAddress;
    if (validatedDerivedAcccount) {
       return { address, formatFound: chain };
    }
    // should be extracted as a separate component/function
    const getFormat = (prefix: string) => {
       const intPrefix: number = parseInt(prefix, 10);
       if (intPrefix === GENERIC_SUBSTRATE_PREFIX /* 42 */) {
          return GENERIC;
       }
       const chainsConfigs = getChainConfigs();
       return Object.keys(chainsConfigs).find((key) => chainsConfigs[key].SS58Format === intPrefix);
    }
    
    const parts = rest?.split(',');
    const prefix = parts![2].split(' ');
    const formatFound = getFormat(prefix);
    
    const address = getDeriveAccount({
        SS58Format,
        address,
        bridgeId
      });

      return { address, formatFound };
    }
```

The whole function could be made more readable.

src/components/Receiver.tsx
```diff
@@ -0,0 +1,79 @@
...
+  } = useSourceTarget();
+
+  return (
+    <Container className={classes.container}>
+      <div className={classes.input}>
+        <ReceiverInput setError={setError} />
+      </div>
+      {!genericReceiverAccount && (
+        <div className={classes.derived}>
+          {derivedReceiverAccount && <Account value={derivedReceiverAccount} chain={targetChain} isDerived />}
```
tomusdrw, line 63: IMHO to unify the way accounts are displayed we should always display the bottom part. In case it's just target chain format we can repeat the address.

tomusdrw: Also right now we don't handle correctly accounts with prefixes being neither source, target or generic.

src/components/ReceiverInput.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+    if (formatFound === GENERIC) {
+      dispatchTransaction(TransactionActionCreators.setGenericAccount(receiver));
+      return;
+    }
+
+    if (formatFound === targetChain) {
+      setReceiver(formattedAccount);
+    } else {
+      dispatchTransaction(TransactionActionCreators.setDerivedAccount(formattedAccount));
```
tomusdrw, line 115: We shouldn't be assuming that it's sourceChain account if `formatFound` is neither `GENERIC` nor `targetChain`.

tomusdrw (changes requested):
It's much better now, but there is a lot of details that need improvement. Some of them are minor things, but some are pretty critical.
The components logic seems to be overcomplicated as well and when I tested the UI I can get it into weird states, like:
derived accounts being clearly wrong (either in sender selector or receiver selector):
image

Clicking "Native" ends up with "derived" (so does clicking "Companion"):
image

Address being neither source nor target nor generic - no idea what address will receive funds:
image

Derived DAVE showing 0 balance after selection ,but has balance durign the selection:
image
image

Able to send transaction with address being detected as invalid (balance can be left empty too).
image

src/hooks/usePrevious.ts
wirednkod: Actually this hook is the best way to retrieve the previous value without having a second state to keep the value (plus its reusable for multiple components - so I would leave it as is.

wirednkod (approved):

src/components/Receiver.tsx
hbulgarini: actually that brought some hard restrictions when i needed to implement the receiver component. Having a component that always retrieve two lines makes difficult to implement different use cases as the generic case for example.

src/components/ReceiverInput.tsx
hbulgarini: why not? what are the other options then?

src/components/Sender.tsx
```diff
@@ -30,78 +33,112 @@ interface Props {
...
   };
 
+  const getName = (account: AccountType) => (account!.meta.name as string).toLocaleUpperCase();
+
   const AccountSelected = () => {
     if (account) {
-      const text = (account.meta.name as string).toLocaleUpperCase();
-      return <Account text={text} value={value} />;
+      const text = getName(account);
+      return <Account accountName={text} value={value} chain={sourceChain} />;
```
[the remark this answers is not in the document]
hbulgarini: i think this can be addressed after ... do you see it as a hard requirement?

src/hooks/usePrevious.ts
hbulgarini: yes, actually this is quite used and there is a lot of  examples out there.

src/components/ReceiverDerivedAccount.tsx
```diff
@@ -0,0 +1,63 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+
+import React from 'react';
+import { makeStyles } from '@material-ui/core/styles';
+import { useSourceTarget } from '../contexts/SourceTargetContextProvider';
+import { useTransactionContext } from '../contexts/TransactionContext';
+import Account from './Account';
+
+//import GenericAddress from './GenericAccount';
```
tomusdrw, line 23: Let's remove if unused.

src/components/GenericAccount.tsx
```diff
@@ -0,0 +1,107 @@
...
+  },
+  companion: {
+    border: '1px solid',
+    borderTop: 'none',
+    borderRadius: '0 0 5px 5px',
+    padding: '5px 0'
+  }
+}));
+
+const GenericAccount = ({ value }: Props) => {
```
tomusdrw, line 53: Please remove if this is unused.

src/components/AccountDisplay.tsx
```diff
@@ -0,0 +1,97 @@
...
+    return shorterItem(address);
+  };
+
+  if (fromSender) {
+    return (
+      <Container onClick={onClick} className={classes.onlyBalance}>
+        <p>{balance || '-'}</p>
+      </Container>
+    );
+  }
```
tomusdrw, line 80: Seems like two completelty different components to me (balance display & account display), but let's leave it like that for now.

src/components/AccountDisplay.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  return (
+    <Container onClick={onClick} className={classes.container}>
```
tomusdrw, line 83: I'd add a `title/tooltip` with full account address, can be left for separate PR though.

src/components/ReceiverInput.tsx
tomusdrw: Yes, there is 16k possible SS58 prefix values, and we should only handle 3 of them: `42 (generic)`, source, target.

tomusdrw: Still something to fix.

src/components/Sender.tsx
```diff
@@ -30,78 +33,98 @@ interface Props {
...
         value={value}
         onClick={() => {
-          onChange(value, chain);
+          onChange(value, source);
         }}
       >
-        <Account text={text} value={value} showDerivedBalance chain={chain} />
+        <div className={classes.row}>
+          <Account accountName={text} value={value} chain={source} />
+          <Account accountName={text} value={value} isDerived fromSender chain={target} />
```
tomusdrw, line 71: `fromSender` just displays balance, why bother with name and `isDerived`?

src/components/Sender.tsx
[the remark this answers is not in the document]
tomusdrw: Sure, please create an issue for this.

src/components/Transfer.tsx
```diff
@@ -50,46 +57,25 @@ const Transfer = ({ className }: Props) => {
...
-        <p>{receiverAddress && estimatedFee && `Estimated source Fee: ${estimatedFee}`}</p>
-      </Container>
-    </>
+
+      <Receiver />
+
+      <TextField onChange={onChange} value={transferInput} label="Amount" variant="outlined" />
+
+      <Button variant="contained" disabled={isButtonDisabled()} onClick={sendLaneMessage}>
+        Transfer from {sourceChain} to {targetChain}
```
tomusdrw, line 73:
```suggestion
       Send Bridge Message 
```
Current label is super confusing, the transfer is NOT between networks.

src/hooks/useApiBalance.ts
```diff
@@ -0,0 +1,60 @@
...
+  const { SS58Format } = chainsConfigs[chain === targetChain ? targetChain : sourceChain];
+  const { bridgeId } = chainsConfigs[chain === targetChain ? sourceChain : targetChain];
+  let addressResult = address;
+  if (isDerived) {
+    addressResult = getDeriveAccount({
+      SS58Format,
+      address,
+      bridgeId
+    });
+  }
```
tomusdrw, line 53:
```suggestion
  const addressResult = !isDerived ? address : getDeriveAccount({
      SS58Format,
      address,
      bridgeId
    });
```

src/util/getReceiverAddress.ts
```diff
@@ -25,21 +26,38 @@ interface Props {
...
+  try {
+    const [validatedDerivedAcccount, rest] = checkAddress(receiverAddress, SS58Format);
+    if (validatedDerivedAcccount) {
+      return { address: receiverAddress, formatFound: chain };
     }
-    return address;
+    // should be extracted as a separate component/function
+    const getFormat = (prefix: string) => {
+      const intPrefix: number = parseInt(prefix, 10);
+      if (intPrefix === GENERIC_SUBSTRATE_PREFIX /* 42 */) {
```
tomusdrw, line 38:
```suggestion
      if (intPrefix === GENERIC_SUBSTRATE_PREFIX) {
```
I was just suggesting that the constant should contain that value.

src/components/AccountIdenticon.tsx
```diff
@@ -0,0 +1,41 @@
...
+import ctx from 'classnames';
+import React from 'react';
+interface Props {
+  address: string;
+  formatFound?: string;
+}
+
+const useStyles = makeStyles(() => ({
+  placeholder: {
+    opacity: '30%'
```
tomusdrw, line 29: I still think the placeholder should be grayscale to make it look even less like actual identicon.
```suggestion
    opacity: '30%', // as a fallback if filter is not supported
    filter: 'grayscale(1)'
```

src/components/ReceiverDerivedAccount.tsx
```diff
@@ -0,0 +1,63 @@
...
+    /*   return <GenericAddress value={genericReceiverAccount} />; */
+  }
+  if (derivedReceiverAccount) {
+    return (
+      <div className={classes.derived}>
+        <Account value={derivedReceiverAccount} chain={targetChain} isDerived />
+      </div>
+    );
+  }
+  return null;
```
tomusdrw, line 60: IMHO to unify the display of the component we should always render the bottom part and just repeat the same stuff that is in the input field, but that's more up to @goldsteinsveta. AFAIR on designs we do display `raw/derived` always and while I already mentioned I think we can still improve on naming, but at least presentation wise I think it was good.

tomusdrw (approved):
looks much better now! A bunch of smaller things to address, but can be merged after that.

src/components/Transfer.tsx
hbulgarini: it was actually on the design ... i can change it

src/components/GenericAccount.tsx
hbulgarini: i prefer this component to stay even if it is not used. Removing it will lead us to work on something that needs adjustments not recreating the component from scratch.

src/components/ReceiverDerivedAccount.tsx
hbulgarini: i prefer this component to stay even if it is not used. Removing it will lead us to work on something that needs adjustments not recreating the component from scratch.

src/components/GenericAccount.tsx
tomusdrw: No, please remove. It will make it harder to review in the future and dead code should not be in the repository.

src/components/Sender.tsx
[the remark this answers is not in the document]
hbulgarini: https://github.com/paritytech/parity-bridges-ui/issues/103

src/components/AccountDisplay.tsx
hbulgarini: Right , it was not hard to create it now.

src/components/AccountDisplay.tsx
hbulgarini: https://github.com/paritytech/parity-bridges-ui/issues/104

src/components/ReceiverInput.tsx
hbulgarini: Done.

src/components/Sender.tsx
hbulgarini: Completely removed.

hbulgarini (commented):
In the last commits i'm addressing all your comments and also i'm solving the issue with the derived account logic.

src/components/Sender.tsx
```diff
@@ -30,78 +33,112 @@ interface Props {
...
-          <Account text="Derived Account" value={derivedAccount} />
-        </div>
-      )}
+      <InputLabel className="label" id="demo-simple-select-label">
+        {sourceChain} Account
+      </InputLabel>
+      <div className="senderSelect">
+        <FormControl className="formControl">
+          <Select value={value} name="name" renderValue={(): React.ReactNode => <AccountSelected />}>
+            {chains.map((chain) => renderAccounts(chain))}
```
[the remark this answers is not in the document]
hbulgarini: Your code has an issue and this is why i reverted it to the original one: 

When you flip the chains because a target account was selected, the order in the dropdown is flipped as well.

hbulgarini:
know issue: https://github.com/paritytech/parity-bridges-ui/issues/105

hbulgarini:
> It's much better now, but there is a lot of details that need improvement. Some of them are minor things, but some are pretty critical.
> The components logic seems to be overcomplicated as well and when I tested the UI I can get it into weird states, like:
> derived accounts being clearly wrong (either in sender selector or receiver selector):
> image
> 
> Clicking "Native" ends up with "derived" (so does clicking "Companion"):
> image
> 
> Address being neither source nor target nor generic - no idea what address will receive funds:
> image
> 
> Derived DAVE showing 0 balance after selection ,but has balance durign the selection:
> image
> image
> 
> Able to send transaction with address being detected as invalid (balance can be left empty too).
> image

In the PR i have solved already in this PR the following issues:

- Incorrect derived accounts (now all derived accounts shows the same as the CLI)
- Balances reflected as part of the fix described above.
- GenericAccount component removed and to be implemented in issue #101 
- Major improvements / simplification to components.

PR #81, opened 2021-04-22
Refactor code to support import tree shaking

### LINKED ISSUE #73: Babel configuration to permit three shaking on imports.
Babel configuration to permit three shaking on imports. 

Discussion started on: https://github.com/paritytech/parity-bridges-ui/pull/69/files#r616063656

wirednkod:
There was no need for babel configuration - works out of the box with create-react-app.
Fixes #73


src/components/Transactions.tsx
```diff
@@ -25,7 +25,7 @@ import { useTransactionContext } from '../contexts/TransactionContext';
 import { useUpdateTransactionContext } from '../contexts/TransactionContext';
 import { TransactionStatusEnum, TransanctionStatus } from '../types/transactionTypes';
 import shortenItem from '../util/shortenItem';
-import TransactionStatus from './TransactionStatus';
+import { TransactionStatus } from '.';
```
tomusdrw, line 28: I feel the `index.ts` files are unnecessary and are just boilerplate to maintain. I'm also quite reluctant to believe that introducing an extra "hub" file helps with tree shaking in any way.

src/components/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+// (at your option) any later version.
+//
+// Parity Bridges UI is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+export { default as Account } from './Account';
```
tomusdrw, line 16: IMHO the entire file doesn't really add value.

src/hooks/useConnectedReceiver.ts
```diff
@@ -49,3 +49,5 @@ export default function useConnectedReceiver(): Function {
 
   return setConnectedReceiver;
 }
+
+export default useConnectedReceiver;
```
tomusdrw, line 53: IMHO this isn't really better than previous, more concise code.

src/hooks/useMessagesLane.ts
```diff
@@ -18,7 +18,7 @@ import { ApiPromise } from '@polkadot/api';
 import BN from 'bn.js';
 import { useEffect, useState } from 'react';
 
-import useLaneId from '../hooks/useLaneId';
+import { useLaneId } from '../hooks';
```
tomusdrw, line 21: Again, I'm not sure what the gain actually is. Like I can accept if we decide to never use default exports (I'm a big fan of that approach actually), but having "hub" files is not really something that I fully understand. Is it about saving few keystrokes to avoid repeating `useLaneId` on both sides?

tomusdrw (changes requested):
I'm not entirely sure about the gains here. I fully support bracket-importing from external libraries, but I'm not fully sold on the extra maintenance costs of `index.ts` files internally in seemeangly arbitrary places (why `import { useLaneId } from ../hooks/` and not `import { useLaneId } from ../` or `import { hooks: { useLaneId } } from '../'`.

src/components/index.tsx
wirednkod: The actual case is that it makes the export of components to be done straight from the components dir. so instead of:
`import Account from './components/Account';`
it helps to change to:
`import { Account } from './components';`
The example above may look not-much-diff but in case of multiple imports, that reduces the lines and readability.

If though this is not a preferable way I can just rollback - just let me know.

src/components/Transactions.tsx
wirednkod: The actual case is that it makes the export of components to be done straight from the components dir. so instead of:
`import Account from './components/Account';`
it helps to change to:
`import { Account } from './components';`
The example above may look not-much-diff but in case of multiple imports, that reduces the lines and readability.

If though this is not a preferable way I can just rollback - just let me know.

src/hooks/useMessagesLane.ts
wirednkod: It mainly comes down to this:
From:
```
import Accounts from '../components/Accounts';
import CustomCall from '../components/CustomCall';
import DashboardCard from '../components/DashboardCard';
import Remark from '../components/Remark';
import SnackBar from '../components/SnackBar';
import Transactions from '../components/Transactions';
import Transfer from '../components/Transfer';
```
To:
```
import { Accounts, CustomCall, DashboardCard, Remark, SnackBar, Transactions, Transfer } from '../components';
```

It is not a matter of right or wrong - but mostly of a decision making in the team on what approach is preferable
At the end of the day you are right and I am right and Hector is right :)

PR #82, opened 2021-04-22
Add detection of extensions and accounts

wirednkod:


.env
```diff
@@ -1,10 +1,10 @@
 REACT_APP_CHAIN_1=Rialto
 REACT_APP_CHAIN_2=Millau
-REACT_APP_SUBSTRATE_PROVIDER_CHAIN_1=wss://wss.rialto.brucke.link
-REACT_APP_SUBSTRATE_PROVIDER_CHAIN_2=wss://wss.millau.brucke.link
+REACT_APP_SUBSTRATE_PROVIDER_CHAIN_1=ws://127.0.0.1:9944
```
hbulgarini, line 3: This file shouldn't be changed.

Please use `.env.local` to have your own links.

src/contexts/KeyringContextProvider.tsx
```diff
@@ -39,16 +38,27 @@ export function KeyringContextProvider(props: KeyringContextProviderProps): Reac
...
+
+        if (extExists.length === 0) {
+          // no extension installed, or the user did not accept the authorization
+          // in this case we should inform the use and give a link to the extension
+          return;
+        } else {
+          setExtensionExists(true);
+        }
         let allAccounts = await web3Accounts();
+        allAccounts?.length > 0 && setAccountExists(true);
```
hbulgarini, line 61: minor: i think the `> 0` is not needed since lenght equal to zero would return false anyway

src/screens/Main.tsx
```diff
@@ -35,8 +36,23 @@ interface Props {
 }
 
 export function Main({ className }: Props) {
+  const { extensionExists, accountExists } = useKeyringContext();
   const { sourceChainDetails, targetChainDetails } = useSourceTarget();
 
+  const proj = [];
```
hbulgarini, line 42: i would extract this logic to a separate component so is easier as well to apply designs

src/screens/Main.tsx
```diff
@@ -58,7 +74,7 @@ export function Main({ className }: Props) {
         </Grid>
         <Grid container>
           <Grid item md={12}>
-            <Accounts />
+            {proj}
```
hbulgarini, line 77: this would replace can be replaced as `<NewComponent />`

hbulgarini (changes requested):

.env
wirednkod: That was a draft PR - :)

src/screens/Main.tsx
wirednkod: I will not do that at this point - Already talked with @goldsteinsveta and agreed to prepare the ground until Figma is ready - and then we will start putting in the correct position.

src/screens/Main.tsx
wirednkod: Agreed - will do that in later PR when Figma is prepared and we start building the actual compoentns

PR #83, opened 2021-04-23
UI: Base Layout

goldsteinsveta:
#47 fixed

Screenshots, highlighted:
Screenshot 2021-04-23 at 17 53 10Screenshot 2021-04-23 at 17 53 21


src/components/theme.tsx
```diff
@@ -0,0 +1,179 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
```
hbulgarini, line 1: Is this file necessary to be created as `.tsx` ? i feel this can be a `.ts` file which can live in a different folder than components. For example: `./src/styles` ? In case you want to leave it as a component, please rename the file as `Theme.tsx`

src/screens/Main.tsx
```diff
@@ -39,50 +40,52 @@ export function Main({ className }: Props) {
 
   return (
     <>
```
hbulgarini, line 42: we can remove this since `BoxMain` enclose all

hbulgarini (changes requested):

hbulgarini:
@goldsteinsveta it seems we have some conflicts with App / Maint.tsx please let mw know if you need help for solving them.

src/components/theme.tsx
wirednkod: Not really agree with that. All files that contain JSX elements should be TSX while pure Typescript files should be TS.

wirednkod (approved):
I agree with @hbulgarini on the removal of <></>. Other than that PR lgtm

src/components/theme.tsx
hbulgarini: right, this is why i asked first to rename it to ".ts" . Anyway i left the chance to rename it as component because i was not sure if sveta was planning to introduce JSX code in it. In case she does not, then rename it to ".ts" is the right thing to do.

wirednkod: yeap - my mistake

src/screens/Main.tsx
goldsteinsveta: done

src/components/theme.tsx
goldsteinsveta: done, it's .ts now

wirednkod (approved):

hbulgarini (approved):
LGTM!

Issue #84, opened 2021-04-23
Ensure that `.env` file is present and is being validated by unit test.


tomusdrw:

hbulgarini:
@wirednkod  assigned to you as requested.

PR #91, opened 2021-04-25
Add tests for env file and contents

wirednkod:


wirednkod:
Closes #84

hbulgarini:
I don't see any problems testing an `.env` file. If during the build we copy the corresponding env.<branch> to the `.env` this approach might work. I would like to know what @tomusdrw  thinks about.

wirednkod:
The point discussed with @tomusdrw was to make sure through unit tests - that the .env file exists and (I extended it) check that all values inside exist. In case we follow the process that you propose - that means that no .env file should be committed and we need to create for each env, one env.<branch> file.  Im fine with both

src/util/tests/envVariablesValidations.test.ts
```diff
@@ -54,3 +56,42 @@ describe('envVariablesValidations', () => {
...
+  test('REACT_APP_SS58_PREFIX_CHAIN_2 environmental variable should exist and have a value', () => {
+    expect(envVars?.REACT_APP_SS58_PREFIX_CHAIN_2).toBeTruthy();
+  });
+  test('REACT_APP_BRIDGE_ID_CHAIN_1 environmental variable should exist and have a value', () => {
+    expect(envVars?.REACT_APP_BRIDGE_ID_CHAIN_1).toBeTruthy();
+  });
+  test('REACT_APP_BRIDGE_ID_CHAIN_2 environmental variable should exist and have a value', () => {
+    expect(envVars?.REACT_APP_BRIDGE_ID_CHAIN_2).toBeTruthy();
+  });
+});
```
tomusdrw, line 97: We already maintain a list of expected variables in `envVariablesValidations` file. Why can't we simply call the function that validates env in this test?
I'm against duplicating the names.

tomusdrw (changes requested):

src/util/tests/envVariablesValidations.test.ts
wirednkod: @tomusdrw The tests above are checking that the variables exist in the actual .env file (so if by mistake I delete 1 variable and it does NOT exist inside the .env file - test will fail)

In the `envVariablesValidations` file the process.env is created on the spot (instead of being imported from the actual .env file). 

You think that importing the .env and testing against that is over-testing?

tomusdrw: > In the envVariablesValidations file the process.env is created on the spot (instead of being imported from the actual .env file).

Not sure what you mean, `checkExpectedVariables` is simply going through the pre-defined list of variables and checking that they are available in `process.env`.

> You think that importing the .env and testing against that is over-testing?

Nope, I think we should import `.env` file but just use `checkExpectedVariables` function instead of manually listing all env variables once again.

From `dotenv` docs:
> config will read your .env file, parse the contents, assign it to process.env, and return an Object with a parsed key containing the loaded content or an error key if it failed.

So by calling `config()` you are already importing the variables to `process.env` the only change that is required is the assertions, basically the whole code should be just:
```suggestion
  test('expected ENV variables are set', () => {
    expect(checkExpectedVariables()).toBeTruthy();
  });
});
```

There is still opportunity of one more test, where we check that `.env` file does not have any unexpected variables, but that's not that much important and can be postponed until we don't have anything better to do :)

src/util/envVariablesValidations.ts
```diff
@@ -36,6 +36,15 @@ const checkExpectedVariables = () => {
   return true;
 };
 
+const checkUnexpectedVariables = () => {
+  for (const v of Object.keys(process.env)) {
+    if (!expectedVariables.includes(v)) {
+      throw new Error(`Unexpected ${v} variable found.`);
+    }
```
tomusdrw, line 43: I'm not sure if this is the right way to solve the issue, wouldn't `process.env` contain some random stuff? Or are the system environment variables filtered (i.e. only `REACT_APP` variables are getting in)? If the latter then it's good :)

wirednkod: This test ensures that all variables set in .env file provided in the repo will be test agains (so the `REACT_APP` variables). I can always update to check for the specific ones onl y (Start with REACT_APP) but i think as soon as the process.env is loaded inside the scope of this repo - no random stuff should exist

tomusdrw (approved):

PR #92, opened 2021-04-25
fix error that was appearing in console

wirednkod:
fix error that was appearing in console ("must have keys") by simplifying the code

Screenshot 2021-04-25 at 6 18 13 PM


hbulgarini (approved):
LGTM!

Issue #56, opened 2021-04-16
Support canceling from the extension.


hbulgarini:
When a transaction is cancelled from the extension , the UI still fire a transaction which of course it is wrong.

PR #93, opened 2021-04-25
Fix cancel not to show transaction being fired and pop correct cancel…

wirednkod:
… message

### DIFF
--- a/src/hooks/useSendMessage.ts
+++ b/src/hooks/useSendMessage.ts
@@ -65,7 +65,6 @@ function useSendMessage({ isRunning, isValidCall, setIsRunning, input, type, wei
   };
 
   const makeCall = async (id: string) => {
-    let startProcess: boolean = true;
     try {
       if (!account || isRunning) {
         return;
@@ -83,7 +82,7 @@ function useSendMessage({ isRunning, isValidCall, setIsRunning, input, type, wei
         sourceAccount = account.address;
       }
       const unsub = await bridgeMessage.signAndSend(sourceAccount, { ...options }, ({ events = [], status }) => {
-        if (startProcess) {
+        if (status.isReady) {
           dispatchTransaction(
             TransactionActionCreators.createTransactionStatus({
               block: null,
@@ -99,7 +98,6 @@ function useSendMessage({ isRunning, isValidCall, setIsRunning, input, type, wei
               type
             })
           );
-          startProcess = false;
         }
         if (status.isBroadcast) {
           dispatchMessage(MessageActionsCreators.triggerInfoMessage({ message: 'Transaction was broadcasted' }));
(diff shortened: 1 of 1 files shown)


wirednkod:
Fixes #56

src/hooks/useSendMessage.ts
```diff
@@ -96,8 +82,25 @@ function useSendMessage({ isRunning, isValidCall, setIsRunning, input, type, wei
         options.signer = injector.signer;
         sourceAccount = account.address;
       }
-
       const unsub = await bridgeMessage.signAndSend(sourceAccount, { ...options }, ({ events = [], status }) => {
+        if (startProcess) {
```
hbulgarini, line 86: i would like to know if there is any internal state in the callback similar as `status.isInBlock` to check if the transaction was signed. In case not , i'm fine with this ( i does the job and solves the bug )

hbulgarini (approved):
I left a comment which is not a blocker for merge.

wirednkod:
@hbulgarini That was the first thing I checked but I did not find a relevant status. The existing enum contains:
```
   readonly isFuture: boolean;
    readonly isReady: boolean;
    readonly isBroadcast: boolean;
    readonly asBroadcast: Vec<Text>;
    readonly isInBlock: boolean;
    readonly asInBlock: Hash;
    readonly isRetracted: boolean;
    readonly asRetracted: Hash;
    readonly isFinalityTimeout: boolean;
    readonly asFinalityTimeout: Hash;
    readonly isFinalized: boolean;
    readonly asFinalized: Hash;
    readonly isUsurped: boolean;
    readonly asUsurped: Hash;
    readonly isDropped: boolean;
    readonly isInvalid: boolean;
```
Now while testing (cancel the sign) I did not see any information returning with specific status (which makes sense - cause as I have understood - the extension is a mediator before reaching the rpcs)

src/hooks/useSendMessage.ts
tomusdrw: I think there is `status.isReady || status.isFuture` that can be used for that.

Also it seems we are not handling an error scenario at all here (netiher pool inclusion nor 3rd argument to the callback, namely `dispatchError`), but I guess that's separate issue.

wirednkod:
@tomusdrw I have tested the error scenario (e.g. not enough balance) and you can see it below.

Screenshot 2021-04-26 at 12 48 14 PM

Im not sure though - if that is what you meant

tomusdrw:
That's great! Then we are sure that the transaction is at least included in the transaction pool and it's good enough for now.

In the future there are two other error conditions that we might want to handle:
1. The transaction is not included in the block for a long time (or removed from the pool).
2. The transaction is included in block, but it fails to dispatch `dispatchError`.

Added a separate issue for these.

PR #97, opened 2021-04-27
UI sidebar / dashboard

### LINKED ISSUE #70: Apply design to dashboard cards
Move the existing top source / target dashboard to the right pane of the design:

Screen Shot 2021-04-20 at 15 23 46

goldsteinsveta:
closes #70 

Screenshot:
Screenshot 2021-04-27 at 12 59 39


goldsteinsveta:
hey @hbulgarini I can't wrap my mind around `useBridgeBlocks()`...it seems to not return infos in your `DashboardCard` either. Lmk if I can merge it with placeholders "real/bridged" for data as it is on this screenshot:

src/screens/Main.tsx
```diff
@@ -27,19 +36,23 @@ import Remark from '../components/Remark';
...
 }
 
 function Main({ className }: Props) {
-  const { sourceChainDetails, targetChainDetails } = useSourceTarget();
-
   return (
     <BoxMain>
-      <BoxSidebar>{`${sourceChainDetails.sourceChain} => ${targetChainDetails.targetChain}`}</BoxSidebar>
+      <BoxSidebar>
+        <div>
```
wirednkod, line 49: is this div needed at all?

wirednkod (changes requested):

src/screens/Main.tsx
goldsteinsveta: for justify space-between layout for direct children of BoxSidebar

goldsteinsveta: I'd keep it like this. Div stands for division, so it makes sense even semantically to create this separation with divs, rather than with margins. For future version there might be menu items, also in divs

wirednkod (approved):

src/components/Network.tsx
```diff
@@ -0,0 +1,113 @@
...
+        <KeyboardReturnIcon fontSize="small" />
+      </IconButton>
+      <Box p>{targetChainDetails.targetChain}</Box>
+    </Box>
+  );
+};
+
+export const NetworkStats = () => {
+  const classes = useStyles();
+  const dbSource = useDashboard(ChainDetails.SOURCE);
```
hbulgarini, line 95: may be i'm not following correctly the design ... but are we not going to show target chain details as well? In case we need to:  `NetworkStats` should receive `chain` as a prop and then call the component twice like:

```
< NetworkStats  chain={ChainDetails.SOURCE} />
< NetworkStats  chain={ChainDetails.TARGET} />
```

In case this is not what you have design , you can dismiss this comment

hbulgarini (commented):
Are we not going to show target network information?

src/components/Buttons.tsx
```diff
@@ -0,0 +1,58 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+
+import { Button, ButtonProps } from '@material-ui/core';
+import { makeStyles } from '@material-ui/core/styles';
+import ForumIcon from '@material-ui/icons/Forum';
+import React from 'react';
+
+// As this is placed as a child in the Material UI Select component, for some reason style components classes are not working.
```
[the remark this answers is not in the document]
goldsteinsveta: I don't quite see when it might inform my actions tbh. So I would not show it

src/components/Network.tsx
goldsteinsveta: @hbulgarini I do not see when the "target to source" stats might be useful, what actions this data might inform. So I would not show it. But I'm open to be proven wrong now or by user feedback/requests in the future :) However pls consider merging this first and create an issue to add the stats for the opposite direction from the state of UI.

hbulgarini (approved):

PR #107, opened 2021-04-29
UI TransactionDisplay

goldsteinsveta:
The first iteration on the message flow card (aka `TransactionDisplay`). 

Screenshot:

Screenshot 2021-04-29 at 20 16 1

Changes:

- more UI components
- Unboxed Main layout
- Tweaked TransactionTypes
- Pulled relevant for transactions (and seemingly actual) data into a sidebar and the message card, Simplified text labels from TransactionStatus


src/components/Icons.tsx
```diff
@@ -0,0 +1,37 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
```
hbulgarini, line 1: just curious: are we going to use this component to show the corresponding chain icon?

src/screens/Main.tsx
```diff
@@ -89,18 +81,7 @@ function Main({ className }: Props) {
       </BoxSidebar>
       <BoxUI>
         <MenuAction items={items} menuIdx={index} changeMenu={setIndex} />
-        <Container className={className}>
```
hbulgarini, line 92: i will miss you intent of dashboard :/

hbulgarini (approved):
The PR looks pretty good . I don't have anything to point out.
Anyway i would like to have @tomusdrw to approve this PR before it is merged since in contains some texting that i want him to be aware.

src/screens/Main.tsx
goldsteinsveta: it'll be back :) I promise!

src/components/Network.tsx
```diff
@@ -97,16 +97,21 @@ export const NetworkStats = () => {
   return (
     <>
       <Box className={classes.statsEntry}>
-        Bridged blocks:
+        Finilised blocks:
         <span>
-          {dbSource.bestBridgedFinalizedBlock} / {dbSource.bestBlockFinalized}
+          {dbSource.bestBlockFinalized} / {dbSource.bestBlock}
```
wirednkod, line 102: In all the dbsource cases below you must ensure that dbSource exists (add a ? before each dot). If it does not - then that could break the page

src/components/TransactionDisplay.tsx
```diff
@@ -14,47 +14,61 @@
...
+    },
+    '& .header': {
+      fontWeight: 500
+    }
   }
+}));
+
+export const TransactionDisplay = ({ transaction, steps }: Props) => {
+  const classes = useStyles();
+  if (!steps.length) return null;
```
wirednkod, line 47: You should always return a component and never a null. 
It is better to make the length step BEFORE rendering the component. That means outside the TransactionDisplay - before it is called... something like: 
```
!steps.length && <TransactionDisplay transaction={transaction} steps={steps} />
```

wirednkod (approved):

wirednkod (changes requested):
as per comments

src/components/Network.tsx
hbulgarini: good catch

src/components/TransactionDisplay.tsx
hbulgarini: i'm not agree with this, returning null is quite used and actually is documented 
https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

src/components/Network.tsx
hbulgarini: if you use vscode you can set to format the file during the onSave and adding the ? is one of the things that are added automatically by you.

src/components/TransactionDisplay.tsx
goldsteinsveta: ok, I'm not correcting unless it is decided otherwise

wirednkod: Just think of it this way:
Why to render (and go through the lifecycle) of a component as soon as you want to return null? - just do not render it.

Documentation you sent talks about "rare" situations (I would use it only if some internal option of the component was needed to hide it - some state change etc etc).

src/components/Network.tsx
wirednkod: I do not like to let things do automatically stuff for me - I like to have control of the code :)

src/components/TransactionDisplay.tsx
hbulgarini: i see what you mean, but anyway lifecycle components is not a such expensive process to avoid.

goldsteinsveta: I haven't corrected this for now, will wait for consensus from memory mgmt gurus! Pls approve

src/components/TransactionDisplay.tsx
```diff
@@ -14,47 +14,61 @@
...
-          <h4>Status: {status}</h4>
-          {steps.map(({ chainType, label, status }, idx) => (
-            <p key={idx}>
-              {chainType}: {label}: {status}
-            </p>
-          ))}
-        </CardContent>
+    <>
+      <ButtonSwitchMode disabled> Payload</ButtonSwitchMode>
+      <ButtonSwitchMode color="primary"> Reciept</ButtonSwitchMode>
```
tomusdrw, line 52:
```suggestion
      <ButtonSwitchMode color="primary"> Receipt</ButtonSwitchMode>
```

src/components/TransactionStatus.tsx
```diff
@@ -103,41 +103,53 @@ function TransactionStatus({ transaction, onComplete }: Props) {
...
       {
         chainType: sourceChain,
-        label: 'Waiting for block finality on source',
+        label: 'Include message in block',
+        onChain: transaction.block,
+        status: transaction.block ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS
+      },
+      {
+        chainType: sourceChain,
+        label: 'Finilise block',
```
tomusdrw, line 131:
```suggestion
        label: 'Finalise block',
```

src/components/TransactionStatus.tsx
```diff
@@ -103,41 +103,53 @@ function TransactionStatus({ transaction, onComplete }: Props) {
...
         status: completionStatus(sourceTransactionFinalized)
       },
       {
         chainType: targetChain,
-        label: 'Source Chain block finality received on target',
+        label: 'Bridge block',
```
tomusdrw, line 136:
```suggestion
        label: 'Relay block',
```
On first impression I'd prefer `Relay` rather than `Bridge`, cause so far we've never used `bridge` as a verb. Although I know it's a bit technical.

tomusdrw:
```suggestion
        label: 'Receive block',
```

I just had another idea and I think we can call this "Receive block", since the `chainType` is target, wdyt?

src/components/TransactionStatus.tsx
```diff
@@ -103,41 +103,53 @@ function TransactionStatus({ transaction, onComplete }: Props) {
...
         status: completionStatus(blockFinalityRelayed)
       },
-      { chainType: targetChain, label: 'Message delivered', status: completionStatus(messageDelivered) },
-      { chainType: targetChain, label: 'Message finality', status: completionStatus(messageFinalizedOnTarget) },
+      {
+        chainType: targetChain,
+        label: 'Message number',
```
tomusdrw, line 141:
```suggestion
        label: 'Deliver message',
```
I'd keep this in-line with other messages (`verb + what`).

src/components/TransactionStatus.tsx
```diff
@@ -103,41 +103,53 @@ function TransactionStatus({ transaction, onComplete }: Props) {
...
+        onChain: transaction.messageNonce,
```
tomusdrw, line 142: Is it actually useful to have `messageNonce` for anything? Maybe let's make it a target chain block in which the message was delivered instead?

src/components/TransactionStatus.tsx
```diff
@@ -103,41 +103,53 @@ function TransactionStatus({ transaction, onComplete }: Props) {
...
+        status: completionStatus(messageDelivered)
+      },
+      {
+        chainType: targetChain,
+        label: 'Finilise message in target block',
```
tomusdrw, line 147:
```suggestion
        label: 'Finalise message',
```
It's clear it's target from `chainType` imho.

src/components/TransactionStatus.tsx
```diff
@@ -103,41 +103,53 @@ function TransactionStatus({ transaction, onComplete }: Props) {
...
+        status: completionStatus(messageFinalizedOnTarget)
+      },
       {
         chainType: sourceChain,
-        label: 'Delivery confirmation on source',
+        label: 'Confirm delivery on source',
```
tomusdrw, line 152:
```suggestion
        label: 'Confirm delivery',
```

src/components/TransactionDisplay.tsx
goldsteinsveta: will be resolved in https://github.com/paritytech/parity-bridges-ui/issues/108

src/components/TransactionStatus.tsx
goldsteinsveta: I was torn here, relay sounds good

src/components/TransactionStatus.tsx
goldsteinsveta: makes sense

src/components/TransactionStatus.tsx
goldsteinsveta: It's kind of is. It's useful now with current sidebar where you see the number of `outboundLanes.totalMessages` that is less than `transaction.messageNonce` when transaction is included in source block. It's like a countdown to your message nonce. Pending messages value on the outbound lane does not return the number of pending messages either. It's a patch on the UI side for the values that do not return what they suppose to

goldsteinsveta: Mb this makes sense only for debugging. I've added a static string to finalised on target step. Will need to refactor to pull that value. Or perhaps @hbulgarini can help me with making TransactionStatus do what I want it to do? I.e. replace `'NNN'` with actual number of finalzed block on target

tomusdrw (approved):

goldsteinsveta:
thank you @tomusdrw

PR #109, opened 2021-04-30
Adding generic account component

hbulgarini:
This is an initial approach for the generic account behavior.
There is no UX applied for letting the user know that the native and derived accounts are selectable. @goldsteinsveta  could you please take care of it? if there is anything i can do about it in this PR to make it look better just let me know.


tomusdrw (approved):

tomusdrw:
Merging earlier, cause I intend to alter few pieces. @wirednkod, @goldsteinsveta feel free to still review and we can address fixes in follow up PR.

Issue #101, opened 2021-04-28
Solve accounts logic and derive function on receiver field.


hbulgarini:
Issue described here: https://github.com/paritytech/parity-bridges-ui/pull/79#pullrequestreview-646365419

hbulgarini:
Solved #118 & #79

Issue #100, opened 2021-04-28
Solve derivation account logic from Sender select component.


hbulgarini:
Issue described here: https://github.com/paritytech/parity-bridges-ui/pull/79#pullrequestreview-646365419

hbulgarini:
closed #118 & #106

PR #118, opened 2021-05-01
Unify account display a bit

tomusdrw:
This PR is there to make sure that a bunch of conditions are always satisfied:
1. We use consistent `native/companion` naming
2. In case of generic address the values in selection box can be related to original input (i.e. `native(<originalinput>) and companion(<originalinput>)`).
3. We don't display accounts for generic address, cause it doesn't make sense (the user haven't decided what kind of account it is yet).
image

### DIFF
--- a/src/components/Account.tsx
+++ b/src/components/Account.tsx
@@ -18,7 +18,7 @@ import React from 'react';
 
 import useApiBalance from '../hooks/useApiBalance';
 import useBalance from '../hooks/useBalance';
-import AccountDisplay from './AccountDisplay';
+import AccountDisplay, { AddressKind } from './AccountDisplay';
 
 interface Props {
   accountName?: string;
@@ -40,7 +40,7 @@ const Account = ({ accountName, value, chain, hideAddress = false, isDerived = f
       friendlyName={accountName}
       address={address}
       balance={state.formattedBalance}
-      addressKind={isDerived ? 'companion' : undefined}
+      addressKind={isDerived ? AddressKind.COMPANION : undefined}
       hideAddress={hideAddress}
     />
   );
--- a/src/components/Account.tsx
+++ b/src/components/Account.tsx
@@ -21,25 +21,27 @@ import useBalance from '../hooks/useBalance';
 import AccountDisplay from './AccountDisplay';
 
 interface Props {
-  value: string;
   accountName?: string;
-  className?: string;
   chain?: string | undefined;
+  className?: string;
+  hideAddress?: boolean;
   isDerived?: boolean;
   onClick?: () => void;
+  value: string;
 }
 
-const Account = ({ accountName, value, chain, isDerived = false }: Props) => {
+const Account = ({ accountName, value, chain, hideAddress = false, isDerived = false }: Props) => {
   const { api, address } = useApiBalance(value, chain, isDerived);
 
   const state = useBalance(api, address, true);
 
   return (
     <AccountDisplay
-      accountName={accountName}
+      friendlyName={accountName}
       address={address}
       balance={state.formattedBalance}
-      isDerived={isDerived}
+      addressKind={isDerived ? 'companion' : undefined}
+      hideAddress={hideAddress}
     />
   );
 };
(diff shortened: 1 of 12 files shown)


hbulgarini:
Is this PR related to any of the issues created lately? for example? 117

tomusdrw:
I didn't make it with any particular issue in mind by it's related (potentially closes): #101, #100 and related to #119

wirednkod (approved):

src/components/Account.tsx
```diff
@@ -21,25 +21,27 @@ import useBalance from '../hooks/useBalance';
...
-const Account = ({ accountName, value, chain, isDerived = false }: Props) => {
+const Account = ({ accountName, value, chain, hideAddress = false, isDerived = false }: Props) => {
   const { api, address } = useApiBalance(value, chain, isDerived);
 
   const state = useBalance(api, address, true);
 
   return (
     <AccountDisplay
-      accountName={accountName}
+      friendlyName={accountName}
```
hbulgarini, line 40: sincerely i don't see the benefit of this replace:  if the value comes from an `accountName` why replacing the prop and diverge the name?

src/components/Account.tsx
```diff
@@ -21,25 +21,27 @@ import useBalance from '../hooks/useBalance';
...
       address={address}
       balance={state.formattedBalance}
-      isDerived={isDerived}
+      addressKind={isDerived ? 'companion' : undefined}
```
hbulgarini, line 43: seems weird to provide `undefined ` as a value, why not null ? or handle this logic in the component?

src/components/GenericAccount.tsx
```diff
@@ -88,16 +89,29 @@ const GenericAccount = ({ value }: Props) => {
...
 
+  const shortGenericAddress = shorterItem(value);
   return (
     <Container className={classes.container}>
       {(!selected || selected === NATIVE) && (
         <div className={classes.native} onClick={setNativeAsTarget}>
-          <AccountDisplay accountName="Native" address={nativeAddress} balance={nativeState.formattedBalance} />
+          <AccountDisplay
+            address={nativeAddress}
+            addressKind="native"
```
hbulgarini, line 99: if it is a kind, it is not better to move this to a constant?

hbulgarini: even an enum ....

src/components/Account.tsx
tomusdrw: It's because `AccountDisplay` is used in different context than `Account`. In this other context (Receiver Account) we can't really associated an "accountName` with the account, cause it's not coming from keystore/extension, but we still want to provide a name that is friendlier to the user.

If you insist I'm fine renaming it back to `accountName`, but IMHO it was misleading in context where `accountName` is just made up (like short address for instance).

src/components/Account.tsx
tomusdrw: Because `addressKind` is optional (so it can be `undefined`). Passing `null` would complicate the expected type and would introduce ambiguity between `null` and `undefined` - should the behavior be different? - as a component user you have no idea if it's going to be or not.

src/components/GenericAccount.tsx
tomusdrw: Well, I thought that the reason for having TypeScript is to make sure this kind of stuff is checked.

If I define the type to be `'native' | 'companion'` then I can't make a typo anyway. Enums obviously are super cool if there amount of values is much greater, so that you don't have to repeat the type, but they also introduce boilerplate.

I'm totally fine with an enum though, sure, will update.

src/components/Account.tsx
tomusdrw: We can't handle this logic in component, cause boolean flag `isDerived` is not sufficient to express different kinds the account may have (3 possibilities + string), unless you want to have `boolean | string | null` which IMHO is way more misleading.

hbulgarini: i missed the optional definition for `addressKind ` ... makes sense....

src/components/Account.tsx
hbulgarini: that's fine, leave as `friendlyName `

hbulgarini (approved):
lgtm

PR #120, opened 2021-05-02
Fix account derivation and balance display

tomusdrw:
There was a bug in `substrate-relay` CLI and derived accounts were not generated correctly (see https://github.com/paritytech/parity-bridges-common/pull/952).
As a consequence it was implemented incorrectly in the UI as well, which this PR fixes.

Also it fixes the way balances are displayed and adds units (note the display is still a bit weird due to #118 missing).
image


wirednkod (approved):

hbulgarini (approved):
This make sense ... it used to work like this before ... 
The last time i compared the accounts with the CLI i noticed the derivation accounts were different. I thought this was on purpose.

Issue #121, opened 2021-05-02
Encoded target call has extra bytes in front


tomusdrw:
For some reason encoding a call using `targetApi` ends up with two extra bytes in front. I think it might be due to the call encoding entire transaction not necessarily just `Call` enum.

We should figure out elegant way how to properly encode call:
https://github.com/paritytech/parity-bridges-ui/blob/8996e64ef142fe770176222f0f388c8309e5ed26/src/hooks/useTransactionType.ts#L70

hbulgarini:
@tomusdrw i had some conversations about this with other polkadotjs api users and it seems this is not a bug since the two extra bytes comes from the SCALE encoder. Do you think that we should reviewing this or consider it as something to fix?

tomusdrw:
IMHO we should understand what these 2 bytes mean and if that number is constant (i.e. maybe for some calls there are 3 bytes or just 1?). Not understanding this might lead to subtle bugs.
Perhaps this is some sort of variant prefix or length encoded?

After we understand what that is we can use a better function name to get rid of these bytes (e.g. `trimLengthPrefix` or `decodeVariant`, etc) to future proof the code instead of using magical constants.

Issue #114, opened 2021-05-01
Fix console errors


tomusdrw:
Example:
![image](https://user-images.githubusercontent.com/581548/116781740-34509280-aa85-11eb-8433-c0f5d252bc4b.png)

tomusdrw:
Another one:
![image](https://user-images.githubusercontent.com/581548/116783304-4a168580-aa8e-11eb-8551-d5005ea0c1d2.png)

PR #123, opened 2021-05-02
Make transfer work

tomusdrw:
This PR (together with #120 & #118) finally make it possible to send a transfer using UI.

The main issue is that currently target-chain call encoding is incorrect (see #121) - this PR creates a workaround for this and adds a bunch of useful logging to compare with CLI values.

### DIFF
--- a/src/components/Network.tsx
+++ b/src/components/Network.tsx
@@ -79,7 +79,7 @@ export const NetworkSides = () => {
   const { sourceChainDetails, targetChainDetails } = useSourceTarget();
   return (
     <Box marginY={2} className={classes.main}>
-      <Box>{sourceChainDetails.sourceChain}</Box>
+      <Box p>{sourceChainDetails.sourceChain}</Box>
       <Divider />
       <IconButton size="small">
         <KeyboardReturnIcon className="onlyHover" fontSize="small" />
--- a/src/hooks/useTransactionPreparation.ts
+++ b/src/hooks/useTransactionPreparation.ts
@@ -95,28 +95,25 @@ export default function useTransactionPreparation({
   ]);
 
   useEffect(() => {
-    const getPayload = async () => {
-      if (account && call && weight) {
-        const payload = {
-          call: compactAddLength(call),
-          origin: {
-            SourceAccount: account.addressRaw
-          },
-          // TODO [#122] This must not be hardcoded.
-          spec_version: 1,
-          weight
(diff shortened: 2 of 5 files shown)


wirednkod (approved):

src/components/Network.tsx
```diff
@@ -79,7 +79,7 @@ export const NetworkSides = () => {
   const { sourceChainDetails, targetChainDetails } = useSourceTarget();
   return (
     <Box marginY={2} className={classes.main}>
-      <Box>{sourceChainDetails.sourceChain}</Box>
+      <Box p>{sourceChainDetails.sourceChain}</Box>
```
wirednkod, line 82: This is not correct. Probably you meant: <Box component="p"> ?

wirednkod (changes requested):

src/components/Network.tsx
tomusdrw: So I honestly how this is supposed to work, but it's not about the component type, but rather about using `padding` from `theme`.
There are some details here: https://material-ui.com/system/spacing/#notation

I initially removed that cause the prop-validation complaints it's not a valid prop, but well, it looks like shit when removed, hence I guess we need it.
The error can be dealt with via #114

src/hooks/useTransactionPreparation.ts
```diff
@@ -68,13 +69,10 @@ export default function useTransactionPreparation({
...
         lane_id: laneId,
-        payload: u8aToHex(payloadType.toU8a())
+        payload: payloadType.toHex()
       });
 
-      const estimatedFeeCall = await sourceApi.rpc.state.call<Codec>(
-        estimatedFeeMethodName,
-        u8aToHex(messageFeeType.toU8a())
-      );
+      const estimatedFeeCall = await sourceApi.rpc.state.call<Codec>(estimatedFeeMethodName, messageFeeType.toHex());
```
hbulgarini, line 75: what's the benefit fo this reformatting? personally , i prefer the two lines

hbulgarini (approved):
One comment that it's not a blocker though

wirednkod (approved):

src/hooks/useTransactionPreparation.ts
tomusdrw: 🤷 don't ask me, I just obey prettier :)

PR #127, opened 2021-05-03
Make TransactionDisplay to appear before submission

### LINKED ISSUE #108: TransactionStatus card to appear before hitting the submit button, lock the form after hitting submit
1. Once all fields are filled correctly, the TransactionStatus card should appear above the Submit button
2. On submit, all fields and Submit button have to be disabled. 
3. The disabled Submit button (or component placed instead of the Submit button) should show that transaction is in progress indicating that user has to wait. Alternatives CTAs are: "See in my local storage" opens the history drawer, "New transfer" cleans up the form or its parts for the new transaction

Effectively you see TransactionStatus card in the main UI only in two cases: when you have filled the form correctly and did not hit submit yet, and in the disabled (after submit) main UI form that "holds" values of the most recent transaction. Potentially the card in the disabled UI could be the only place where we live-update at this prototype stage

Screenshot 2021-05-11 at 17 51 53

wirednkod:
Fixes #108 
Screenshot 2021-05-03 at 1 49 09 PM


src/components/TransactionStatus.tsx
```diff
@@ -101,29 +120,40 @@ function TransactionStatus({ transaction, onComplete }: Props) {
...
 
-    const completionStatus = (status: boolean) =>
-      transaction.block && status ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
+    const completionStatus = (status: boolean = false): TransactionStatusEnum => {
+      if (transaction.id === 0) {
+        return TransactionStatusEnum.NOT_STARTED;
+      } else if (!status) {
+        return transaction.block ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
+      } else {
+        return transaction.block && status ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
```
tomusdrw, line 137:
```suggestion
      }
      
      if (!status) {
        return transaction.block ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
      }
      
      return transaction.block && status ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
```

I find the early-exit pattern much easier to follow when the rest is not nested in `else` branch, cause the conditions are in visually-predictable locations.

src/hooks/useSendMessage.ts
```diff
@@ -148,7 +148,6 @@ function useSendMessage({ isRunning, isValidCall, setIsRunning, input, type, wei
         return isRunning || !account;
         break;
       case TransactionTypes.TRANSFER:
-        console.log('asdf', receiverAddress, account);
```
tomusdrw, line 151: 😰

src/screens/Main.tsx
```diff
@@ -94,11 +94,13 @@ function Main() {
               {searchItems(index)?.component}
             </Grid>
           </Grid>
-          <Grid container>
-            <Grid item md={12}>
-              <Transactions />
+          {index !== 3 && (
```
tomusdrw, line 97: `index !== 3` is ugly, can you use a constant or at least extract to a variable, like so:
```
const isConnectToWalletScreen = index === CONNECT_WALLET_SCREEN;
```

tomusdrw (commented):
looks good, couple of grumbles

src/components/TransactionStatus.tsx
```diff
@@ -101,29 +120,40 @@ function TransactionStatus({ transaction, onComplete }: Props) {
...
-    const completionStatus = (status: boolean) =>
-      transaction.block && status ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
+    const completionStatus = (status: boolean = false): TransactionStatusEnum => {
+      if (transaction.id === 0) {
+        return TransactionStatusEnum.NOT_STARTED;
+      }
+      if (!status) {
+        return transaction.block ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
+      }
+      return transaction.block && status ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
```
tomusdrw, line 138:
```suggestion
      if (status) {
        return transaction.block ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
      }
      return TransactionStatusEnum.IN_PROGRESS;
```

Now I've noticed the conditions are a bit weird, I think that's the correct behavior.

wirednkod: What you propose is not correct. There is only 1 case that there is no status passed to the function (1st step):
```
{
        chainType: sourceChain,
        label: 'Include message in block',
        onChain: transaction.block,
        status: completionStatus()
      }
   ```
   On this case we may have the same if/else but we do not check for status(if provided) compared to all rest of steps - and still we want the TransactionStatusEnum.COMPLETED to return

Comment (tomusdrw) [reply src/components/TransactionStatus.tsx:138]: But what you have now has no sense at all - `status` is known to be `true`, so you can remove `&& status` and then you end up with exactly the same thing in both `if` and outside conditions so it means `status` does not matter at all.

Comment (wirednkod) [reply src/components/TransactionStatus.tsx:138]: It matters on the first step where the call is completionStatus() [meaning status is false (init from function param)].
We need to have this step for the (non given status) step 1, but also for the rest of the steps where status SHOULD be given - but in case it is not (status = false), then the transactionStatus is in progress.
If I alter the code the way you propose, then step 1 will ALWAYS be IN_PROGRESS and will never be completed (as no parameter, ever, is passed to the completionStatus() function.

File: src/components/TransactionStatusMock.tsx [hunk 5]
```diff
@@ -0,0 +1,119 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
```
hbulgarini, line 1: i wonder if we need another component for this, can we try to add the initial values on `TransactionStatus` component? maybe i'm missing a reason for not do so.

wirednkod: The TransactionStatus components changes the transactions through the relevant hook/context. I tried to use that but there is no way that you can dispatch the transaction and replace the empty one instantly. In order to avoid to remove all the dispatch/reducer logic just in order to show an empty transaction - I preferred to make a simple Mock Component that will be destroyed as soon as transactions will start flow.

src/components/TransactionStatus.tsx
tomusdrw: Please take one more look at this code:
```
if (!status) {
        return transaction.block ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
      }
      /// status == true - cause another case was handled above
      return transaction.block && status ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
```

So it can be simplified to:
```
if (!status) {
        return transaction.block ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
      }
      return transaction.block ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
```

Which in turn can be simplified to:
```
 return transaction.block ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
```

hence `status` does not matter and that's why I assume there might be a flaw in your logic. I'm not arguing that my solution is the correct one, but I do argue that current code is incorrect.

src/components/TransactionStatusMock.tsx
hbulgarini: everything clear.... go ahead with this.

src/components/TransactionStatus.tsx
wirednkod: Code you propose:
```
const completionStatus = (status: boolean = false): TransactionStatusEnum => {
      if (transaction.id === 0) {
        return TransactionStatusEnum.NOT_STARTED;
      }
      return transaction.block && status ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
    };
```

outcome:
Screenshot 2021-05-03 at 6 20 25 PM

Code I propose:
```
const completionStatus = (status: boolean = false): TransactionStatusEnum => {
      if (transaction.id === 0) {
        return TransactionStatusEnum.NOT_STARTED;
      }
      if (!status) {
        return transaction.block ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
      }
      return transaction.block && status ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
    };
 ```
 
Outcome:
Screenshot 2021-05-03 at 6 27 04 PM

Look at the 2nd line (" Rialto: Include message in block") - that is why I say that status matters.

File: src/components/TransactionStatusMock.tsx (hunk 5 above)
Comment (wirednkod) [reply src/components/TransactionStatusMock.tsx:1]: Go ahead is not enough :) Approval is missing :P

**Review by hbulgarini — APPROVED** [review]
lgtm , anyway please proceed to get the discussions agreed with @tomusdrw  before merging.

File: src/components/TransactionStatus.tsx (hunk 4 above)
Comment (tomusdrw) [reply src/components/TransactionStatus.tsx:138]: > Code you propose:

Please try this:
```js
 return transaction.block ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
```

Obviously if you do `transaction.block && status` and `status == false` we will never get into completed state. I'm still not convinced that this is correct. If I'm right with your code all should steps should light up as completed as soon as `transaction.block` is set and I don't think that's expected behavior.

tomusdrw: Actually went through a trouble of running your code and it's definitely not correct cause it all turns into `COMPLETED` state right after the transaction is included as I suspected. Requesting changes then.

tomusdrw (changes requested):
All steps go into `COMPLETED` state right after the first one goes which is not expected.

src/components/TransactionStatus.tsx
```diff
@@ -101,29 +120,37 @@ function TransactionStatus({ transaction, onComplete }: Props) {
...
+      onComplete();
+    }
+
+    setSteps([
       {
         chainType: sourceChain,
         label: 'Include message in block',
         onChain: transaction.block,
-        status: transaction.block ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS
+        status: completionStatus(true)
```
tomusdrw, line 153:
```suggestion
        status: completionStatus(!!transaction.block)
```

src/components/TransactionStatus.tsx
```diff
@@ -101,29 +120,37 @@ function TransactionStatus({ transaction, onComplete }: Props) {
...
     };
 
-    const completionStatus = (status: boolean) =>
-      transaction.block && status ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
+    const completionStatus = (status: boolean = false): TransactionStatusEnum => {
+      if (transaction.id === 0) {
+        return TransactionStatusEnum.NOT_STARTED;
+      }
+      return transaction.block && status ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
+    };
```
tomusdrw, line 136:
```suggestion
    const completionStatus = (isCompleted: boolean): TransactionStatusEnum => {
      if (transaction.id === 0) {
        return TransactionStatusEnum.NOT_STARTED;
      }
      return isCompleted ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
    };
```

Having a default of `false` doesn't really make much sense, cause the status would never be `COMPLETED`. I think setting it to `true` is more sensible, because you are basically saying: If only there is a `block` then it's done.

However, it's even better if we don't define default and simply require the caller to always pass the status. This function has less dependencies then (good) and also is more straightforward what it does, if you were to write documentation and describe it in a sentence it would be:
"This functions returns completion status of some a transaction. If the `transaction.id` is not know it's considered not started, and later it's state is controlled by `isCompleted` parameter."

tomusdrw (approved):
Looks good, small comment about overall design.

Issue #65, opened 2021-04-16
Apply styles on Custom call component.


hbulgarini:
There is a new component similar to Transfer or Remark that permits the user to send a custom call to the message lane.
User has two input fields where the first one is the input text where the call must be provided in a hex format and the weight value. If the user provides a correct call, the decoded call will be displayed bellow:

![Screen Shot 2021-04-20 at 15 34 40](https://user-images.githubusercontent.com/12089572/115446666-ee2a4200-a1ed-11eb-878e-31b877df8625.png)

The CustomCall component is already doing all the logic, but with no design.

hbulgarini:
@goldsteinsveta  , before you start working on this one , please let me show you the feature so you can design it accordingly.

Issue #64, opened 2021-04-16
Apply styles on Transfer component.


hbulgarini:
Transfer.tsx component need to apply the following design enclosed in the red square.

![Screen Shot 2021-04-20 at 15 37 18](https://user-images.githubusercontent.com/12089572/115447038-685ac680-a1ee-11eb-9783-bb381087118d.png)

Also a little UX change is needed: Instead of showing the fee as soon as all the information is provided , the `Calculate fees` button needs to show the value when it is clicked.

goldsteinsveta:
based on discussion, I removed the calculate fees step

PR #128, opened 2021-05-03
UI: input components

### LINKED ISSUE #63: Apply styles on Remark component.
This is mainly the same design as #64  but without no receiver field.

goldsteinsveta:
This is a first iteration. Closes #63, #64, #65
Screenshot 2021-05-03 at 19 04 33

### DIFF
--- a/src/components/Remark.tsx
+++ b/src/components/Remark.tsx
@@ -17,7 +17,7 @@
 import { TextField } from '@material-ui/core';
 import React, { useState } from 'react';
 import { ButtonSubmit } from '../components';
-
+import { useSourceTarget } from '../contexts/SourceTargetContextProvider';
 import { useTransactionContext } from '../contexts/TransactionContext';
 import useLoadingApi from '../hooks/useLoadingApi';
 import useSendMessage from '../hooks/useSendMessage';
@@ -26,6 +26,8 @@ import { TransactionTypes } from '../types/transactionTypes';
 const Remark = () => {
   const [isRunning, setIsRunning] = useState(false);
   const [remarkInput, setRemarkInput] = useState('0x');
+  const { sourceChainDetails, targetChainDetails } = useSourceTarget();
+
   const areApiReady = useLoadingApi();
   const { estimatedFee } = useTransactionContext();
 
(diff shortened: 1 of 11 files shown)


goldsteinsveta:
I would pixel perfect later

src/components/Remark.tsx
```diff
@@ -46,23 +45,15 @@ const Remark = ({ className }: Props) => {
...
-        <Button variant="contained" disabled={isButtonDisabled()} onClick={sendLaneMessage}>
-          Send Remark
-        </Button>
-        <p>{estimatedFee && `Estimated source Fee: ${estimatedFee}`}</p>
-      </Container>
+      <TextField onChange={onChange} value={remarkInput} variant="outlined" fullWidth />
+      <ButtonSubmit disabled={isButtonDisabled()} onClick={sendLaneMessage}>
+        Send bridge remark from {sourceChainDetails.sourceChain} to {targetChainDetails.targetChain}
+      </ButtonSubmit>
+      <p>{estimatedFee && `Estimated source Fee: ${estimatedFee}`}</p>
```
wirednkod, line 54: add <p> after the "estiamtedFee" check. There is no reason to have an empty tag if estimatedFee does not exist

src/components/Transfer.tsx
```diff
@@ -14,29 +14,39 @@
...
+      }
+    }
   }
 }));
 
 const Transfer = () => {
   const classes = useStyles();
   const [isRunning, setIsRunning] = useState(false);
-  const [transferInput, setTransferInput] = useState('0');
+  const [transferInput, setTransferInput] = useState('');
```
wirednkod, line 48: Always add the type of the state:
```
const [transferInput, setTransferInput] = useState<string>('');
```

src/screens/Main.tsx
```diff
@@ -53,12 +54,6 @@ const MenuContents = [
     title: 'Custom Call',
     isEnabled: true,
     component: <CustomCall />
-  },
-  {
```
wirednkod, line 57: Oh no! why would you do that? This is used in a different PR and should be part of the dropdown menu

wirednkod (changes requested):

src/screens/Main.tsx
goldsteinsveta: Which PR? Do we really need it? Doesn't seem to be the case for me. Should I revert changes?

wirednkod: > Which PR? Do we really need it? Doesn't seem to be the case for me. Should I revert changes?

Hm... Nah - Just leave it and we'll fix it after merge

wirednkod (approved):

tomusdrw (approved):

PR #129, opened 2021-05-03
Fix warnings and errors that appear on console

### LINKED ISSUE #114: Fix console errors

wirednkod:
Fixes #114


hbulgarini (approved):

src/hooks/useBalance.ts
```diff
@@ -42,22 +42,28 @@ const initValues = {
...
-          setState({
-            chainTokens: data.free.registry.chainTokens[0],
-            formattedBalance: formatBalance(data.free, {
-              decimals: api.registry.chainDecimals[0],
-              forceUnit: '-',
-              withSi: providedSi
-            }),
-            free: data.free
-          });
+          mountedRef.current &&
```
tomusdrw, line 57: Ugh, this is super hacky. Can we try to hide it in a re-usable hook instead? I suspect that this might not be the only place where we do that and even if not it will come handy for the future.

```js
function useAsyncEffect (func, deps) {
  const mountedRef = useRef(true);
  useEffect(() => {
    return () => {
      mountedRef.current = false;
    };
  }, []);
  useEffect( () => { 
      const applyEffects = (callback) => {
         if (mountedRef.current) { callback() }
      }
      func(applyEfffects)
  }, deps);
}
```

and it would be used like that:
```js
useAsyncEffect((applyEffects) => {
   const u = await api.query.system.account(address, ({data}) => {
      applyEffects(() => {
         setState(...);
      });
   });
}, [api.query.system, setState]);
```

tomusdrw (commented):
looks good, but I'd enclose the boilerplate into a hook.

src/hooks/useBalance.ts
wirednkod: This is not hacky - it is actually the correct way to go concerning asyncs in lifecycle - This way you make sure that there is no unmounting of the component upon resolve of the promise.
I agree the reusable (yet another) hook.

wirednkod: This is not exactly a valid solution for many reasons - but I understood your point and you are right.
Instead of 1 - I created 2 reusable hooks - that will be needed in the very near future.
In addition instead of useEffect, I used the useCallback hook in order to avoid unnecessary calls.

wirednkod: @tomusdrw I tried the solution above and that does not work (obv) as you cannot call a hook inside of another hook (useEffect). Based on the example above the `useAsyncEffect((applyEffects) => {` would be called inside `useEffect((): (() => void) => {` which cannot be done.

As an alternative, and for reusability reasons I fixed 2 new hooks. 1 called useIsMounted and 1 called useAsyncEffect:
`useIsMounted`:
```

import { useEffect, useRef, useCallback } from 'react';

export const useIsMounted = () => {
  const mountedRef = useRef(false);
  const isMounted = useCallback(() => mountedRef.current, []);

  useEffect(() => {
    mountedRef.current = true;

    return () => {
      mountedRef.current = false;
    };
  }, []);

  return isMounted;
};
```
`useAsyncEffect`:
```
import { useCallback } from 'react';
import { useIsMounted } from './useIsMounted';

export const useAsyncEffect = () => {
  const isMounted = useIsMounted();
  const asyncEffect = useCallback(
    (func: () => void) => {
      isMounted() && func;
    },
    [isMounted]
  );

  return asyncEffect;
};
```

and called the latter in a acceptable from react manner as can be seen below:
```
const u = await api.query.system.account(address, ({ data }): void => {
    asyncEffect(
       setState({
          chainTokens: data.free.registry.chainTokens[0],
           formattedBalance: formatBalance(data.free, {
           decimals: api.registry.chainDecimals[0],
           forceUnit: '-',
           withSi: providedSi
        }),
        free: data.free
      })
   );
});
return Promise.resolve(u);
```
Unfortunately that is not solving the issue. 
The only way so far that I managed to solve the "mount" issue is to use the useIsMounted hook straight in the code which actually makes more sense to me (already used in Burnr wallet and works like a charm):

```
const isMounted = useIsMounted();
.....
isMounted() && setState({
```
This will be reusable and without the need of creating extra useEffects for checking the mounted.current reference.
What do you think?

tomusdrw (approved):

PR #130, opened 2021-05-03
Fix call encoding (again)

tomusdrw:
This is a follow up on #123.

Initially when I was testing the UI I used a more verbose form (like below). Before making a PR I decided to avoid duplicating of method calls and cache the result in `apiPromise`.

Looks like for some reason this is actually broken, and returns a completely different call, so I suspect the same object cannot be used for both `transactionPayment` chain and regular transfer call.

### DIFF
--- a/src/hooks/useTransactionType.ts
+++ b/src/hooks/useTransactionType.ts
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
 
-import { hexToU8a, isHex } from '@polkadot/util';
+import { hexToU8a, isHex, u8aToHex } from '@polkadot/util';
 import { useEffect, useState } from 'react';
 
 import { useAccountContext } from '../contexts/AccountContextProvider';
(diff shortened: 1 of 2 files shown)


src/hooks/useTransactionType.ts
```diff
@@ -55,26 +55,24 @@ export default function useTransactionType({ input, type, weightInput }: Props):
 
   useEffect(() => {
     async function getValues() {
-      let apiPromise, paymentInfo, apiCall;
+      let paymentInfo, apiCall;
```
wirednkod, line 58: use types on each variable and init them. I think we should always do that as soon as we use TS

wirednkod (changes requested):

src/hooks/useTransactionType.ts
tomusdrw: These types are not easy to express and my understanding was that type script is pretty good in type inference. It's really just used as temporary holders - I'd define them in `switch` branches, but since `switch` is such a poor JS construct they would leak, so the linter prevents me from doing so.

I worked around it, please let me know wdyt.

wirednkod (approved):

PR #132, opened 2021-05-04
AsyncState hook for update state only when component is mounted

wirednkod:
I have been thinking the PR #114 and the discussion I had with @tomusdrw about the isMounted and async calls as can be found here: https://github.com/paritytech/parity-bridges-ui/pull/129#pullrequestreview-650681805 and I realised that instead of having a "rule of thumb" that `inside an async/await to always make sure before updating the state to put first the isMounted hook` - it would be easier if we had a custom useState hook that would do that for us.

This PR is the outcome of the thought above.

Instead of:
```
import { useState } from 'react';
const [state, setState] = useState<State>(initValues);
....
isMounted() && setState({...
```
it makes it easier and faster to:
```
import { useMountedState } from './useMountedState';
const [state, setState] = useMountedState<State>(initValues);
....
setState({....
````

### DIFF
--- a/src/hooks/useBalance.ts
+++ b/src/hooks/useBalance.ts
@@ -19,12 +19,12 @@ import { VoidFn } from '@polkadot/api/types';
 import { Balance } from '@polkadot/types/interfaces';
 import { formatBalance } from '@polkadot/util';
 import BN from 'bn.js';
-import { useState, useEffect } from 'react';
+import { useEffect } from 'react';
 
 import { MessageActionsCreators } from '../actions/messageActions';
 import { useUpdateMessageContext } from '../contexts/MessageContext';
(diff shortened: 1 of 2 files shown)


tomusdrw (approved):
Awesome stuff, that's exactly the solution I was trying to challenge you to implement with my poor attempts of writing `useAsyncEffect`.

hbulgarini (commented):
beautiful

hbulgarini (approved):

wirednkod:
> Awesome stuff, that's exactly the solution I was trying to challenge you to implement with my poor attempts of writing `useAsyncEffect`.

I did not like the isMounted either. I wanted something more straightforward and you confused me with the "useEffect" direction.. its your fault 😂😂

PR #135, opened 2021-05-04
Minor warnings fixes

wirednkod:


hbulgarini (changes requested):
Please take into account that with this change, the placeholder identicon disappears when no account is selected  in the sender dropdown. You need to tweak this logic to do so:

https://github.com/paritytech/parity-bridges-ui/blob/master/src/components/AccountDisplay.tsx#L82

hbulgarini:
I realized that 

> Please take into account that with this change, the placeholder identicon disappears when no account is selected in the sender dropdown. You need to tweak this logic to do so:
> 
> https://github.com/paritytech/parity-bridges-ui/blob/master/src/components/AccountDisplay.tsx#L82

Well this bug was introduced before this change but it worth to solve this issue in this PR if you agree.

wirednkod:
I am (really hard) trying to understand how the change I made has ANYTHING to do with the placeholder disappearing. Can you please explain @hbulgarini ?

hbulgarini:
> I am (really hard) trying to understand how the change I made has ANYTHING to do with the placeholder disappearing. Can you please explain @hbulgarini ?

the value 'init' was initially set so the check of the address existence does not hide the identicon. When we remove it , the identicon disappear as part of that check.

https://github.com/paritytech/parity-bridges-ui/blob/master/src/components/AccountDisplay.tsx#L82

src/components/AccountDisplay.tsx
```diff
@@ -79,9 +81,9 @@ const AccountDisplay = ({ address = '', addressKind, balance, friendlyName, hide
 
   return (
     <Container onClick={onClick} className={classes.container}>
-      <div className={classes.icon}>{address && <AccountIdenticon address={address} />}</div>
-      <div className={classes.address}>
-        <p>{displayText()}</p>
+      <div className={classes.icon}>{<AccountIdenticon address={address} />}</div>
```
tomusdrw, line 84: I guess now it's not conditional we can actually remove the `'init'` stuff, no?

tomusdrw (approved):
lgtm!

src/components/AccountDisplay.tsx
wirednkod: Yeap.. i did not? FFS!!!!

hbulgarini (approved):

Issue #86, opened 2021-04-23
Read chain metadata from RPC


tomusdrw:
Things like:
- Network name (`system.chain()` RPC)
- SS58 prefix (`system.ss58prefix` constant)
- Bridge Ids (`system.properties()` RPC)

should be read from the RPC node directly instead of being hardcoded in `.env` variables.

Issue #62, opened 2021-04-16
Build list of existing components that need to be tested.


hbulgarini:
To build the list of components to start creating tests.
Define snapshot strategy.

PR #136, opened 2021-05-04
Read chain metadata configs.

hbulgarini:
Closes: #86 

We should pay special attention to this PR since replacing the hardcoded values in the `.env` forced me to move the onchain configs to the sourceTarget reducer and read them async. Also i took the time to refactor some code in the the hook that builds the connection that makes it cleaner. 

A pending task is to add the corresponding tests to increase the reliability of this core process of the UI (related #62 )
```
src/configs/connectionChainInformation.ts
src/configs/substrateConfigs/index.ts
src/hooks/useConnections.ts
```


.env
```diff
@@ -1,10 +1,5 @@
...
+REACT_APP_PAIR=RialtoMillau
 REACT_APP_SUBSTRATE_PROVIDER_CHAIN_1=wss://wss.rialto.brucke.link
 REACT_APP_SUBSTRATE_PROVIDER_CHAIN_2=wss://wss.millau.brucke.link
 REACT_APP_LANE_ID=0x00000000
-REACT_APP_KEYRING_DEV_LOAD_ACCOUNTS=false
-REACT_APP_SS58_PREFIX_CHAIN_1=48
-REACT_APP_SS58_PREFIX_CHAIN_2=60
-REACT_APP_BRIDGE_ID_CHAIN_1=rlto
-REACT_APP_BRIDGE_ID_CHAIN_2=mlau
+REACT_APP_KEYRING_DEV_LOAD_ACCOUNTS=false
```
wirednkod, line 5: add a new line

README.md
```diff
@@ -10,16 +10,11 @@ The goal of the UI is to provide the users a convenient way of interacting with
 In the `.env` file at root project directory, you have to define the following env variables:
```
wirednkod, line 10: That is not correct. As soon as .env is being committed then this should refer ONLY to `REACT_APP_RIALTO_SUBSTRATE_PROVIDER` and `REACT_APP_RIALTO_SUBSTRATE_PROVIDER` and only for .env.local (or else people will be pushing the .env to PRs)

src/App.tsx
```diff
@@ -18,28 +18,21 @@ import { Backdrop, CircularProgress, createMuiTheme, ThemeProvider } from '@mate
...
 import { KeyringContextProvider } from './contexts/KeyringContextProvider';
 import { MessageContextProvider } from './contexts/MessageContext';
 import { SourceTargetContextProvider } from './contexts/SourceTargetContextProvider';
 import { TransactionContextProvider } from './contexts/TransactionContext';
-import { useApiConnection } from './hooks/useApiConnection';
+import { useConnections } from './hooks/useConnections';
 import Main from './screens/Main';
 
-const providers = getChainProviders();
+const connectionDetailsOne = getConnectionChainInformation('1');
```
wirednkod, line 31: I dont understand how this hardcoded things (including the getConnectionChainInformation) is of any use... you are calling a function with hardcoded value, which then calls another function (getSubstratePairConfigs) from another file to get some more hardcoded values based on a hardcoded string ('RialtoMillau') and construct a JSON. It would be easier if we had a simple Map in a file and feed the values straight from there...

src/configs/connectionChainInformation.ts
```diff
@@ -0,0 +1,53 @@
...
+import { WsProvider } from '@polkadot/api';
+
+import { checkEnvVariable } from '../util/envVariablesValidations';
+import { createPolkadotJsUrl } from '../util/createPolkadotJsUrl';
+import { getSubstratePairConfigs } from './substrateConfigs';
+import { ConnectionChainInformation } from '../types/sourceTargetTypes';
+
+const getProvider = (provider: string) => new WsProvider(provider);
+
+const getConnectionChainInformation = (chainNumber: string): ConnectionChainInformation => {
```
wirednkod, line 26: as mentioned in previous comment - I'm trying to understand the use of this file

wirednkod (changes requested):

README.md
hbulgarini: i'm not getting exactly what you mean, but at least you made me realize that variables are incorrectly called.

They should be:
REACT_APP_SUBSTRATE_PROVIDER_CHAIN_1=wss://wss.rialto.brucke.link
REACT_APP_SUBSTRATE_PROVIDER_CHAIN_2=wss://wss.millau.brucke.link

src/App.tsx
hbulgarini: I still depend on the .env values for constructing the api connections basically. 
The purpose of  the hardcoded values ("1" & "2") is to process those connection details per chain in  an agnostic way, meaning that instead of calling  variables in the code `RialtoHasher` it has to be `Chain1Hasher` so instead we need to implement a new pair like Rococo<>Westend , Chain1 be initialized with Rococo and Chain2 with Westend.

Then the use of the .env variable  REACT_APP_PAIR is for determining setting at code level specific custom values for each pair/chain ( you can see it with millau custom hasher )

src/configs/connectionChainInformation.ts
hbulgarini: explained above.

src/App.tsx
hbulgarini: Anyway is better to extract the "1" and "2" to a const

hbulgarini (commented):
Comments replied.

README.md
hbulgarini: Anyway, please take into account that the other purpose of the .env file is for deployment purposes. So in this case updating the .env is right because a new variable was added and some of them were removed. Also we will split specific `.env` files for deploying specific bridge pair.

src/configs/connectionChainInformation.ts
hbulgarini: The use of this file is to generate dynamically and agnostically  all the necessary values for constructing the API connection so then we can read the on-chain metadata config values. I say agnostically because at this point there is not `Rialto` or `Millau` , there are `chain1` and `chain2` that are filled with the values that comes from the `getSubstratePairConfigs ` that this yes generates specific set up for Rialto and Millau. 

I know that for only one pair all this complexity does not have too much sense , but if we add a new pair in the future, let's say MIllau<>Westend, then is just a matter of adding a new case in the file `src/configs/substrateConfigs/index.ts`. As i mentioned before, in this file we should provide specific chain related information.

Anyway, i think is better to move the `provider` and `polkadotjsUrl`to the file `src/configs/substrateConfigs/index.ts` so this file does not handle any construction logic.

hbulgarini: After reviewing the code , i think this file was redundant since i was doing the similar logic in two files. So i removed this one and i unified everything on file: `src/configs/index.ts`

wirednkod (approved):

tomusdrw (commented):
Definitely a step into right direction, I left a bunch of comments, but we can address them in follow up PRs - I'll create separate issues so that it can be addressed in parallel.

Most notably I'd like to avoid hardcoding "Bridge Pair". We should only rely on the two endpoints and detect network names from there and only then load custom types that matches these network names (in future we should simply load `Rococo`'s custom types as default).

hbulgarini (commented):
Replied comments.

.env.example
```diff
@@ -1,10 +1,5 @@
-REACT_APP_CHAIN_1=<Chain Name #1>
-REACT_APP_CHAIN_2=<Chain Name #2>
+REACT_APP_PAIR=<Pair identifier>
```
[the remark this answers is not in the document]
tomusdrw: > One possible solution that we might try is to create the api connection without any custom type or hasher and try to obtain the chain name ... process al the dynamic values, disconnect the API and connect it again with all the dynamic values filled.

Yup, that sounds like a good solution.

src/components/GenericAccount.tsx
```diff
@@ -56,27 +55,24 @@ const DERIVED = 'DERIVED';
...
   } = useSourceTarget();
 
-  const { dispatchTransaction } = useUpdateTransactionContext();
-  const chainsConfigs = getChainConfigs();
-  const { bridgeId } = chainsConfigs[sourceChain];
-  const { SS58Format: targetSS58Format } = chainsConfigs[targetChain];
-
-  const nativeAddress = encodeAddress(value, targetSS58Format);
+  const nativeAddress = encodeAddress(value, ss58Format);
   const nativeState = useBalance(sourceApi, nativeAddress, true);
```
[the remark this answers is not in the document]
hbulgarini: addressed.

.env.example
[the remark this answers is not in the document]
hbulgarini:
```
One possible solution that we might try is to create the api connection without any custom type or hasher and try to obtain the chain name ... process al the dynamic values, disconnect the API and connect it again with all the dynamic values filled.
```
I have created a PR to this branch where i have included a proposal for this approach #146 

I thought that creating a PR to this branch was going to be even cleaner for disscussing.

src/configs/index.ts
```diff
@@ -0,0 +1,79 @@
...
+  types: ApiOptions['types'];
+  polkadotjsUrl: string;
+  provider: ProviderInterface;
+}
+
+function hasherH512(data: any) {
+  return u8aConcat(blake2AsU8a(data), keccakAsU8a(data));
+}
+
+const getProviderInfo = (chainNumber: string, types: ApiOptions['types']) => {
```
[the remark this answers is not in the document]
tomusdrw: I agree, but my concern was with the boilerplate code that this fact generates. It's an internal function, simply use a number or `'1'|'2'` instead of introducing extra constants.

src/configs/substrateCustomTypes/customTypesMillau.json
```diff
@@ -0,0 +1,172 @@
+{
```
tomusdrw, line 1: I guess these files were not meant to be committed.

src/constants.ts
```diff
@@ -17,6 +17,9 @@
 export const SOURCE = 'sourceChain';
 export const TARGET = 'targetChain';
 
+export const CHAIN_1 = '1';
+export const CHAIN_2 = '2';
+
```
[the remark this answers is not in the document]
tomusdrw: But it's all private usage. The only place where it is exported outside is via `chainNumber` which, honestly, given it's name I'd expect to be a `number`.

I don't get what problem it really solves - as mentioned in the other comment I think it would be better to define the expected type to be `'1'|'2'` instead of generic `string` & then define some constants to overcome an issue of "I don't know what `string` I need to pass to this function."

While I am a fan of constants for things like arbitrary strings or arbitrary numbers, the main purpose of that is to give them meaning, examples
```
const SEND_MESSAGE_RUNTIME_API = 'palletBridgeGrandpa::send_message';
const NUMBER_OF_SUPPORTED_CHAINS = 3;
```
In this example, we basically do:
```
export const ONE_CHARACTER_STRING_WITH_NUMBER_ONE = '1'
export const ONE_CHARACTER_STRING_WITH_NUMBER_TWO = '2'
```
for purpose of using it when calling private functions. That really sounds like too much. If you really think they do provide value, please at least move them to the `src/configs/index.ts` file and make them local. IMHO they should never be used again (if they are that indicates some more severe code issue) and they just pollute the "global" file.

src/hooks/useConnections.ts
```diff
@@ -0,0 +1,77 @@
...
+      const chainName2 = chain2Configs.chainName;
+      const apiReady = apiConnection1.isApiReady && apiConnection2.isApiReady;
+
+      const getBridgedSS58Format = (
+        chainConfig: Configs,
+        bridgedChainConfig: Configs,
+        bridgedChain: string
+      ): Configs => {
+        const bridgeId: number[] = bridgedChainConfig!.bridgeIds![bridgedChain];
+        chainConfig.bridgeId = bridgeId;
```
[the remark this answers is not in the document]
tomusdrw: > this is used for updating one time only value

That's how it starts :) I don't really understand why we need it anyway? We can simply do:
```
targetChain.bridgeIds[sourceChain.name]
```
every time we want that value. Now you need to provide some dummy intiial values to overcome this issue (so you add more sub-optimal code in a consequence of adding this code in the first place).
Code like that makes it much harder to understand where things are coming from which imho is a recipe for disaster (please read on "code smells").

src/types/sourceTargetTypes.ts
```diff
@@ -15,14 +15,36 @@
...
+export interface ConnectionChainInformation {
+  chainNumber: string;
+  hasher?: (data: Uint8Array) => Uint8Array;
+  provider: ProviderInterface;
+  types?: ApiOptions['types'];
+  polkadotjsUrl: string;
+}
+
+export interface Configs {
+  bridgeId: Array<number>;
```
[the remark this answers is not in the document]
tomusdrw: My point was that this is redudant information, cause that's something we should read from `targetChain.bridgeIds`. Let's continue here: https://github.com/paritytech/parity-bridges-ui/pull/136/files#r626473191

src/types/sourceTargetTypes.ts
```diff
@@ -32,22 +54,34 @@ export interface ApiPromiseContextType {
...
-    sourceApiConnection: ApiPromiseConnectionType;
-    sourceChain: string;
-    sourcePolkadotjsUrl: string;
-  };
-  [ChainDetails.TARGET]: {
-    targetApiConnection: ApiPromiseConnectionType;
-    targetChain: string;
-    targetPolkadotjsUrl: string;
-  };
+  [ChainDetails.SOURCE]: SourceState;
```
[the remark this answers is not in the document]
tomusdrw: 1. We already have `Connection` which is identical.
2. We need to duplicate the type for `source` and `target`.

So essentially we now have 3 types to encode exactly the same information. What's wrong with using `Connection` here? And what's the deal with `ChainDetails` enum (i.e. what's wrong with using literal fields?)? Why this can't be simply:
```
export interface SourceTargetState {
  source: Connection,
  target: Connection,
}
```

src/configs/index.ts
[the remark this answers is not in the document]
tomusdrw: Created an issue to discuss further: https://github.com/paritytech/parity-bridges-ui/issues/148

src/types/sourceTargetTypes.ts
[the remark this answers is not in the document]
tomusdrw: Tracking issue: https://github.com/paritytech/parity-bridges-ui/issues/149

Issue #111, opened 2021-05-01
Large balance in `amount` crashes the app


tomusdrw:
![image](https://user-images.githubusercontent.com/581548/116780988-40862100-aa80-11eb-8a9b-681d69c9325e.png)

hbulgarini:
@wirednkod , as you will revisit the whole process in this issue please take into account that when you set an incorrect character and you delete it , the validation error message does not get reseted.

steps to reproduce the issue:
1) on the amount field enter not allowed character like `C` or `#`.
2) delete and leave the input empty.

Error message remains even when the input is empty.

PR #138, opened 2021-05-05
Fix valid float on amount input

wirednkod:


hbulgarini (approved):

src/components/Transfer.tsx
```diff
@@ -60,7 +60,7 @@ const Transfer = () => {
   });
 
   const onChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    setTransferInput(event.target.value);
+    event.target.value.length <= 12 && setTransferInput(event.target.value);
```
tomusdrw, line 63: This prevents sending anything greater than 1000 tokens.

Also it is imho pretty bad experience when the input field is limited (like it enforces you to do changes in particular order - you need to remove some numbers first before adding them - imagine you are at limit and want to change 0->9. You can't type `9` and press del/backspace, but need to do it in the other way).

We should display an error if the input is not correct and let the user correct it..

tomusdrw (changes requested):

src/components/Transfer.tsx
tomusdrw: Note that we do support typing `1k` now (which translates to 1000 tokens) and this PR disables that. While the intention to fix this is noble, we should do it properly.

wirednkod: > Note that we do support typing `1k` now (which translates to 1000 tokens) and this PR disables that. While the intention to fix this is noble, we should give it a little bit more love.

Oh we need to discuss this :D

src/util/evalUnits.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @returns {[number | null, string]} an array of 2 items
+ * the first is the actual calculated number (or null if none) while
+ * the second is the message that should appear in case of error
+ */
+export function evalUnits(input: string): [number | null, string] {
+  if (!floats.test(input) && !ints.test(input) && !alphaInts.test(input) && !alphaFloats.test(input)) {
+    return [null, EvalMessages.GIBBERISH];
+  }
+
+  if (floats.test(input) || ints.test(input)) {
```
tomusdrw, line 61: I'm not sure if regex are the right solution for floats and ints. Couldn't we do;
```
const isFloat = parseFloat(value).toString() == value;
const isInt = parseInt(value).toString() == value;
```
?

src/util/evalUnits.ts
```diff
@@ -0,0 +1,71 @@
...
+    return [parseFloat(input), EvalMessages.SUCCESS];
+  } else if (alphaInts.test(input) || alphaFloats.test(input)) {
+    const numericPart = parseFloat(input.replace(/[,]/g, '.'));
+    const charPart = input.replace(/[0-9.,]/g, '');
```
tomusdrw, line 65: This means you can also write something like `k1.3` or `1.k3` and it will be interpreted correctly. Maybe let's be a bit strict here and just allow the prefix to be at the very end?

src/components/Transfer.tsx
```diff
@@ -42,24 +43,31 @@ const useStyles = makeStyles((theme) => ({
...
   const classes = useStyles();
   const [isRunning, setIsRunning] = useState(false);
+  const [helperText, setHelperText] = useState('');
   const [transferInput, setTransferInput] = useState<string>('');
+  const [actualInput, setActualInput] = useState<number | null>();
   const { sourceChainDetails, targetChainDetails } = useSourceTarget();
 
   const areApiReady = useLoadingApi();
-
+  const transformToBase = 10 ** targetChainDetails.targetApiConnection.api.registry.chainDecimals[0];
```
tomusdrw, line 55: So it kind of is `base` already, we transform to smallest derived units. For lack of a better word, maybe we can call them `planck` units? :)

tomusdrw (commented):
Looks good, although note that we are risking overflows/rounding errors here and long term we should consider using decimal/bignumber library.
I guess it might be solved as part of #111, but if it's not we can also create a separate issue for this.

src/util/evalUnits.ts
wirednkod: Not really.. both are doing very specific parsing that does not fit to our case. E.g. parseFloat will parse correctly 10.33 but if , decimal symbol is used then it will parse it to 10. Thus can lead to severe mistakes. The Regex combination above ensures that you have specific format for int, floats, string int, string floats and parses correctly the . and , decimal symbols. This way we keep textfield as input but reduce possible mistakes that user may do

src/util/evalUnits.ts
wirednkod: The regex that are declared above and the sequence of checks plus the early exit ensures that what you mention will never happen. Try to see the tests that I added (i think they cover all possible scenarios)

src/util/evalUnits.ts
```diff
@@ -0,0 +1,71 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
```
hbulgarini, line 1: Aside comment: This should be shared with all JS developers at parity if there is not something available already.

src/util/evalUnits.ts
hbulgarini: for integers you can rely on`Number.isInteger()` instead of a regex

hbulgarini: I was diving a little bit more and found this old repo that solved this with the following code:

```
    // A numeric is a variable that contains a numeric value, regardless its type
    // It can be a String containing a numeric value, exponential notation, or a Number object
    // See here for more discussion: http://stackoverflow.com/questions/18082/validate-numbers-in-javascript-isnumeric/1830844#1830844
    isNumeric: function(n) {
      return !isNaN(parseFloat(n)) && isFinite(n);
    },

    // An integer contains an optional minus sign to begin and only the digits 0-9
    // Objects that can be parsed that way are also considered ints, e.g. "123"
    // Floats that are mathematically equal to integers are considered integers, e.g. 1.0
    // See here for more discussion: http://stackoverflow.com/questions/1019515/javascript-test-for-an-integer
    isInteger: function(i) {
      return _.isNumeric(i) && i % 1 === 0;
    },

    // A float is a numbr that is not an integer.
    isFloat: function(n) {
      return _.isNumeric(n) && !_.isInteger(n);
    },
```

All the functions called from `isNumeric` are available in the  global `Numbers` object, so we don't have to re define this .

src/util/evalUnits.ts
```diff
@@ -0,0 +1,71 @@
...
+  { value: 1e24, symbol: 'Y' }
+];
+
+const floats = /^[0-9]*[.,]{1}[0-9]*$/;
+const ints = /^[0-9]+$/;
+const alphaFloats = /^[0-9]*[.,]{1}[0-9]*[a-zA-Z]{1}$/;
+const alphaInts = /^[0-9]*[a-zA-Z]{1}$/;
+
+export enum EvalMessages {
+  GIBBERISH = 'Input is not correct. Use numbers, floats or expression (e.g. 1k, 1.3m)',
```
goldsteinsveta, line 42: I've entered suggested `1.3m` and it killed the UI

tomusdrw: Yeah, most likely #111, but I suspect that you might have ended up with a `NaN` instead of the error in #111 because we do a bunch of calculations earlier.

src/util/evalUnits.ts
tomusdrw: > This can lead to severe mistakes. 

I agree and that's why I made the comment. You check that it's float by a regex, but then you use `parseFloat` on it, which might end up with different results.

src/util/evalUnits.ts
wirednkod: This is due to #111 . I'll probs try to catch that next

src/util/evalUnits.ts
wirednkod: > but then you use parseFloat

This completely slipped. It is strange that tests did not get it (note: test was wrong so it was not capturing it - fixed it as well). Thank you - I used the replace of decimal symbol to make sure works in all cases

goldsteinsveta (approved):

src/components/ExtensionAccountCheck.tsx
```diff
@@ -28,7 +28,7 @@ const ExtensionAccountCheck = ({ component }: Props): JSX.Element => {
   if (!extensionExists) {
     msg = 'Connect to a wallet. Install polkadotjs extension';
   } else if (!accountExists) {
-    msg = 'There are no accounts in the extension. Please create one. Please create one';
+    msg = 'There are no accounts in the extension. Please create one';
```
tomusdrw, line 31: :joy:

tomusdrw (approved):

PR #140, opened 2021-05-05
Fix switch account bug transactions & refactor

hbulgarini:
This depends on the merge of  PR #136 first to be reviewed.

This PR contains:
1) Bug fixes for active transactions when chains are switched.
2) Refactor of `TransactionStatus` component. 
3) Add checks for stop asking status on useEffects once the transaction got completed.


src/components/TransactionStatus.tsx
```diff
@@ -14,21 +14,13 @@
...
-import { Codec } from '@polkadot/types/types';
-import BN from 'bn.js';
-import React, { useEffect, useState } from 'react';
-
+import React from 'react';
 import { Box, Card, makeStyles, Typography } from '@material-ui/core';
 import { ButtonSwitchMode } from './Buttons';
 import { IconTxStatus } from './Icons';
+import useTransactionSteps from '../hooks/useTransactionSteps';
+import { TransanctionStatus } from '../types/transactionTypes';
```
tomusdrw, line 22:
```suggestion
import { TransactionStatus } from '../types/transactionTypes';
```

src/hooks/useChainGetters.ts
```diff
@@ -61,7 +65,18 @@ const useSS58Format = (): State => {
...
+  const getApiByChain = (chain: string): ApiPromise => {
+    switch (chain) {
+      case targetChain:
+        return targetApi;
+      case sourceChain:
+        return sourceApi;
+      default:
+        return {} as ApiPromise;
+    }
+  };
```
tomusdrw, line 77: This is starting to become a bit overwhelming. Let me summarize why:
1. We have two functions `getApiByChain` and `getSS58ByChain`.
2. Both functions implement a `switch` statement to select something based on a `string` which is the chain name. (duplication)
3. The exact same information is retrieved from `useSourceTarget` hook and it's available under keys `sourceChainDetails` and `targetChainDetails` (which keys are part of some random enum, that is only used in some places, not all) - (duplication again).
4. This information however is under chain-specifc keys like `sourceSs58Format` or `sourceChain`, etc. (what we will learn in a bit is inconvenience and type duplication again (see #149).
4. Now in places where we use `get*` functions, we have a variable called `targetChain` which contains the `string` - so we know exactly this is target chain context and we actually happen to know it's name, cause...
5. ... it's coming from yet another hook/context (`getTransactionSourceTarget`) which surprisingly enough is using `useSourceTarget` too, but has a bit of logic to select the right chain name based on a boolean flag.

So essentially what we want to achieve is:
1. Get source or target chain details.
2. Sometimes we want to flip source/target depending on a boolean flag.

What I'd like to argue now is that the current solution is not optimal for the problem at hand, my proposal would be:
```js
// no boilerplate
const { sourceDetails, targetDetails } = useSourceTarget(); 
const sourceSs58 = sourceDetails.ss58;
```
Instead of (current):
```js
// boilerplate / voodoo
const { sourceChain } = getTransactionSourceTarget();
const { getSs58ByChain } = useChainGetters();
const sourceSs58 = getSs58ByChain(sourceChain);
```

And also:
```js
const { sourceDetails, targetDetails } = useSourceTarget(); 
const currentChain = flipChain ? sourceDetails : targetDetails;
currentChain.ss58 // instead of currentChain.sourceSs58;
```
in case we need to swap the chains. Unified types will help here.

tomusdrw: Secondly. AFAICT the whole chain flipping is needed for a feature that is not needed at all. Transaction history should be displayed only for current source chain. "Swapping chains" and observing a previously sent transaction is not something we need right now and arguably it's not a use case at all (I can elaborate why if needed, but long story short - users are interacting with one chain, it doesn't make sense for them to send a bunch of bridge transactions on source and then switch to target to send a bunch of other transactions).

tomusdrw (commented):
Please try to avoid mixing refactorings and bugfixes in one PR. Also it would help if there was an issue reported for the bugfix, so that we can discuss earlier on the feasibility and importance of the bug.
The situation with displaying transaction history while switching networks should not really be considered - the set of transactions should simply be different, we shouldn't be swapping the display.

src/hooks/useChainGetters.ts
hbulgarini: I started coding the change you proposed based on your valid points. But then i saw a benefit of the `useChainGetters` that might be duplicated or similar, but in my opinion brings simplicity at the time to use it.  
Let's see the example of the `useAccounts` hook : https://github.com/paritytech/parity-bridges-ui/pull/140/files#diff-4a157fd124ac75c9f47afe92c19eb600cd48205f57af4565eb4ceb136b988c23R53

if i'm not mistaken applying what you mention for getting the corresponding `ss58Format` for the chain  on this hook i would have to add the hook `useSourceTarget `, then call a reusable function  to determine if it is flipped or not in order to get the corresponding SS58Format ( flipped is needed because sometimes the sourceTarget can work as source and sometimes as target ).

With my approach i'm just relying on calling one hook that retrieves a function that makes everything for us under the hood.
I understand your point about duplicating and i agree as in some way that i can get exactly the same using the `useSourceTarget()`, but i like creating wrappers that then when they are used simplifies exports and logics.

hbulgarini: > Secondly. AFAICT the whole chain flipping is needed for a feature that is not needed at all. Transaction history should be displayed only for current source chain. "Swapping chains" and observing a previously sent transaction is not something we need right now and arguably it's not a use case at all (I can elaborate why if needed, but long story short - users are interacting with one chain, it doesn't make sense for them to send a bunch of bridge transactions on source and then switch to target to send a bunch of other transactions).

This comment is confusing. This bug is not for solving  the transactions listed in the history side panel, it is as well for the active transaction that is running. Based on what we agreed on the design , the way to determine the source/target direction is by choosing the account from the sender dropdown, get the corresponding chain from this account so we can set it  as source target.

If we take into account that user might want to choose an account from the target, then this would force us to switch the existing source / target direction. Hence , if there is active transactions running we need to remain with the same source and target that was used a that time when the transaction was triggered.

tomusdrw: > Let's see the example of the useAccounts hook :

I feel that this particular hook suffers from a previous decisions that were made:
1. We use `chainName` to identify chains (although in other parts of code we just call them chain 1 and chain 2). Even with multiple chains, I feel that `chainNumber` would be sufficient to identify the chain. Or we could simply pass the entire chain definition object there, not just the name.
2. The `setCurrentAccount` does two things at once: switches the chain and changes the account address.
3. Because we store `accounts` for ALL chains in one array (which I think is not the best solution) we need to look-up account that we want to switch to.
4. Instead of using some opaque account identifier (like number on that array, address which does not depend on encoding, raw address) we pass `value` (not entirely sure what that even is, but assuming it's some address) and then we re-encode the address to format that is expected **after** the chain switch to look up the address in the shared `accounts` array.

So simply changing the way we select accounts removes the need to have `getValuesByChain`. IHMO accounts should be associated with particular network and stored split by-network. This way we would simply look for an account in a limited set already.

hbulgarini: > 1. We use chainName to identify chains (although in other parts of code we just call them chain 1 and chain 2). Even with multiple chains, I feel that chainNumber would be sufficient to identify the chain. Or we could simply pass the entire chain definition object there, not just the name.


`chain1` & `chain2` are only called like this only during the initialization of the connections. Referring the chains with this generic would break the logic for building the pallet's name dynamically ( so chain names are still needed ).

`2. The setCurrentAccount does two things at once: switches the chain and changes the account address.`
i don't see problems with this. Maybe changing the name of the setter?

Regarding `3` & `4` i think it might worth to evaluate your comments but i see it too much for this PR since the target of this piece of work was the transaction status update logic. If you agree we can create a GH issue specifically for that refactor ( in case we all agree )  and leave this PR as it is ( which i think it was properly answered )

tomusdrw:
> I'm not agree with this. Refactor was part of this bugfix solution .... i got two benefits by refactoring during this bug solution:

Sure, I'm not saying you should not fix bugs by refactoring things. I only meant to avoid mixing bugfixes and refactorings in one PR. From your list of things in the description seems that these things are unrelated, so they could have been separate PRs.

hbulgarini:
> > I'm not agree with this. Refactor was part of this bugfix solution .... i got two benefits by refactoring during this bug solution:
> 
> Sure, I'm not saying you should not fix bugs by refactoring things. I only meant to avoid mixing bugfixes and refactorings in one PR. From your list of things in the description seems that these things are unrelated, so they could have been separate PRs.

Agree , the description gives the impression that i mixed two activities. I wanted to emphasize that i improved the readability of the `TransactionStatus` component in a nutshell.

src/hooks/useChainGetters.ts
tomusdrw: >  If you agree we can create a GH issue specifically for that refactor ( in case we all agree ) and leave this PR as it is ( which i think it was properly answered )

Sure, the refactor goes to it's own PR/issue, I've only originated the discussion, because as I said, the code there started to become overwhelmingly complicated (and duplicated).

hbulgarini: > No, that's not exactly right. We happen to switch the networks, but only because we have two of them. Selecting the sender account determines the network we are going to speak to. In the future that network might be bridged with one or more other networks and not necessarily all will be the same.

Right. This was clear from the beginning but it was also said that is mid term, so today choosing a different source means swap current source/target. In my opinion, adding more than two chains is not so hard to do with the current design, but is an enhancement clearly.

> We remember which chains the transaction originates from, and we simply render the component passing api instances instead of relying on hook magic.

Is this not pretty close to the way i'm currently doing it? https://github.com/paritytech/parity-bridges-ui/pull/140/files#diff-8762bd5279e5304ba0047479b3927e524649384b2270f3023437979a5df37ed1R42 

> We only display transactions for current selected source chain. Meaning if we switch source chain, we display different set of transactions.

I think that this applies to current transactions , but for the history of transactions ( which is a valuable feature to have ) i would show all transactions. Anyway for active transactions: #158 

> IMHO (2) is the right solution for now, I'm sorry I didn't catch it earlier, but I think it's actually confusing and harming to display some old transaction below completely different set of parameters. 

What do you mean by `completely different set of parameters` ? I see the following logic remaining in the future with only showing the transactions related to the source: https://github.com/paritytech/parity-bridges-ui/pull/140/files#diff-8762bd5279e5304ba0047479b3927e524649384b2270f3023437979a5df37ed1R42

tomusdrw: > What do you mean by completely different set of parameters ?

It's about context switching and mental overhead. The users learns that in particular places of the UI some information can be correlated. For instance "Source chain is displayed in the left hand side panel, but you also see it in the transaction status when things are getting confirmed". Or: "When you wait for finality confirmation you can take a look at last finalized block on the left hand side panel to see how far you are from there".
By displaying information from opposite chain you break these mental assumptions and that leads to confusion in the best case ("it used to match last time I used the UI") and mistakes in the worst case (user makes an action based on some false understanding).
That's why I think that even if technically viable, and maybe potentially useful, this should not be displayed, just to make sure we build consistent UI, where the information displayed is not based on the context of previous actions the user did.

hbulgarini: Also: #159 159

src/hooks/useTransactionNonces.ts
```diff
@@ -0,0 +1,93 @@
...
+import { getSourceTargetRole } from '../util/chainsUtils';
+import { TransactionStatusType } from '../types/transactionTypes';
+import getSubstrateDynamicNames from '../util/getSubstrateDynamicNames';
+interface Props {
+  transaction: TransactionStatusType;
+}
+
+const useTransactionNonces = ({ transaction }: Props) => {
+  const [nonceOfTargetFinalizedBlock, setNonceOfTargetFinalizedBlock] = useState<null | number>(null);
+  const [latestReceivedNonceRuntimeApi, setLatestReceivedNonceRuntimeApi] = useState(0);
```
tomusdrw, line 36: This is set in async context so it should be using `isMounted()`.

tomusdrw (approved):
looks good, just please fix the set-state-after-unmouonted issue and it's good to go.

src/hooks/useTransactionNonces.ts
hbulgarini: https://github.com/paritytech/parity-bridges-ui/pull/140/commits/b5fb972630cfbf19197e0823f6411cf641d30488

Issue #141, opened 2021-05-06
Warn in case we keep updating state on unmounted component (leak detection)


tomusdrw:
With https://github.com/paritytech/parity-bridges-ui/pull/132 it's super easy now to avoid console warnings about setting state on unmounted component.

I'd suggest adding a counter & warning [here](https://github.com/paritytech/parity-bridges-ui/pull/132/files#diff-4be7966faa76f92418580c3deb62aa11280b3b345e54d3b219197258ad8bb020R42).
So that in case the callback is triggered and `isMounted()===false` for two (or 5 or sth) times in a row, we display an error suggesting a leak.

PR #143, opened 2021-05-06
Fix leak issue on useBalance

wirednkod:


hbulgarini:
is there any issue related to this PR?

hbulgarini (approved):

src/hooks/useBalance.ts
```diff
@@ -45,34 +45,29 @@ const useBalance = (api: ApiPromise, address: string, providedSi: boolean = fals
...
+            forceUnit: '-',
+            withSi: providedSi
+          }),
+          free: data.free
         });
-        return Promise.resolve(u);
-      } catch (e) {
+      })
+      .then((u): void => {
+        unsubscribe = u;
```
tomusdrw, line 62: You have just replaced one race condition into another. IF the component is unmounted before we hit this line we will never unsubscribe.

wirednkod: If the component is unmounted then I do not understand how this line is relevant to unsubscription...?

tomusdrw: 1. We mount the component.
2. `useEffect` fires and we start with `api.query.system.account`
3. The promise is not resolved yet (i.e. we didn't invoke `then` yet).
4. The component is unmounted and cleanup runs, but since we didn't reach `then` yet, `unsubscribe === null` so we don't unsubscribe from anything.
5. The promise resolves and we set `unsubscribe = u` - the subscription is now estabilisihed, but the component has already been unmounted, so every `setState` will fail because `isMounted() === false`. Also see #141 for something that would allow us to detect some problems.

Now we have a leak, which is really hard to find, but esentially the subscription will be there during entire lifetime of application.

wirednkod: You are right on this. I have though found the leak and roll back also the code to the previous one.
Have a look on #144

PR #144, opened 2021-05-06
Try again to fix usebalance leak

wirednkod:
Memory leak was coming from useMountedState.
There is something in that custom hook that was triggering loops (irrelevant completely to how the useEffect is written - chain   promise or async/await).
Will investigate in different PR how to make the useMountedState usable instead of using isMounted && setState - but for now I would like to fix the memory leak.

### DIFF
--- a/src/hooks/useMountedState.ts
+++ b/src/hooks/useMountedState.ts
@@ -31,10 +31,10 @@ import { useIsMounted } from './useIsMounted';
  * D is the type passed when then useMountedState is used
  * e.g. const [pass, setPass] = useMountedState<string>('someString');
  */
-export const useMountedState = <D>(initialValue: D) => {
+export const useMountedState = <D>(initialState: D | (() => D)) => {
   const isMounted = useIsMounted();
 
-  const [state, setState] = useState<D>(initialValue);
+  const [state, setState] = useState<D>(initialState);
 
   const setMountedState = (value: D) => {
     if (isMounted()) {
--- a/src/hooks/useMountedState.ts
+++ b/src/hooks/useMountedState.ts
@@ -42,5 +42,5 @@ export const useMountedState = <D>(initialState: D | (() => D)) => {
     }
   };
 
-  return [state, setMountedState];
+  return [state, setMountedState] as const;
 };
--- a/src/hooks/useMountedState.ts
+++ b/src/hooks/useMountedState.ts
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
 
-import { useState } from 'react';
+import { useCallback, useState } from 'react';
 import { useIsMounted } from './useIsMounted';
 
 /**
@@ -36,11 +36,14 @@ export const useMountedState = <D>(initialState: D | (() => D)) => {
 
   const [state, setState] = useState<D>(initialState);
 
-  const setMountedState = (value: D) => {
-    if (isMounted()) {
-      setState(value);
-    }
-  };
+  const setMountedState = useCallback(
+    (value: D) => {
+      if (isMounted()) {
+        setState(value);
+      }
+    },
+    [isMounted]
+  );
 
   return [state, setMountedState] as const;
 };
--- a/src/hooks/useBalance.ts
+++ b/src/hooks/useBalance.ts
@@ -19,7 +19,6 @@ import { Balance } from '@polkadot/types/interfaces';
 import { formatBalance } from '@polkadot/util';
 import BN from 'bn.js';
 import { useEffect } from 'react';
-import { useIsMounted } from './useIsMounted';
 
 import { useMountedState } from './useMountedState';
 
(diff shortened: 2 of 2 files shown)


hbulgarini (approved):
Approving it for solving the issue as soon as possible.
Then if we can bring back the `useMountedState` that would be nice ( not a blocker or a need for v1 or even future versions )

tomusdrw:
> and whenever the new value arrives - React could not understand that the object changed or not - so it was forcing a re-render

Can you elaborate how re-rendering is connected with the leak? Cause I don't see immediately how fixing re-rendering fixes the leak as well - I guess it's just made less severe?

hbulgarini:
> > and whenever the new value arrives - React could not understand that the object changed or not - so it was forcing a re-render
> 
> Can you elaborate how re-rendering is connected with the leak? Cause I don't see immediately how fixing re-rendering fixes the leak as well - I guess it's just made less severe?

i think this PR solves more the re-rendering loop rather than the memory leak .. actually comparing the performance of my PR #136 originated from branch originated from a master version previous to the useMountedState, the performance was not so bad ... actually it was perfectly usable .
but this PR solves the infinite loop re-rendering and in my opinion this is why i would go ahead with the merge of this PR as it is since it solves an issue with is deployed now

wirednkod:
> > and whenever the new value arrives - React could not understand that the object changed or not - so it was forcing a re-render
> 
> Can you elaborate how re-rendering is connected with the leak? Cause I don't see immediately how fixing re-rendering fixes the leak as well - I guess it's just made less severe?

Just to make things clear I will try to place a bit of what happened and what I mean by "leak".
I wanted to identify the Error that was appearing with description the "cannot update state when component is not mounted". In order to do so - I introduced the useIsMounted hook that eventually was replaced by the useMountedState hook. After that I checked the request from @tomusdrw concerning a potential "leak" (I'll explain why the "") that may or may not exist concerning the subscriptions for the useBalance. 
Once I tried to identify that I realised that the "leak" is actually a re-rendering issue that forces useBalance to run non-stop (thus not a leak but a "leak"). I thought initially that the problem was identified from the Promise.resolve/reject that were not run in time and alter the code to .then pattern. Unfortunately the case was lying with the useMountedState hook - that was forcing re-render of the components using it - forcing re-use of the useBalance and thus continuous subscriptions/un-subscriptions that was calling api non-stop and overloading memory/CPU usage. This can be seen in the screenshot below 

Screenshot 2021-05-06 at 8 17 51 PM

From the Summary at the bottom it was identified that "leak" was due to what is marked as Scripting _**(wavy pattern of JS Heap - blue line).**_. Meaning that JavaScript code was triggering expensive calls/recalculation/re-rendering that caused UI to lag. This was one of the major causes of poor performance or as I named it "leak".  After replacing the useMountedState to useState and revert code to use the `isMounted() && useState` the outcome can be seen  on following screenshot:

Screenshot 2021-05-06 at 8 39 28 PM

The Scripting actions are reduced from 4,78 secs to 1,16 secs and specific action's loading time from 5,066 secs to 1,521 secs.

So the "leak" as I mentioned before was not the suspicion of subscriptions initialising but never close - but the continuous/non-stop re-initialisation of subscriptions that were though closing.

With the solution given here - the lifecycle of React is now controlled and is triggering only when it actually needs to (state must be updated).
In order to make the useMountedState to work - a deep checking on incoming value (especially on objects/arrays) need to take place to ensure if re-set of state is needed or not. I intend to do that in a future PR.

src/hooks/useMountedState.ts
```diff
@@ -31,16 +31,16 @@ import { useIsMounted } from './useIsMounted';
...
   const isMounted = useIsMounted();
 
-  const [state, setState] = useState<D>(initialValue);
+  const [state, setState] = useState<D>(initialState);
 
   const setMountedState = (value: D) => {
     if (isMounted()) {
       setState(value);
     }
   };
```
tomusdrw, line 43: Please add `useCallback` here and I suspect this will solve your issue. Note that the leak is a separate thing - and also needs to be addressed.

tomusdrw (approved):

src/hooks/useMountedState.ts
```diff
@@ -31,16 +31,19 @@ import { useIsMounted } from './useIsMounted';
...
-      setState(value);
-    }
-  };
+  const setMountedState = useCallback(
+    (value: D) => {
+      if (isMounted()) {
+        setState(value);
+      }
+    },
+    [isMounted]
```
tomusdrw, line 45: Should we have `setState` here? I guess it doesn't change anyway.

wirednkod: Im not sure I understand. setState is the only actuon that exists inside the callback (and the reason why we created the useMountedState). if we remove that - then.... what remains?

tomusdrw: I meant to add `setState` to the array of dependencies, like so:
```
[isMounted, setState]
```
so that in case `setState` function changes the callback is created anew.

wirednkod: if added - `setState` is just a dependency - and as assigned as useState at the beginning of the component Im not sure that it would make any difference, to be added where you propose.

Issue #139, opened 2021-05-05
Create configuration files / features to customize the bridge for other chains pair.


hbulgarini:
In case a new pair needs to be added to the UI it would be nice to have a configuration file that implements all the necessary Api constructor values for both chains.

At the moment we are providing custom types and custom hashers. So at least two processes would be implemented:

1) In the config file provide URL for custom types in case they are.
2) Implement an empty JS file that will be used for exporting any custom hasher.

PR #146, opened 2021-05-10
Making connections even more dynamic and ready for new chains.

hbulgarini:
Address some of the comments of PR #136 
But also it address the first version of what it was proposed on #139


src/util/tests/envVariablesValidations.test.ts
```diff
@@ -75,8 +75,4 @@ describe('Validate .env file', () => {
   test('expected ENV variables are set', () => {
     expect(checkExpectedVariables()).toBeTruthy();
   });
-
```
hbulgarini, line 78: Needed to remove this since now some variables are optional.

src/hooks/useApiChainName.ts
```diff
@@ -0,0 +1,62 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+
+import { ApiPromise } from '@polkadot/api';
+import { useEffect, useState } from 'react';
+import logger from '../util/logger';
+import { getChainName } from '../util/getConfigs';
+import { ConnectionChainInformation } from '../types/sourceTargetTypes';
+import { BASE_CUSTOM_TYPES } from '../constants';
```
tomusdrw, line 22: I'm a bit confused, why do we still need `BASE_CUSTOM_TYPES` and `useApiChainName` hook at all? Since that with your decision to put custom types URLs to ENV variables (which is not something that I suggested in other PR, but I do like (cause it's even more configurable)) we can create API connection right away with proper types?

src/hooks/useApiChainName.ts
```diff
@@ -0,0 +1,62 @@
...
+      await apiPromise
+        .disconnect()
+        .catch((err) => logger.error(`Chain ${chainNumber}: Error disconnecting. Details: ${err.message}`));
+      logger.info(`Chain ${chainNumber} disconnected`);
+    };
+
+    if (apiPromise.isConnected && chainName) {
+      disconnect();
+    }
+  }, [apiPromise, chainName, chainNumber]);
```
tomusdrw, line 59: I'm really struggling to understand why we need this file at all, but even if we do, splitting the connect-disconnect cycle is imho a bad idea Why not simply have that as one single `useEffect`, where you basically:
1. Connect
2. Set State (watch out for mounted state,  [look for `let ignore = ` in React docs](https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies))
3. Disconnect.

src/hooks/useConnections.ts
```diff
@@ -22,14 +22,23 @@ import { SourceTargetState, ChainDetails, ConnectionChainInformation, Configs }
 interface Props {
   connectionDetailsOne: ConnectionChainInformation;
   connectionDetailsTwo: ConnectionChainInformation;
+  initConnectionDetailsOne: ConnectionChainInformation;
+  initConnectionDetailsTwo: ConnectionChainInformation;
```
tomusdrw, line 26: What's wrong with `1` and `2` that we have to actually spell them? Also I'd very much prefer if we used `arrays` everywhere, at some point we might need to connect to `three` or even `dozen` of other chains and adding more configuration variables to this interface is not the right way to go.

src/types/sourceTargetTypes.ts
```diff
@@ -26,7 +26,7 @@ export interface ApiPromiseConnectionType {
 
 export interface ConnectionChainInformation {
   chainNumber: string;
-  hasher?: (data: Uint8Array) => Uint8Array;
+  hasher: ((data: Uint8Array) => Uint8Array) | undefined;
```
tomusdrw, line 29: Is this required? It's almost exactly the same, but just less convenient (you always require the `hasher` field, but in case it's not needed it must have `undefined` value).

src/util/getConfigs.ts
```diff
@@ -20,10 +20,15 @@ export const getConfigs = async (apiPromise: ApiPromise) => {
...
-  const chainName = systemChain.split(' ')[0];
   const bridgeIds = prop.get('bridgeIds');
 
-  return { bridgeId: [], bridgeIds, chainName, ss58Format: parseInt(ss58Format.toString()) };
+  return { bridgeId: [], bridgeIds, ss58Format: parseInt(ss58Format.toString()) };
+};
+
+export const getChainName = async (apiPromise: ApiPromise) => {
+  const systemChain = await apiPromise.rpc.system.name();
+  const chainName = systemChain.split(' ')[0];
```
tomusdrw, line 31: What happens here? We should just take the chain name as-is.

src/util/tests/encodeUrl.test.ts
```diff
@@ -18,7 +18,7 @@ import { stringToU8a } from '@polkadot/util';
 import { base64Encode } from '@polkadot/util-crypto';
 import { zlibSync } from 'fflate';
 
-import customTypesRialto from '../../configs/customTypesRialto.json';
+import customTypesRialto from '../../configs/substrateCustomTypes/customTypesChain1.json';
```
tomusdrw, line 21: This might break the test in case different env variables are set. I'd suggest to actually fetching the rialto types to make sure it's deterministic.

src/util/tests/envVariablesValidations.test.ts
tomusdrw: We can instead list the optional ones?

src/util/tests/envVariablesValidations.test.ts
```diff
@@ -17,7 +17,7 @@
 import dotenv from 'dotenv';
 
 import { checkEnvVariable } from '../envVariablesValidations';
-import { checkExpectedVariables, checkUnexpectedVariables } from '../envVariablesValidations';
```
tomusdrw, line 20: If you remove the function import and it's literally the only place where that function is used, please remove the function as well (but note I don't think it should be removed, just altered to cater for optional vars.)

tomusdrw (commented):
It's a tiny bit different than what we discussed initially, but it's actually better than the original proposal (which was suggesting to hardcode the list of "known chains").

However there are couple of code improvement suggestions.

hbulgarini (commented):
Comments addressed.

src/util/envVariablesValidations.ts
```diff
@@ -15,13 +15,19 @@
...
   'REACT_APP_SUBSTRATE_PROVIDER_CHAIN_2',
   'REACT_APP_LANE_ID',
   'REACT_APP_KEYRING_DEV_LOAD_ACCOUNTS'
 ];
 
+const optionalVariables = [
+  'REACT_APP_CUSTOM_TYPES_URL_CHAIN_1',
+  'REACT_APP_CUSTOM_TYPES_URL_CHAIN_2',
+  'REACT_APP_CUSTOM_HASHER_CHAIN_1',
+  'REACT_APP_CUSTOM_HASHER_CHAIN_2'
```
tomusdrw, line 28: Cool stuff. Wdyt about chaing the variable names though to have `CHAIN_1` in front, so that you can easily sort them:
```
REACT_APP_CHAIN_1_CUSTOM_TYPES_URL_=https://raw.githubusercontent.com/paritytech/parity-bridges-common/master/deployments/types-rialto.json
REACT_APP_CHAIN_1_SUBSTRATE_PROVIDER=wss://wss.rialto.brucke.link
REACT_APP_CHAIN_2_CUSTOM_TYPES_URL=https://raw.githubusercontent.com/paritytech/parity-bridges-common/master/deployments/types-millau.json
REACT_APP_CHAIN_2_CUSTOM_HASHER_=blake2Keccak256Hasher

REACT_APP_LANE_ID=0x00000000
REACT_APP_KEYRING_DEV_LOAD_ACCOUNTS=false
```
The reason for that is that I think it's easier to think "per chain" instead of "per functionality". Like when you start defining your env variables, you will most likely go with:
1. What are the things for one chain I need to configure?
2. What are the thigns for second chain I need to configure?

Instead of:
1. What are the custom types for chain1 and chain2?
2. What are the custom hashers for chain1 and chain2?
3. What are the URLs for chain1 and chain2?

Basically - it's easier to enumerate chains than it is to enumerate all possible configuration options from a developer perspective that is configuring the UI.

hbulgarini: Yeap. Sounds good to me

.env
```diff
@@ -1,5 +1,11 @@
...
+REACT_APP_CHAIN_1_SUBSTRATE_PROVIDER=wss://wss.rialto.brucke.link
+REACT_APP_CHAIN_2_CUSTOM_HASHER=blake2Keccak256Hasher
+REACT_APP_CHAIN_2_CUSTOM_TYPES_URL=https://raw.githubusercontent.com/paritytech/parity-bridges-common/master/deployments/types-millau.json
+REACT_APP_CHAIN_2_SUBSTRATE_PROVIDER=wss://wss.millau.brucke.link
+
 REACT_APP_LANE_ID=0x00000000
 REACT_APP_KEYRING_DEV_LOAD_ACCOUNTS=false
+
+
+
```
tomusdrw, line 11:
```suggestion
```

tomusdrw (approved):
The hasher type would be good to solve, but the code looks really good already, thanks for bearing with me :)

hbulgarini:
> The hasher type would be good to solve, but the code looks really good already, thanks for bearing with me :)

Solved. Thanks for your patience and help.

PR #153, opened 2021-05-10
UI storage drawer

goldsteinsveta:
Screenshot 2021-05-11 at 17 43 17


src/components/StorageDrawer.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const handleDrawerClose = () => {
+    setDrawer('');
+  };
+
+  return (
+    <>
+      <ButtonDrawerMenu
+        startIcon={<HistoryIcon />}
+        onClick={handleDrawerOpen}
```
wirednkod, line 67: There is no point to create a function and call it for this.... just do the following:
```
onClick={() => setDrawer(`${drawer === 'open' ? '' : 'open'}`)}
```

wirednkod (changes requested):

src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,45 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
```
hbulgarini, line 1: Header seems to be repeated.

src/contexts/GUIContextProvider.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+interface DrawerContextProps {
+  drawer: string;
+  setDrawer: React.Dispatch<React.SetStateAction<string>>;
+}
+interface GUIContextProviderProps {
+  children: React.ReactElement;
+}
+
+export const DrawerContext = React.createContext({} as DrawerContextProps);
```
hbulgarini, line 30: If you add this simple function to the context

```
export function useGUIContext() {
  return useContext(DrawerContext);
}
```

Then on your children components, you just can access to the context  like this:

```
const {drawer} = useGUIContext();
```

The benefit is that imports and usage becomes easier every time you need to use this.

src/contexts/GUIContextProvider.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export function GUIContextProvider({ children }: GUIContextProviderProps): React.ReactElement {
+  const [drawer, setDrawer] = useLocalStorage('storageDrawer');
+  const value = { drawer, setDrawer };
+
+  return (
+    <ThemeProvider theme={createMuiTheme(light)}>
+      <DrawerContext.Provider value={value}>{children}</DrawerContext.Provider>
```
hbulgarini, line 38: This is not bad , but i like the pattern applied on other contexts where i created two contexts in the provider file: one context for values and the other for updating, so you separate concerns on your state ( values & updating )

It would be something like this:

```
    <DrawerContext.Provider value={{drawer}}>
      <UpdateDrawerContext.Provider value={{ setDrawer }}>
         {children}
      </UpdateDrawerContext.Provider>
    </DrawerContext.Provider>
``` 

Also i know that it was highly criticized, but managing the states with reducers makes the context extremely powerful  and simple for the children components using and updating this state. On this last one, is not a hard requirement since the usage is pretty basic.

goldsteinsveta: I see your point and your contexts too. I'd use Context-UpdateContext when reducers are sensible to use at all in this context, if that's ok

wirednkod: > tbh I'm thinking now to get rid of that context and not to merge this at all, as the delay that it takes for drawer to open is just annoyingly long

Im not sure I agree - the open/close of the drawer is pretty fine to my eyes - I dont see any delay or glitch

wirednkod (approved):

src/contexts/GUIContextProvider.tsx
goldsteinsveta: I don't think it's to do with animation, but I'll look into the transitions one more time. Or maybe it's my computer was boiled by the end of the day :)

hbulgarini (approved):
lgtm!

PR #155, opened 2021-05-11
Unused vars: warn/error

goldsteinsveta:
warn about unused vars

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -36,7 +36,7 @@
         }
     },
     "rules": {
-        "@typescript-eslint/no-unused-vars": "error",
+        "@typescript-eslint/no-unused-vars": "warn",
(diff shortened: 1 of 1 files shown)


hbulgarini:
i can understand that is annoying to have the compiler failing for a non used var, but for a CI i think this rule should not be changed.

tomusdrw:
> i can understand that is annoying to have the compiler failing for a non used var, but for a CI i think this rule should not be changed.

Sure, I fully agree. That's why I meant it should be warning, but in CI perhaps there is a way to convert warnings into build failures.

goldsteinsveta:
Sorry, I thought I "set it up real quick", but I never dealt with different lint setups myself, so I just closed it

Issue #80, opened 2021-04-21
Apply design on the receiver component feedback.


hbulgarini:
Apply design on the following section that gives feedback and let the user to choose if the receiver should switch to a derived account:

![Screen Shot 2021-04-21 at 19 45 18](https://user-images.githubusercontent.com/12089572/115630271-1c854b80-a2da-11eb-9b67-10cf69997b2c.png)

The functionality was requested in #22

PR #156, opened 2021-05-12
UI Main and Companion account selectors

### LINKED ISSUE #71: Apply design on account selector.

goldsteinsveta:
addresses #71, #80, #104

Screenshot 2021-05-14 at 11 54 46


src/components/GenericAccount.tsx
```diff
@@ -76,42 +77,57 @@ const GenericAccount = ({ value }: Props) => {
   });
   const companionState = useBalance(targetApi, companionAddress, true);
 
+  const looseHelperAccount = () => {
+    setSelected('');
+    dispatchTransaction(TransactionActionCreators.setReceiverAddress(null));
+  };
+
   const setNativeAsTarget = () => {
+    if (selected) {
```
hbulgarini, line 86: could you please elaborate the purpose of setting the receiver to null at this time? have you verified if this is not breaking all the flow that comes on the other components?

src/components/Network.tsx
```diff
@@ -60,34 +56,40 @@ const useStyles = makeStyles((theme) => ({
...
   const classes = useStyles();
   const { sourceChainDetails, targetChainDetails } = useSourceTarget();
+  const dbSource = useDashboard(ChainDetails.SOURCE);
+  const dbTarget = useDashboard(ChainDetails.TARGET);
+
   return (
     <Box marginY={2} className={classes.main}>
-      <Box p={1}>{sourceChainDetails.sourceChain}</Box>
+      <Box p={1} className={classes.statsEntry}>
+        <Typography variant="h4">{sourceChainDetails.sourceChain}</Typography>
```
hbulgarini, line 65: I think it worth to sync this branch with master since we have put back the links to polkadotJS UI:
https://github.com/paritytech/parity-bridges-ui/blob/master/src/components/Network.tsx#L66

src/components/Network.tsx
```diff
@@ -60,34 +56,40 @@ const useStyles = makeStyles((theme) => ({
...
+  const dbTarget = useDashboard(ChainDetails.TARGET);
 
   return (
     <>
       <Box className={classes.statsEntry}>
-        Finalized blocks:
+        Relayed blocks:
         <span>
-          {dbSource?.bestBlockFinalized} / {dbSource?.bestBlock}
+          {dbTarget?.bestBridgedFinalizedBlock} / {dbSource?.bestBlockFinalized}{' '}
```
hbulgarini, line 90: do we need the `{' '}` ?

src/components/ReceiverDerivedAccount.tsx
```diff
@@ -54,21 +50,30 @@ const ReceiverDerivedAccount = () => {
...
+        <AccountDisplay />
+      </div>
+    );
+  }
+
   return (
-    <div className={classes.derived}>
-      <AccountDisplay addressKind={AddressKind.COMPANION} derived />
+    <div className={classes.accountCompanion}>
+      <AccountDisplay address={receiverAddress!} addressKind={AddressKind.NATIVE} isDerived withTooltip />
```
hbulgarini, line 76: why are you changing the `addressKind` ? shouldn't be 
`<AccountDisplay address={receiverAddress!} addressKind={AddressKind.COMPANION} isDerived withTooltip />` ?

src/components/Sender.tsx
```diff
@@ -14,38 +14,58 @@
...
-import AccountDisplay from './AccountDisplay';
-import { Star } from '@material-ui/icons';
+import AccountDisplay, { AddressKind } from './AccountDisplay';
+import { SelectLabel, styleAccountCompanion } from '../components';
 import useChainGetters from '../hooks/useChainGetters';
 
-interface Props {
-  className?: string;
-}
+// TDOO replace MUI Select with MUI Popover it wraps around or Autocomplete to have more control over appearance
```
hbulgarini, line 32: Minor: TODO typo

src/components/Sender.tsx
```diff
@@ -81,23 +101,21 @@ const Sender = ({ className }: Props) => {
...
       >
-        <div className={classes.row}>
-          <Account accountName={text} value={value} chain={source} />
-          <Account accountName={text} value={value} isDerived hideAddress chain={target} />
-        </div>
+        <Account friendlyName={text} value={value} chain={source} />
+        <Account friendlyName={text} value={value} chain={target} isDerived hideAddress />
       </MenuItem>
     ));
     return [
```
hbulgarini, line 115: Is there anyway to provide a better way in MUI to add headers to a group of options within the select? when i did this i felt it was kind of hacky but it was enough for putting something

src/components/GenericAccount.tsx
goldsteinsveta: This is to make sure that when the companion options "menu" is open, the transaction cannot be submitted.
Imagine you you want to switch from companion to native or vice versa manually. If we don't set the receiver address to null, you'll be able to submit transaction even though you haven't selected the option you wanted to switch to.

I did not test flows. Just set the receiver to null (default value) in the same way the origin/companion are set on select

src/components/ReceiverDerivedAccount.tsx
goldsteinsveta: I understood "companion" is an alias for "derived". In this case we send directly to address user has inputed, not to derived from this input, so it's not "companion". Am I confused?

src/components/Sender.tsx
goldsteinsveta: This is the reason for suggesting a lower Popover component: `// TODO replace MUI Select with MUI Popover it wraps around or Autocomplete to have more control over appearance`. I think it all will be nicer if we won't try to push all the logic in this Select

src/components/GenericAccount.tsx
hbulgarini: but at this time the transfer button should be disabled right? or at least this shouldn't solve the issue?

hbulgarini: All this flow will be refactored anyway ... the only thing is to make sure that any part of the flow is broken

src/components/ReceiverDerivedAccount.tsx
hbulgarini: exactly, but you are replacing NATIVE by COMPANION and i think it should remain as COMPANION right?

src/components/Sender.tsx
hbulgarini: Agree 100% , can you please create a GH issue and link the number in the TODO comment?

src/components/ReceiverDerivedAccount.tsx
goldsteinsveta: This is changes for affected component in affected state:

_before_
Screenshot 2021-05-14 at 16 45 02

_after_
Screenshot 2021-05-14 at 16 44 22

goldsteinsveta: I've just removed isDerived prop

hbulgarini: As the UX was enhanced . This change is valid.

src/components/GenericAccount.tsx
hbulgarini: As the UX was enhanced . This change is valid.

hbulgarini (approved):
Looks good to me! pretty nice enhancements to UI :)

Just address minor comments and we are good to merge.

PR #167, opened 2021-05-13
Properly display target delivery finalized block.

### LINKED ISSUE #113: Properly display target delivery finalized block.
Currently in `TransactionStatus` component, we simply display `bestFinalizedBlockOnTarget` in case we detect that the message delivery has been finalized.

Instead when we detect finality we should memoize the block that we delivered in and display that.

hbulgarini:


tomusdrw (approved):
Looks good to me, however as discussed we should move status tracking to a global state (via reducer), so that it is not required to display a particular component in order to have the transactio n tracked separately. The reason for that is that if that component comes and goes, the state will now be lost and then when it re-appears it will not be correct.

So this PR is an improvement over current state, but does not fix #113 entirely.

Issue #159, opened 2021-05-12
Refactor by adding OverrideSourceTarget around TransactionStatus component.


hbulgarini:
The component needs to know that it might be used in different context and needs to 'swap' chains. It's a design that makes it super easy to introduce subtle bugs by just adding a nested component which simply uses source chain instead of source chain flipped into target chain.

Solution to explore:
```
<OverrideSourceTarget source={target} target={source}>
   <TransactionStatus>
</OverrideSourceTarget>
```

hbulgarini:
https://github.com/paritytech/parity-bridges-ui/pull/140#discussion_r630134362

hbulgarini:
To consider: https://github.com/paritytech/parity-bridges-ui/pull/179/files#r644043221

PR #169, opened 2021-05-15
Fixing  steps inconsistences.

hbulgarini:
Transaction steps were being incorrectly being updated due to a regression introduced in PR #140.

This PR fixes the current behavior that  step `Finalize message in target block` was never completing even when last step is ready. After these fixes i was not able to reproduce again the issue that `Delivered Message` gets finished before the Relay message.  

What it is not solving is: Balance gets updated before delivered message step is finished.  I'm tempted to think that the reason of this issue is because the delivered message step is related to reading the storage related with the bridge and reading balance is a different process as far as i understand ( @tomusdrw can you see any more specific scenario based on my assumption ? )

### DIFF
--- a/src/hooks/useTransactionNonces.ts
+++ b/src/hooks/useTransactionNonces.ts
@@ -16,7 +16,7 @@
 
 import { Codec } from '@polkadot/types/types';
 
-import { useEffect } from 'react';
+import { useCallback, useEffect } from 'react';
 
 import { useSourceTarget } from '../contexts/SourceTargetContextProvider';
 import useDashboard from './useDashboard';
@@ -34,7 +34,6 @@ interface Props {
 }
 
 const useTransactionNonces = ({ transaction }: Props) => {
-  const [nonceOfTargetFinalizedBlock, setNonceOfTargetFinalizedBlock] = useMountedState<null | number>(null);
   const [latestReceivedNonceRuntimeApi, setLatestReceivedNonceRuntimeApi] = useMountedState(0);
   const { getValuesByChain } = useChainGetters();
 
@@ -51,12 +50,9 @@ const useTransactionNonces = ({ transaction }: Props) => {
   const { latestReceivedNonceMethodName } = getSubstrateDynamicNames(sourceChain);
   const { api: targetApi } = getValuesByChain(targetChain);
 
-  useEffect(() => {
-    if (!areApiLoading || !transaction || !transaction || isTransactionCompleted(transaction)) {
-      return;
-    }
-    const getNonceByHash = async () => {
-      const blockHash = await targetApi.rpc.chain.getBlockHash(bestBlockFinalizedOnTarget);
+  const getNonceByHash = useCallback(
+    async (targetBlock: number) => {
+      const blockHash = await targetApi.rpc.chain.getBlockHash(targetBlock);
       const latestReceivedNonceCall = await targetApi.rpc.state.call<Codec>(
         latestReceivedNonceMethodName,
         laneId,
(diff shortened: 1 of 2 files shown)


src/hooks/useTransactionSteps.ts
```diff
@@ -77,12 +93,12 @@ const useTransactionSteps = ({ transaction, onComplete }: Props) => {
     const sourceTransactionFinalized = stepEvaluator(transaction.block, bestBlockFinalized);
     const blockFinalityRelayed = stepEvaluator(transaction.block, bestBridgedFinalizedBlockOnTarget);
     const messageDelivered = stepEvaluator(transaction.messageNonce, latestReceivedNonceRuntimeApi);
-    const messageFinalizedOnTarget = stepEvaluator(transaction.messageNonce, targetMessageDeliveryBlock);
+    const messageFinalizedOnTarget = stepEvaluator(transactionNonceOfTargetFinalizedBlock, nonceOfCurrentTargetBlock);
```
hbulgarini, line 96: 2) Comparing apples with pears. I was comparing the transaction message nonce with a block. 
I didn't catch this because when chains are restarted this conditions were satisfied.

src/hooks/useTransactionNonces.ts
```diff
@@ -51,12 +51,9 @@ const useTransactionNonces = ({ transaction }: Props) => {
   const { latestReceivedNonceMethodName } = getSubstrateDynamicNames(sourceChain);
   const { api: targetApi } = getValuesByChain(targetChain);
 
-  useEffect(() => {
-    if (!areApiLoading || !transaction || !transaction || isTransactionCompleted(transaction)) {
-      return;
-    }
-    const getNonceByHash = async () => {
-      const blockHash = await targetApi.rpc.chain.getBlockHash(bestBlockFinalizedOnTarget);
+  const getNonceByHash = useCallback(
```
hbulgarini, line 54: 3) In the past i was calling this function every time `bestBlockFinalizedOnTarget` got updated. As now i need to call this function specifically  i obtain the block of the delivered message event, i had to expose it to the parent hook so this process can be run at that time.

src/hooks/useTransactionSteps.ts
```diff
@@ -77,12 +93,12 @@ const useTransactionSteps = ({ transaction, onComplete }: Props) => {
...
     const sourceConfirmationReceived = stepEvaluator(transaction.messageNonce, latestReceivedNonceOnSource);
     const onChainCompleted = (value: boolean) => completionStatus(value) === TransactionStatusEnum.COMPLETED;
 
     if (messageDelivered && !targetMessageDeliveryBlock) {
-      setTargetMessageDeliveryBlock(bestBridgedFinalizedBlockOnTarget);
+      setTargetMessageDeliveryBlock(bestBlockFinalizedOnTarget);
```
hbulgarini, line 101: 1) For some reason i was setting the bridged block of the target chain , what i needed was the `bestBlockFinalizedOnTarget`

src/hooks/useTransactionSteps.ts
```diff
@@ -54,6 +59,17 @@ const useTransactionSteps = ({ transaction, onComplete }: Props) => {
     bestBlockFinalized: bestBlockFinalizedOnTarget
   } = useDashboard(targetRole);
 
+  useEffect(() => {
+    const getNonceOfTargetFinalizedBlock = async () => {
```
hbulgarini, line 63: 4) Having now the `bestBlockFinalizedOnTarget` related with message delivered and the function to get the related nonce, i'm able to get it and save in the state ( to be moved this during the refactor to a global variable as indicated in previous PR)

src/hooks/useTransactionNonces.ts
```diff
@@ -76,7 +80,12 @@ const useTransactionNonces = ({ transaction }: Props) => {
       setLatestReceivedNonceRuntimeApi(parseInt(latestReceivedNonceCallType.toString()));
     };
 
-    getNonceByHash();
+    const getNonceOfCurrentBlock = async () => {
```
hbulgarini, line 83: 5) But i still need to know the nonce of the current block , otherwise i would not be able to infer if the message was processed or not on the target chain ( this is the same process as before but with a more adequate name )

src/hooks/useTransactionSteps.ts
hbulgarini: 6 & END: the message is finalized once we verify that current nonce of the current block is bigger than the nonce obtained from the delivered message block.

src/hooks/useTransactionNonces.ts
tomusdrw: I'm a bit confused, it's still being called every time `bestBlockFinalizedOnTarget` changes, cause you are invoking that function in `useEffect`. Am I missing something?

tomusdrw: Also the fucntion should be rather be called `getLatestReceivedNonceByHash`, cause now one might wonder what kind of nonce is this.

tomusdrw (approved):

src/hooks/useTransactionNonces.ts
hbulgarini: I'm calling that function twice :

1) one for getting the message nonce related to the specific block i needed 
2) Then i'm calling it every time a `bestBlockFinalizedOnTarget` arrives  so i can match it with the step 1) and infer that step is completed.

Anyway, now i realize that i might avoid the step 1 by improving the logic.
Also mentioning  this in the issue #159

Issue #170, opened 2021-05-17
Receiver balance gets updated before the delivery message step is completed.


hbulgarini:

hbulgarini:
#closed 172

PR #172, opened 2021-05-17
Fix balance update before deliver step.

hbulgarini:
This issue solves #170 and some part of the issue described in #169 and #113.

Now all the values are correctly displayed / updated on time:

Screen Shot 2021-05-17 at 19 05 15
Screen Shot 2021-05-17 at 19 05 23

### DIFF
--- a/src/hooks/useTransactionSteps.ts
+++ b/src/hooks/useTransactionSteps.ts
@@ -37,8 +37,6 @@ interface MessageOnTarget {
 
 const useTransactionSteps = ({ transaction, onComplete }: Props) => {
   const [steps, setSteps] = useState<Array<Step>>([]);
-  //const [messageOnTarget.nonce, setTransactionNonceOfTargetFinalizedBlock] = useMountedState<null | (number > null);
-  //const [messageOnTarget.block, setTargetMessageDeliveryBlock] = useMountedState('');
   const [messageOnTarget, setMessageOnTarget] = useMountedState({} as MessageOnTarget);
   const [finished, setFinished] = useState(false);
 
@@ -69,19 +67,13 @@ const useTransactionSteps = ({ transaction, onComplete }: Props) => {
       return;
     }
 
-    const stepEvaluator = (
-      transactionValue: string | number | null,
-      chainValue: string | number | null,
-      greaterEqual?: boolean
-    ): boolean => {
+    const stepEvaluator = (transactionValue: string | number | null, chainValue: string | number | null): boolean => {
       if (!transactionValue || !chainValue) return false;
 
       const bnChainValue = new BN(chainValue);
       const bnTransactionValue = new BN(transactionValue);
-      if (greaterEqual) {
-        return bnChainValue.gte(bnTransactionValue);
-      }
-      return bnChainValue.gt(bnTransactionValue);
+
+      return bnChainValue.gte(bnTransactionValue);
     };
 
     const completionStatus = (isCompleted: boolean): TransactionStatusEnum => {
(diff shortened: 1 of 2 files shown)


src/hooks/useTransactionSteps.ts
```diff
@@ -90,15 +91,24 @@ const useTransactionSteps = ({ transaction, onComplete }: Props) => {
       return isCompleted ? TransactionStatusEnum.COMPLETED : TransactionStatusEnum.IN_PROGRESS;
     };
 
+    // 1. We wait until the block transaction gets finalized / transaction.block < source.bestBlockFinalized
```
tomusdrw, line 94: I don't think it has to be strictly greater. if `transaction.block == source.bestBlockFinalized` it is already finalized.

src/hooks/useTransactionSteps.ts
```diff
@@ -90,15 +91,24 @@ const useTransactionSteps = ({ transaction, onComplete }: Props) => {
...
     const sourceTransactionFinalized = stepEvaluator(transaction.block, bestBlockFinalized);
+    // 2. When the target chain knows about a bigger source block number we infer that transaction block was realayed to target chain.
     const blockFinalityRelayed = stepEvaluator(transaction.block, bestBridgedFinalizedBlockOnTarget);
```
tomusdrw, line 97: Same here, we should be using `>=` not `>`

src/hooks/useTransactionSteps.ts
```diff
@@ -59,27 +64,23 @@ const useTransactionSteps = ({ transaction, onComplete }: Props) => {
...
   useEffect(() => {
     if (!areApiLoading || !transaction || finished) {
       return;
     }
 
-    const stepEvaluator = (transactionValue: string | number | null, chainValue: string | number | null): boolean => {
+    const stepEvaluator = (
+      transactionValue: string | number | null,
+      chainValue: string | number | null,
+      greaterEqual?: boolean
```
tomusdrw, line 75: This should be the only option. I think in all cases it's enough when we have `>=`.

hbulgarini:
@goldsteinsveta @wirednkod , in case you are interested in see how the transaction steps are fixed.

src/hooks/useTransactionSteps.ts
tomusdrw: The comment still needs an update.

src/hooks/useTransactionSteps.ts
```diff
@@ -121,42 +117,43 @@ const useTransactionSteps = ({ transaction, onComplete }: Props) => {
...
-        labelOnChain: onChainCompleted(messageDelivered) && transaction.messageNonce,
+        labelOnChain: onChainCompleted(messageDelivered) && deliveryBlock,
         status: completionStatus(messageDelivered)
       },
       {
         chainType: targetChain,
         label: 'Finalize message in target block',
-        labelOnChain: onChainCompleted(messageFinalizedOnTarget) && targetMessageDeliveryBlock,
+        // TODO: To apply a different design for blocks and nonces: #175
+        labelOnChain: onChainCompleted(messageFinalizedOnTarget) && deliveryBlock,
```
tomusdrw, line 127: We shouldn't be displaying the same block twice (first in delivery then in finality), let's just not display anything as in source finality.

src/hooks/useTransactionSteps.ts
```diff
@@ -121,42 +117,43 @@ const useTransactionSteps = ({ transaction, onComplete }: Props) => {
...
         status: completionStatus(messageFinalizedOnTarget)
       },
       {
         chainType: sourceChain,
         label: 'Confirm delivery',
-        status: completionStatus(sourceConfirmationReceived)
+        status: completionStatus(messageFinalizedOnTarget && sourceConfirmationReceived)
```
tomusdrw, line 133: Why is this extra check needed? Like it is not possible to have confirmation before finality.

src/hooks/useTransactionSteps.ts
```diff
@@ -121,42 +117,43 @@ const useTransactionSteps = ({ transaction, onComplete }: Props) => {
...
       }
     ]);
 
-    if (sourceConfirmationReceived) {
+    if (messageFinalizedOnTarget && sourceConfirmationReceived) {
```
tomusdrw, line 137: same here

src/hooks/useTransactionSteps.ts
hbulgarini: why duplicating? in the first step we show the finalized block on the source and in this step we show the finality block of the target...

If you feel showing the information of finalized blocks on both  chains is redundant , then i can remove it with no issues. @goldsteinsveta , do you have any comments on this?

tomusdrw: > why duplicating? in the first step we show the finalized block on the source and in this step we show the finality block of the
 
image

Do you see `6405` twice? That's exactly the duplication I mean.

We don't display `6161` twice - it's only included when message is included in block.

Perhaps we should change labels to:
```
- Deliver message in target block 6405
- Finalize message
```
to make it even more clear.

Displaying `nonce` is only useful for debugging really - it should be of no use to the end user, I'd suggest putting it to a tooltip (or some future place where we display more details about the transaction).

hbulgarini: ouh! i didn't catch that ... actually it was a regression on this PR. 

The main idea is:
- Deliver message  <NONCE>
- Finalize message in target block <BLOCK>

as soon as i see we are not showing the message nonce anywhere.

src/hooks/useTransactionSteps.ts
hbulgarini: updated.

src/hooks/useTransactionSteps.ts
hbulgarini: never mind ... your proposal makes sense.

tomusdrw (approved):

PR #173, opened 2021-05-18
Fix how UI reacts when APIs are not connected

### LINKED ISSUE #15: Avoid page breaking when the API is not connected.

wirednkod:
Fixes #15 

Besides avoiding the breaking of UI when API is disconnected, enhancement has been added concerning the fact that 1 of the 2 APIs is not connected and not only both. 
Relevant Icons were added next to chain names at the sidebar.

connected
disconnected
source only disconnected
Screenshot from 2021-05-18 14-09-00

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -27,13 +27,14 @@ import { SourceTargetContextProvider } from './contexts/SourceTargetContextProvi
 import { TransactionContextProvider } from './contexts/TransactionContext';
 import { useConnections } from './hooks/useConnections';
 import Main from './screens/Main';
+import { isEmpty } from 'lodash';
 
 const [connectionDetails1, connectionDetails2] = substrateProviders();
 
 function App() {
   const { connections } = useConnections([connectionDetails1, connectionDetails2]);
 
-  if (Object.keys(connections).length === 0) {
+  if (isEmpty(connections)) {
     return (
       <Backdrop open>
         <CircularProgress color="inherit" />
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -27,14 +27,13 @@ import { SourceTargetContextProvider } from './contexts/SourceTargetContextProvi
 import { TransactionContextProvider } from './contexts/TransactionContext';
 import { useConnections } from './hooks/useConnections';
 import Main from './screens/Main';
-import { isEmpty } from 'lodash';
 
 const [connectionDetails1, connectionDetails2] = substrateProviders();
 
 function App() {
   const { connections } = useConnections([connectionDetails1, connectionDetails2]);
 
-  if (isEmpty(connections)) {
+  if (Object.keys(connections).length === 0) {
     return (
       <Backdrop open>
         <CircularProgress color="inherit" />
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -27,13 +27,14 @@ import { SourceTargetContextProvider } from './contexts/SourceTargetContextProvi
 import { TransactionContextProvider } from './contexts/TransactionContext';
 import { useConnections } from './hooks/useConnections';
 import Main from './screens/Main';
+import { isEmpty } from 'lodash';
 
 const [connectionDetails1, connectionDetails2] = substrateProviders();
 
 function App() {
-  const { connections, apiReady } = useConnections([connectionDetails1, connectionDetails2]);
+  const { connections } = useConnections([connectionDetails1, connectionDetails2]);
 
-  if (!apiReady) {
+  if (isEmpty(connections)) {
     return (
       <Backdrop open>
         <CircularProgress color="inherit" />
--- a/src/components/ExtensionAccountCheck.tsx
+++ b/src/components/ExtensionAccountCheck.tsx
@@ -22,12 +22,12 @@ interface Props {
   component: JSX.Element;
 }
 
-// TODO: Move this to a more generic error-show component
+// TODO #176: Move this to a more generic error-show component
 const statusFunc = (from: string, state: boolean) => `${from} chain status: ${!state ? 'disconnected' : 'connected'}`;
 
 const ExtensionAccountCheck = ({ component }: Props): JSX.Element => {
   const { extensionExists, accountExists } = useKeyringContext();
-  // TODO: Move this to a more generic error-show component
+  // TODO #176: Move this to a more generic error-show component
   const { sourceReady, targetReady } = useLoadingApi();
 
   let msg: string = '';
@@ -35,7 +35,7 @@ const ExtensionAccountCheck = ({ component }: Props): JSX.Element => {
     msg = 'Connect to a wallet. Install polkadotjs extension';
   } else if (!accountExists) {
     msg = 'There are no accounts in the extension. Please create one';
-    // TODO: Move this to a more generic error-show component
+    // TODO #176: Move this to a more generic error-show component
   } else if (!sourceReady || !targetReady) {
     msg = `${!sourceReady ? statusFunc('Source', sourceReady) : ''} ${
       !targetReady ? statusFunc('Target', targetReady) : ''
--- a/src/components/ExtensionAccountCheck.tsx
+++ b/src/components/ExtensionAccountCheck.tsx
@@ -42,7 +42,7 @@ const ExtensionAccountCheck = ({ component }: Props): JSX.Element => {
     }`;
   }
 
-  return <>{accountExists && sourceReady && targetReady ? component : <Alert severity="error">{msg}</Alert>}</>;
+  return <>{msg ? <Alert severity="error">{msg}</Alert> : component}</>;
 };
 
 export default ExtensionAccountCheck;
--- a/src/components/ExtensionAccountCheck.tsx
+++ b/src/components/ExtensionAccountCheck.tsx
@@ -23,12 +23,12 @@ interface Props {
 }
 
 // TODO: Move this to a more generic error-show component
-const statusFunc = (from: string, state: boolean) => `${from} status: ${!state && 'dis'}connected.`;
+const statusFunc = (from: string, state: boolean) => `${from} chain status: ${!state ? 'disconnected' : 'connected'}`;
 
 const ExtensionAccountCheck = ({ component }: Props): JSX.Element => {
   const { extensionExists, accountExists } = useKeyringContext();
   // TODO: Move this to a more generic error-show component
-  const { areApiReady, sourceReady, targetReady } = useLoadingApi();
+  const { sourceReady, targetReady } = useLoadingApi();
 
   let msg: string = '';
   if (!extensionExists) {
@@ -36,11 +36,13 @@ const ExtensionAccountCheck = ({ component }: Props): JSX.Element => {
   } else if (!accountExists) {
     msg = 'There are no accounts in the extension. Please create one';
     // TODO: Move this to a more generic error-show component
-  } else if (!areApiReady) {
-    msg = `${statusFunc('Source', sourceReady)} ${statusFunc('Target', targetReady)}`;
+  } else if (!sourceReady || !targetReady) {
+    msg = `${!sourceReady ? statusFunc('Source', sourceReady) : ''} ${
+      !targetReady ? statusFunc('Target', targetReady) : ''
+    }`;
   }
 
-  return <>{accountExists && areApiReady ? component : <Alert severity="error">{msg}</Alert>}</>;
+  return <>{accountExists && sourceReady && targetReady ? component : <Alert severity="error">{msg}</Alert>}</>;
 };
 
 export default ExtensionAccountCheck;
--- a/src/components/ExtensionAccountCheck.tsx
+++ b/src/components/ExtensionAccountCheck.tsx
@@ -16,22 +16,31 @@
 import React from 'react';
 import { Alert } from '.';
 import { useKeyringContext } from '../contexts/KeyringContextProvider';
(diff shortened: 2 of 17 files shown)


src/components/ExtensionAccountCheck.tsx
```diff
@@ -16,22 +16,33 @@
...
   component: JSX.Element;
 }
 
+// TODO: Move this to a more generic error-show component
+const statusFunc = (from: string, state: boolean) => `${from} chain status: ${!state ? 'disconnected' : 'connected'}`;
+
 const ExtensionAccountCheck = ({ component }: Props): JSX.Element => {
   const { extensionExists, accountExists } = useKeyringContext();
+  // TODO: Move this to a more generic error-show component
+  const { sourceReady, targetReady } = useLoadingApi();
```
hbulgarini, line 31: This values are already available on `useSourceTarget`, but i'm fine if you want to retrieve them from the `useLoadingApi` for readability

src/hooks/useApiConnection.ts
```diff
@@ -78,7 +78,7 @@ export function useApiConnection(connectionDetails: ConnectionChainInformation):
       setConfigs(values);
     };
 
-    if (isReady && isEmpty(configs)) {
+    if (apiPromise.isReady && isEmpty(configs)) {
       getChainConfigs();
     }
   }, [apiPromise, configs, isReady]);
```
hbulgarini, line 84: do we need `isReady` dependency?

src/hooks/useLoadingApi.ts
```diff
@@ -29,10 +35,18 @@ export default function useLoadingApi(): boolean {
   } = useSourceTarget();
 
   const [areReady, setAreReady] = useState(false);
+  const [isSourceReady, setIsSourceReady] = useState(false);
```
hbulgarini, line 38: Can you please elaborate why are we replicating the value coming from `useSourceTarget` to this local state?

hbulgarini (changes requested):

src/components/Network.tsx
```diff
@@ -58,11 +65,13 @@ export const NetworkSides = () => {
   const { sourceChainDetails, targetChainDetails } = useSourceTarget();
   const dbSource = useDashboard(ChainDetails.SOURCE);
   const dbTarget = useDashboard(ChainDetails.TARGET);
+  const { sourceReady, targetReady } = useLoadingApi();
```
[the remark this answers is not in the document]
wirednkod: Yes and no. I prefer to have a useLoadingApi() (as soon as it already existed) - that I will be able to know at once the state of both chains and of each one of them. I just extended the functionality. Also useSourceTarget is not used everywhere where the useLoading is needed.

src/components/ExtensionAccountCheck.tsx
wirednkod: Its not only a matter of readability but also of simplicity.
useSourceTarget carries much more than needed while useLoadingApi hook is straightforward.
For the first you need minimum to do:
```
const {
    sourceChainDetails: {
      sourceApiConnection: { isApiReady: sourceIsApiReady }
    },
    targetChainDetails: {
      targetApiConnection: {  isApiReady: targetIsApiReady }
    }
  } = useSourceTarget();
  ```
  while for the latter its less code and simpler.

File: src/hooks/useApiConnection.ts (hunk 2 above)
Comment (wirednkod) [reply src/hooks/useApiConnection.ts:84]: Yes you do. If `apiPromise.isReady` is removed then on `getConfigs` you will receive on:
```
const systemChain = await apiPromise.rpc.system.name();
```
a `rpc of undefined not found` or something like that (happened to me)

File: src/hooks/useLoadingApi.ts (hunk 3 above)
Comment (wirednkod) [reply src/hooks/useLoadingApi.ts:38]: Makes importing and re-usability easy. Imagine you have a file that both areReady and each one API state state is needed.
You will have to import both `useLoading` and `useSourceTarget` hooks in order to get all Values.
As soon as useLoading hook exist and is used in order to ensure that API are ready for usage - then it made much more sense to extend it and reuse it than import yet another hook to retrieve info that I could get from the same hook

File: src/App.tsx [hunk 5]
```diff
@@ -27,13 +27,14 @@ import { SourceTargetContextProvider } from './contexts/SourceTargetContextProvi
...
+import { isEmpty } from 'lodash';
 
 const [connectionDetails1, connectionDetails2] = substrateProviders();
 
 function App() {
-  const { connections, apiReady } = useConnections([connectionDetails1, connectionDetails2]);
+  const { connections } = useConnections([connectionDetails1, connectionDetails2]);
 
-  if (!apiReady) {
+  if (isEmpty(connections)) {
```
tomusdrw, line 37: Under what circumstances can this happen?

src/App.tsx
```diff
@@ -27,13 +27,14 @@ import { SourceTargetContextProvider } from './contexts/SourceTargetContextProvi
 import { TransactionContextProvider } from './contexts/TransactionContext';
 import { useConnections } from './hooks/useConnections';
 import Main from './screens/Main';
+import { isEmpty } from 'lodash';
```
tomusdrw, line 30: Is it really needed? Like isn't it just doing `parameter.length === 0`?

src/components/ExtensionAccountCheck.tsx
```diff
@@ -16,22 +16,33 @@
...
     msg = 'There are no accounts in the extension. Please create one';
+    // TODO: Move this to a more generic error-show component
+  } else if (!sourceReady || !targetReady) {
+    msg = `${!sourceReady ? statusFunc('Source', sourceReady) : ''} ${
+      !targetReady ? statusFunc('Target', targetReady) : ''
+    }`;
   }
 
-  return <>{accountExists ? component : <Alert severity="error">{msg}</Alert>}</>;
+  return <>{accountExists && sourceReady && targetReady ? component : <Alert severity="error">{msg}</Alert>}</>;
```
tomusdrw, line 45:
```suggestion
  return <>{msg ? <Alert severity="error">{msg}</Alert> : component}</>;
```

src/components/Transactions.tsx
```diff
@@ -37,7 +37,7 @@ const Transactions = ({ type, ...transactionDisplayProps }: Props) => {
 
   return (
     <>
-      {transactions.length ? (
+      {transactions?.length ? (
```
tomusdrw, line 40: Can it be `null/undefined` now? Why it wasn't before this PR?

src/hooks/useApiConnection.ts
tomusdrw: We probably need `[apiPromise, apiPromise.isReady, configs]` instead.

src/hooks/useLoadingApi.ts
tomusdrw: The 2 extra `useState`s for that are not necessary, we can simply return `isSourceApiReady` and `isTargetApiReady` (we do need the state variable for `areReady` though).

src/App.tsx
wirednkod: When chain1Configs or chain2Configs is empty -> meaning for each chain - when the API is not connected:
https://github.com/paritytech/parity-bridges-ui/blob/5e8851725cb774981b88ef3c4fea8732978ffede/src/hooks/useApiConnection.ts#L81
If this does not happen, then getChainConfigs will not run, and thus configs will be empty...

src/App.tsx
wirednkod: It is needed as connections is a Json object and that would not work.
What I could do is `Object.keys(parameter).length === 0`. That would work - should I change it?

src/components/ExtensionAccountCheck.tsx
wirednkod: ............. you got me there

src/components/Transactions.tsx
wirednkod: It collapsed when I removed the API connections for test reasons. It probably remained - I'll remove it (even though I think that there is no harm for it to be there)

wirednkod:
https://github.com/paritytech/parity-bridges-ui/pull/173#discussion_r634521338

As soon as isReady is a boolean (primitive - and we are using Typescript - meaning that it probably wont change) then this is not actually needed. Even though "it is said" that as a primitive dependency, it should work just fine - that is not a guarantee and it is better always to add the dependencies - so I'll add this one.

src/hooks/useLoadingApi.ts
wirednkod: I prefer to know the init state upon first call of the hook (false) instead of initializing the states from another state that is not known at the moment. The performance is the same as the 1st useEffect will run anyway and more control is provided.
If we do not want that then no state is needed for any of the cases. We can just: 
```
return {
    areApiReady: isSourceApiReady && isTargetApiReady,
    sourceReady: isSourceApiReady,
    targetReady: isTargetApiReady
  };
  ```
  but I do not like it as it is too float-y

**Comment by wirednkod** [comment]
@tomusdrw concerning https://github.com/paritytech/parity-bridges-ui/pull/173#discussion_r634523303

I do not agree. The primitive hooks (useState, useEffect etc) should be used inside a custom hook in order to verify that the rerender of components using them is happening efficiently.

File: src/hooks/useLoadingApi.ts (hunk 3 above)
Comment (hbulgarini) [reply src/hooks/useLoadingApi.ts:38]: > The 2 extra `useState`s for that are not necessary, we can simply return `isSourceApiReady` and `isTargetApiReady` (we do need the state variable for `areReady` though).

This is my exactly my point.

File: src/App.tsx (hunk 6 above)
Comment (tomusdrw) [reply src/App.tsx]: Ah I see. Fair enough, let's keep `isEmpty`.

File: src/components/Transactions.tsx (hunk 8 above)
Comment (tomusdrw) [reply src/components/Transactions.tsx]: I'd rather make sure we use correct typings so that this kind of stuff is not necessary. I'm not against it, just the change seemed a bit random.

File: src/hooks/useApiConnection.ts (hunk 2 above)
Comment (tomusdrw) [reply src/hooks/useApiConnection.ts]: @wirednkod `isReady` is not used within the `useEffect`, so I feel there is something weird happening here. Like obviously it works, it just indicates something fishy. Why do we check `apiPromise.isReady` and not `isReady`? Shouldn't the two be synchronized anyway?

File: src/App.tsx (hunk 6 above)
Comment (wirednkod) [reply src/App.tsx]: I prefer to avoid 3rd party libraries where I can (and here I can :) )

File: src/hooks/useLoadingApi.ts (hunk 3 above)
Comment (tomusdrw) [reply src/hooks/useLoadingApi.ts:38]: > but I do not like it as it is too float-y. 

Can you elaborate what problem exactly this creates? I'm abstracting from all the bloat that comes from (`useSourceTarget -> Context -> `connections` passed as props -> `useConnections` hook), but essentially we already have a "state" that we maintain in the top level component that knows if api is ready or not. The hook that you propose is creating local copies of these variables, so now instead of having one "source of truth" we have two and you have to use `useEffect` to keep them in sync.
Obviously, we can argue that the thanks to this the state is more local and it's maybe easier to reason about the hook itself, but long term it's much harder to reason about the entire application - this is just covering bloated code with even more bloated code and we will easily loose track of what is happening where. These kind of things lead to subtle bugs (one component thinks "ready", while other is "not ready") and I feel that historically a solution to these kinds of bugs is just adding more bloat (we miss `useEffect` here, or a dependency in existing `useEffect`, etc).

File: src/hooks/useApiConnection.ts (hunk 2 above)
Comment (wirednkod) [reply src/hooks/useApiConnection.ts]: The `isReady` is expecting the `apiPromise.isReady` to be true and the `configs` to be empty in order to retrieve them.
On line `59` (https://github.com/paritytech/parity-bridges-ui/blob/5e8851725cb774981b88ef3c4fea8732978ffede/src/hooks/useApiConnection.ts#L59-L62) there is a similar check but where `isReady` is false. If here we checked the `isReady` we must wait for `apiPromise.isReady` + 1 render to make the `isReady` true, in order to finally get the `configs`. There is no reason to do that - saves us 1 extra render.

File: src/hooks/useLoadingApi.ts (hunk 3 above)
Comment (wirednkod) [reply src/hooks/useLoadingApi.ts:38]: Agreed and understood your point. I will remove the non-needed useState/useEffect from this component. 
At the end - this component is just a way to "receive" simpler and faster the states of API across the app.

**Review by hbulgarini — APPROVED** [review]
awesome! great work

**Comment by hbulgarini** [comment]
@wirednkod  please resolve the conflicts before merge.

File: src/components/Sender.tsx [hunk 9]
```diff
@@ -130,11 +133,14 @@ const Sender = () => {
     return <AccountDisplay friendlyName="Select sender account" hideAddress />;
   };
 
+  console.log('areApiReady', areApiReady);
```
hbulgarini, line 136: unwanted debug

hbulgarini (changes requested):

src/components/ExtensionAccountCheck.tsx
```diff
@@ -16,22 +16,33 @@
 import React from 'react';
 import { Alert } from '.';
 import { useKeyringContext } from '../contexts/KeyringContextProvider';
+import useLoadingApi from '../hooks/useLoadingApi';
 
 interface Props {
   component: JSX.Element;
 }
 
+// TODO: Move this to a more generic error-show component
```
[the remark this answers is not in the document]
tomusdrw: Please add issue number to `TODO`,

tomusdrw (approved):

hbulgarini (approved):

Issue #174, opened 2021-05-18
Optimize subscription count


tomusdrw:
Since we rely heavily on hooks, I think we do set up **a lot** of subscriptions for the same data. Pretty much each component using any of the hooks will set up a separate subscriptions for basic things like: best block on source/target (via `useDashboard`), etc.

We should instead make sure that we keep the subscription count and RPC queries at minimal possible level to make sure we can switch to a light client / substrate-connect in the future.

hbulgarini:
@todr : do you think that the `useDashboard` queries should be moved to the  `sourceTarget` context ? Also i'm thinking that context can be renamed to `ChainContext`

tomusdrw:
I'm not sure if `sourceTarget` context is set up properly, it seems it relies on `useConnections`. I think we need to re-think the entire architecture and decide what is meant to be part of `state` and what should got to contexts. So it's a bit bigger discussion to have if we want to do a bit of refactoring, we can also think of some intermediate solutions instead.

hbulgarini:
hmm i don't believe the sourceTarget chain is incorrectly designed. Is in charge of:

- building connections 
- produce configuration per chain  
- keep the source  <> target direction. 

I like all the benefits of this context used across the components.

What i don't have clear is if we could add a new object to this context containing the chain info ( called today `useDashboard` ) or build a new context specifically for this.

hbulgarini:
@goldsteinsveta @wirednkod , including you in discussion

hbulgarini:
> I believe that the architecture of the app needs rethinking as well. There are way too many hooks and WAY (I mean a LOT) dependencies of hooks from other hooks

Which is not wrong ... building hooks legos is a way to reduce and re use code.

The main problem is when there are hooks forcing unwanted updates of the global state

hbulgarini:
@tomusdrw  could you please let me know if you consider that after the PR #179 this can be closed? It is not implementing a really sophisticated process to count the existing subscriptions , but i think that the main issue was tackled.

tomusdrw:
All of the "global" subscriptions should be okay now. I quickly checked the codebase and `useBalance` is still a bit problematic (used in 4 components).
I'm fine to close this issue and re-open more specific one for balances, cause it's a bit more tricky, since the subscription depends on the address and is not so easily moved to a global context.

[...]

Issue #162, opened 2021-05-13
Avoid creating new `importedHeader` subscription everytime a bestFinalized header is read 


hbulgarini:
https://github.com/paritytech/parity-bridges-ui/pull/145#discussion_r631089275

Issue #149, opened 2021-05-10
Simplfy types in `sourceTargetTypes`.


tomusdrw:
Follow up on: https://github.com/paritytech/parity-bridges-ui/pull/136/files#r626468506

The amount of types seems excessive.

PR #179, opened 2021-05-25
Refactor chain subscription process and sourceTarget context.

hbulgarini:
- Chain subscription moved to a specific on chain values related context file.
- Simplified state for sourceTarget context.
- Split bestFinalized & importedHeaders so we don't spawn a new subscription every time a new bestFinalized is received.
- Rolledback subscriptions to promises because of issues with cleanup return function.

Related to #174, #162 & #149


src/components/Network.tsx
```diff
@@ -85,36 +84,35 @@ export const NetworkSides = () => {
...
       </Box>
     </Box>
   );
 };
 
 export const NetworkStats = () => {
   const classes = useStyles();
-  const dbSource = useDashboard(ChainDetails.SOURCE);
-  const dbTarget = useDashboard(ChainDetails.TARGET);
+  const { sourceChainValues, targetChainValues } = useOnChainValuesContext();
```
hbulgarini, line 99: useDashboard hook was removed from dependent components and corresponding values are read now from a context.

src/hooks/useBridgedBlocks.ts
```diff
@@ -37,30 +37,43 @@ interface Props {
...
       }
     );
 
     return async (): Promise<void> => {
       unsubBestFinalized && (await unsubBestFinalized)();
-      unsubImportedHeaders && (await unsubImportedHeaders)();
     };
   }, [isApiReady, chain, api, bridgedGrandpaChain]);
 
+  useEffect((): (() => void) | undefined => {
```
hbulgarini, line 59: each subscription is run from specific useEffect

src/types/sourceTargetTypes.ts
```diff
@@ -60,21 +73,15 @@ export enum ChainDetails {
   SOURCE = 'sourceChainDetails',
   TARGET = 'targetChainDetails'
 }
-export interface SourceState {
```
hbulgarini, line 63: After some time it was no point having specific objects named per source / target like `sourceApiConnection ` or `targetConfigs` so i unified everything in one common type.

src/contexts/OnChainValuesContextProvider.tsx
```diff
@@ -0,0 +1,44 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
```
hbulgarini, line 1: This context can include all the different API calls

src/hooks/useMessagesLane.ts
```diff
@@ -33,77 +33,84 @@ interface Props {
...
+        .then((unsub) => {
+          unsubscribeInboundLanes = unsub;
+        })
+        .catch((message) => {
+          dispatchMessage(MessageActionsCreators.triggerErrorMessage({ message }));
+          logger.error(message);
+        });
+    }
 
-    return async (): Promise<void> => {
```
hbulgarini, line 90: This is not right , the return function should not be awaiting to resolve an async function. So i removed the async and call it as it was before.

src/hooks/useMessagesLane.ts
```diff
@@ -33,77 +33,84 @@ interface Props {
...
+            pendingMessages: pendingMessages.isNeg() ? '0' : pendingMessages.toString(),
+            totalMessages: latest_generated_nonce
           });
-          return Promise.resolve(u);
-        }
-        const u = await api.query[bridgedMessages].inboundLanes(laneId, (res: any) => {
-          setter({ bridgeReceivedMessages: res.get('last_confirmed_nonce').toString() });
+        })
+        .then((unsub) => {
+          unsubscribeOutboundLanes = unsub;
```
hbulgarini, line 77: This is not preventing the leak issue that if the promise is not resolved by the time component is being unmounted, the unsub function will not run (anyway returning an async cleanup is not the right way either). Doing some research i found that the correct way to prevent this 100% is by using [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController).  Here it is a good explanation example about the usage: [link](https://dev.to/bil/using-abortcontroller-with-react-hooks-and-typescript-to-cancel-window-fetch-requests-1md4)

The limitation i see is that to be able to use the `abortController.abort()`, polkadotjs api need to support this by accepting a sign parameter so i can control the call with `abortController.signal`. Reading the api source code,  the api does not support this.

 Also the `useMountedState` prevents the issue of not updating the state if the component is not mounted but it is not preventing to execute the call if the component is already unmounted. So i have added back the `isMounted()` to prevent this. Actually polkadotjs UI is doing something similar: [useInactives](https://github.com/polkadot-js/apps/blob/master/packages/page-staking/src/Actions/useInactives.ts#L91)

@tomusdrw @wirednkod  i would like to have your feedback about this, sorry that i did not catch this during previous PR , but i realized about this when i was writing the tests.

src/contexts/OnChainValuesContextProvider.tsx
hbulgarini: also i'm opened to listen suggestions for a better context provider name ... :)

hbulgarini: forget about these two comments. 

1) i think calling this context Subscriptions is quite descriptive.
2) I changed my mind regarding this context usage ... i think it should be only charge of subscription so i just force the re render of components only when the subscriptions are updated.

src/types/subscriptionsTypes.ts
```diff
@@ -0,0 +1,36 @@
...
+  SOURCE = 'sourceSubscriptions',
+  TARGET = 'targetSubscriptions'
+}
+export interface Subscriptions {
+  bestBlock: string;
+  bestBlockFinalized: string;
+  bestBridgedFinalizedBlock: string;
+  bridgeReceivedMessages: string;
+  outboundLanes: {
+    latestReceivedNonce: string;
```
wirednkod, line 27: create a type called OutboundLanes containing the Object format:
```
OutboundLanes: {
latestReceivedNonce: string;
pendingMessages: string;
totalMessages: string;
}```
Then assign the type to outboundLanes:
```
outboundLanes: OutboundLanes;
```

**Review by wirednkod — CHANGES_REQUESTED** [review]

File: src/hooks/useSubscriptions.ts [hunk 8]
```diff
@@ -15,19 +15,32 @@
...
-  return { ...blockInfo, ...bridgedBlocks, ...messagesLane, local, polkadotjsUrl };
+const useSubscriptions = (ChainDetail: ChainDetails): Output => {
+  const {
+    apiConnection: { api, isApiReady },
+    target,
+    source,
+    polkadotjsUrl
+  } = useChainProfile(ChainDetail);
+
+  const blockInfo = useBlocksInfo({ api, chain: source, isApiReady });
```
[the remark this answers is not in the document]
hbulgarini: I have never agree that this use is incorrect. A hook inside a hook is composability, so this will not be changed.

src/hooks/useMessagesLane.ts
hbulgarini: >  I though that there should be no API calls inside hooks - but all to be on a root level and been controlled by external functions. Did that change?

That's the idea, but that's a bigger refactor that i didn't want to include in this PR since it was going to make it even more bigger ( it's quite big )

hbulgarini: Also , i have opened this issue in polkadotjs api gh: https://github.com/polkadot-js/api/issues/3597

src/hooks/useSubscriptions.ts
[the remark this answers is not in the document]
wirednkod: As we have discussed before, my opinion is that it is a matter of simplicity - as soon as useMessagesLane, useBridgedBlocks, useBlocksInfo are not used anywhere else, I dont understand why they need to be hooks. In any case, as soon as you decided that this will not be changed - my point is moot.

src/hooks/useMountedState.ts
```diff
@@ -37,6 +38,8 @@ export const useMountedState = <D>(initialState: D | (() => D)) => {
     (value: D) => {
       if (isMounted()) {
         setState(value);
+      } else {
+        logger.warn('Warning: Trying to update the state of unmounted component.');
```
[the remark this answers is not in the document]
hbulgarini: @tomusdrw  requested this in: https://github.com/paritytech/parity-bridges-ui/issues/141

src/hooks/useBlocksInfo.ts
```diff
@@ -26,41 +27,37 @@ interface Props {
...
     let unsubscribeBestNumberFinalized: () => void;
-    api.derive.chain
-      .bestNumberFinalized((res) => {
-        setBestBlockFinalized(res.toString());
-      })
-      .then((unsub) => {
-        unsubscribeBestNumberFinalized = unsub;
-      })
-      .catch((e) => logger.error('error reading blocks', e));
+    if (api && isApiReady && chain && isMounted()) {
```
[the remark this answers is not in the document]
hbulgarini: you are right on both . With the changes to handle the UI when is disconnected and the last unsub pattern those are not needed anymore.

src/hooks/useMountedState.ts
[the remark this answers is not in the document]
hbulgarini: whatever, i cannot apply this until we do a major refactor about state management.

src/hooks/tests/useBlocksInfo.test.ts
```diff
@@ -0,0 +1,85 @@
...
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+
+import { ApiPromise } from '@polkadot/api';
+import { renderHook } from '@testing-library/react-hooks';
+import useBlocksInfo from '../useBlocksInfo';
+import logger from '../../util/logger';
+
+jest.spyOn(logger, 'error');
+
+const chain = 'chain';
```
tomusdrw, line 24:
```suggestion
const CHAIN = 'chain';
```

IMHO this kind of constants should rather be uppercased to differentiate them from regular variables.

src/hooks/useBlocksInfo.ts
```diff
@@ -26,41 +27,36 @@ interface Props {
...
-      unsubscribeBestNumberFinalized && unsubscribeBestNumberFinalized();
-      unsubscribeBestNumber && unsubscribeBestNumber();
+      unsubscribeBestNumber
+        .then((u) => {
+          u();
+        })
+        .catch((e) => logger.error('error unsubscribing bestNumber', e));
+      unsubscribeBestNumberFinalized
+        .then((u) => u())
+        .catch((e) => logger.error('error unsubscribing bestNumberFinalized', e));
```
tomusdrw, line 57: I think with `async/await` it would look a bit nicer.

src/hooks/useBlocksInfo.ts
```diff
@@ -26,41 +27,36 @@ interface Props {
...
+    }
     return function cleanup() {
-      unsubscribeBestNumberFinalized && unsubscribeBestNumberFinalized();
-      unsubscribeBestNumber && unsubscribeBestNumber();
+      unsubscribeBestNumber
+        .then((u) => {
+          u();
+        })
+        .catch((e) => logger.error('error unsubscribing bestNumber', e));
+      unsubscribeBestNumberFinalized
```
tomusdrw, line 55: This may still fail, because if there is an exception in the first call, the second promise might not be initialized. Worth adding `?` at the end and changing the type accordingly - IMHO best to just initialize the variables with `null` initially, so that the compiler can tell you that they might not be set.

tomusdrw: Also we run the code conditionally (`if` statement), so it can be `null` if the conditions are not met.

src/hooks/useBlocksInfo.ts
```diff
@@ -26,41 +27,36 @@ interface Props {
...
-      .bestNumberFinalized((res) => {
-        setBestBlockFinalized(res.toString());
-      })
-      .then((unsub) => {
-        unsubscribeBestNumberFinalized = unsub;
-      })
-      .catch((e) => logger.error('error reading blocks', e));
+    let unsubscribeBestNumber: Promise<VoidFn>;
+    let unsubscribeBestNumberFinalized: Promise<VoidFn>;
+    if (isApiReady && chain) {
```
tomusdrw, line 36: I find it better to early-exit (like in previous code) instead of keep indenting/nesting the code with every condition.

src/hooks/useBridgedBlocks.ts
```diff
@@ -36,32 +37,53 @@ interface Props {
...
+    }
+
+    return () => {
+      unsubBestFinalized
+        .then((u) => {
+          u();
+        })
+        .catch((e) => logger.error('error unsubscribing unsubBestFinalized', e));
+    };
+  }, [isApiReady, chain, api, bridgedGrandpaChain, setBestFinalizedBlock]);
```
tomusdrw, line 62:
```suggestion
    const shouldProceed: boolean = !!(isApiReady && chain);
    if (!shouldProceed) {
      return;
    }
    
    const unsubBestFinalized = api.query[bridgedGrandpaChain].bestFinalized((res: CodecHeaderId) => {
        const bestFinalized = res.toString();
        setBestFinalizedBlock(bestFinalized);
    });

    return async () => {
      try {
        (await unsubBestFinalized)();
      } catch (e) {
        logger.error('error unsubscribing unsubBestFinalized', e)
      }
     };
  }, [isApiReady, chain, api, bridgedGrandpaChain, setBestFinalizedBlock]);
```

This is imho a bit cleaner, but it's more a matter of preference. Alternatively please just change `then(u => u())` to this more concise form.

src/hooks/useBridgedBlocks.ts
```diff
@@ -36,32 +37,53 @@ interface Props {
...
+
   useEffect((): (() => void) | undefined => {
-    const shouldProceed: boolean = !!(api && isApiReady && chain);
-    if (!shouldProceed) return;
+    let unsubImportedHeaders: Promise<VoidFn>;
+    const shouldProceed: boolean = !!(api && isApiReady && chain && bestFinalizedBlock);
```
tomusdrw, line 66: Repeating the condition sucks, IMHO it was much cleaner previously in a single `useEffect.

src/hooks/useChainProfile.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  let source = sourceChain;
+  let target = targetChain;
+  if (chainDetail === ChainDetails.TARGET) {
+    source = targetChain;
+    target = sourceChain;
+    return { source, target, ...restTarget };
+  }
+
+  return { source, target, ...restSource };
```
tomusdrw, line 49:
```suggestion
  if (chainDetail === ChainDetails.TARGET) {
    return { source: targetChain, target: sourceChain, ...restTarget };
  }

  return { source: sourceChain, target: targetChain, ...restSource };
```

src/hooks/useMessagesLane.ts
tomusdrw: No, I'm pretty sure you are incorrect.
```js
return async () => {
  unsubscribeOutboundLanes && (await unsubscribeOutboundLanes)();
  unsubscribeInboundLanes && (await unsubscribeInboundLanes)();
};
```
is (almost) exactly the same as:
```js
return () => {
   unsubscribeOutboundLanes && unsubscribeOutboundLanes.then(u => u());
   unsubscribeInboundLanes && unsubscribeInboundLanes.then(u => u());
};
```
The only differences are the return value (in the first case it's going to be a `Promise<void>` in the second case it's going to be `void`) and the unsubscribe ordering (await is going to be sequential, the other parallel) essentialy it's going to do exactly the same things.

I'd vote for reverting all of the code changes here, cause the new code is imho much less readable and definitely less concise.

src/hooks/useSubscriptions.ts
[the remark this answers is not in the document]
tomusdrw: Composing hooks is definitely expected and desirable - we shouldn't be inlining huge amounts of code, especially if there is a re-use potential.

That said, there is a thin line between composing for improving readability and making the code more complex due to too many details being hidden.
In this particular case I think it looks good as-is.

src/types/subscriptionsTypes.ts
```diff
@@ -0,0 +1,38 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+
+export enum ChainSubscriptions {
+  SOURCE = 'sourceSubscriptions',
+  TARGET = 'targetSubscriptions'
+}
```
tomusdrw, line 20: I'm not a fan of the enum-object-choose-enum approach (it's next time we do it).
The way I see this long term is that we simply have:
```
const chainRelatedObjects = [a, b, c];
const chainIdx = 0;

chainRelatedObjects[chainIdx];
```

So in that case `SubscriptionsContextType` (and `SourceTargetState`) instead of being an object should be an array, and there should be one function that just gives you `chainIdx` based on the chain name and it's simply used to retreive the correct subscription context or api instance, etc.

That's something that avoid code duplication long term.

src/util/chainsUtils.ts
```diff
@@ -24,14 +24,14 @@ interface Input {
...
+    sourceChainDetails: { chain: currentSourceChain }
     // eslint-disable-next-line react-hooks/rules-of-hooks
   } = useSourceTarget();
 
   const sourceChainsMatch = sourceChain === currentSourceChain;
 
-  const sourceRole = sourceChainsMatch ? ChainDetails.SOURCE : ChainDetails.TARGET;
-  const targetRole = sourceChainsMatch ? ChainDetails.TARGET : ChainDetails.SOURCE;
+  const sourceRole = sourceChainsMatch ? ChainSubscriptions.SOURCE : ChainSubscriptions.TARGET;
+  const targetRole = sourceChainsMatch ? ChainSubscriptions.TARGET : ChainSubscriptions.SOURCE;
```
tomusdrw, line 34: That's a super confusing change. We have `getSourceTargetRole`, but instead we return `ChainSubscriptions`? So how to get `ChainDetails` now?
Why is fucntion called `sourceTargetRole` returning anything that deals with `subscriptions` anyway?
I'm surprised this doesn't break any existing code and the compiler is not able to catch this.

tomusdrw (changes requested):

src/hooks/useBlocksInfo.ts
hbulgarini: fair point ... i will create a separate effect for each so each effect is in charge on one concern only.

src/types/subscriptionsTypes.ts
hbulgarini: hmmmm , interesting .. but how you would identify which connection plays the source and target role?  where would you put that logic?

src/util/chainsUtils.ts
hbulgarini: Honestly i don't see how this would break if i took care of all places where this is used and updated accordingly: For example:  https://github.com/paritytech/parity-bridges-ui/pull/179/files#diff-8762bd5279e5304ba0047479b3927e524649384b2270f3023437979a5df37ed1R52

Now i understand your point about the change might sound confusing but in the end is returning the chain role  related to the source/target subscriptions. In case you find a better naming, please let me know and i will update them since i'm not convinced either.

Anyway, i think this util fly away after this: https://github.com/paritytech/parity-bridges-ui/issues/159

src/hooks/useBlocksInfo.ts
hbulgarini: let's keep the discussion here: https://github.com/paritytech/parity-bridges-ui/pull/179/files#r641805104

src/hooks/useBridgedBlocks.ts
hbulgarini: https://github.com/paritytech/parity-bridges-ui/pull/179/files#r641805104

src/hooks/useBlocksInfo.ts
tomusdrw: I'm not a fan of blanket-applying some rules. Of course it makes sense to separate concerns in different `useEffects`, but we should also avoid code repetition - if we do a lot of related stuff and we would need to pass the same dependencies to multiple `useEffects` or copy paste some code - it's simply boilerplate.

IMHO less amount of code is simply less bugs and more readability. This particular case is imho somewhere at the border, I would probably accept both - two separate hooks or a single one, but I feel refactoring this as part of that PR is just introducing noise - like I don't see anything wrong with the previous code.

src/hooks/useBridgedBlocks.ts
tomusdrw: It's not only about `async/await`.

src/hooks/useBridgedBlocks.ts
tomusdrw: > What about if i extract the shouldProceed outside useEffects and i make the useEffect to depend on it? should that be a solution?

Yes, that's a much better solution:
1. No code duplication
2. Less dependencies to `useEffect`.

I still think that one `useEffect` would be easier to understand here. These two effecst are not independent - the second one depends on the first one and these kind of dependencides are hard to understand if `useEffects` are separate - you basically need to go through the entire flow of "if new block is updated, then this other things gets fired, which updates the `importedHeaders`".

src/hooks/useMessagesLane.ts
tomusdrw: You are missing the point here. Of course `async/await` in `useEffect` is causing us troubles - that's the exact reason why we have this discussion, we are thinking about abstracting this out and we have leaks due to subscriptions or setState-after-unmount. But using `then/catch` instead of `async/await` is not solving anything - it's just different syntax for the same stuff.

With current approach we **must** do asynchronous clean up and my point is that there is no difference between having a function and chaining promises or writing `async/await` function - the effect is exactly the same we wait for the promise to resolve and then we unsubscribe.

src/types/subscriptionsTypes.ts
tomusdrw: Well the selected account is associated with some chain name or chain index. And:
```js
const { selectedAccount } = useSelectedAccount();
const connections = useConnections();
const sourceChain = connections[selectedAccount.chainIndex];
// or even
/// const sourceChain = selectedAccount.chain
```

src/util/chainsUtils.ts
tomusdrw: > i took care of all places where this is used and updated according

Yes, then the name should signify it is dealing with subscriptions and `sourceTargetRole`. Like I honestly find it really hard to understand the reasoning behind returning roles and then extracting the correct connection/subscription based on the role.

Why can't we have a hook that returns the connection/subscriptions directly?

I agree that this is a bit OT, let's just change the name and refactor this during re-architecturing.

src/hooks/useMessagesLane.ts
hbulgarini: It is not exactly the same because of the type of the function that you are returning. With async await, you are returning a promise type instead of void. And i think that's not right ... please check the `EffectCallback` type of react code and read the comment:

```
    // NOTE: callbacks are _only_ allowed to return either void, or a destructor.
    // The destructor is itself only allowed to return void.
    type EffectCallback = () => (void | (() => void | undefined));
```

With `async/await` the function is returning a Promise type and the comment clearly say that must return void.

src/types/subscriptionsTypes.ts
hbulgarini: i like the approach , i will take it into account.

src/util/chainsUtils.ts
hbulgarini: ok, any recommendations for renaming?

src/hooks/useBlocksInfo.ts
hbulgarini: I think this change is necessary because it keeps consistency across other subscription hooks. But the try and catch issue must be solved, i think that adding another useEffect also helps to make the code easier to read.

src/hooks/useBridgedBlocks.ts
hbulgarini: please check now.

src/util/chainsUtils.ts
tomusdrw: I'm not sold on the entire "role" thing - like this function literally returns a field name from subscriptions object. Then maybe let's call it `getChainSubscriptionsKey`?

src/hooks/tests/useBridgedBlocks.test.ts
```diff
@@ -0,0 +1,119 @@
...
+        expect(unsubBestFinalized).toHaveBeenCalled();
+      });
+    });
+
+    it('should NOT call the query api.query.chain2.bestFinalized because the api is not ready', () => {
+      props.isApiReady = false;
+      renderHook(() => useBridgedBlocks(props));
+      expect(bestFinalizedMock).not.toHaveBeenCalled();
+    });
+  }); */
```
tomusdrw, line 84: Uncomment or remove.

src/hooks/useBlocksInfo.ts
```diff
@@ -26,41 +27,56 @@ interface Props {
...
+  const [bestBlock, setBestBlock] = useMountedState('');
+  const [bestBlockFinalized, setBestBlockFinalized] = useMountedState('');
+
+  const isReady = !!(isApiReady && chain);
 
   useEffect(() => {
-    if (!api || !isApiReady || !chain) {
-      setBestBlock('');
-      setBestBlockFinalized('');
+    let unsubscribeBestNumber: Promise<VoidFn>;
```
tomusdrw, line 36:
```suggestion
    if (!isReady) {
      return;
    }

    try {
      const unsubscribeBestNumber = api.derive.chain.bestNumber((res) => {
        setBestBlock(res.toString());
      });
      return () => {
         unsubscribeBestNumber
           .then((u) => u())
           .catch((e) => logger.error('error unsubscribing bestNumber', e));
      };
    } catch (e) {
      logger.error('error reading blocks', e);
    }
```

src/hooks/useBlocksInfo.ts
```diff
@@ -26,41 +27,56 @@ interface Props {
...
+  }, [chain, setBestBlock, isReady, api.derive.chain]);
 
-    let unsubscribeBestNumberFinalized: () => void;
-    api.derive.chain
-      .bestNumberFinalized((res) => {
+  useEffect(() => {
+    let unsubscribeBestNumberFinalized: Promise<VoidFn>;
+
+    if (!isReady) {
+      return;
```
tomusdrw, line 63: Please simplify as above. The `let `stuff is really not needed for anything.

src/hooks/useBridgedBlocks.ts
tomusdrw: No, it's still broken. if `shouldProceed === false you will call `.then` on `undefined` value in cleanup.

src/hooks/useMessagesLane.ts
```diff
@@ -33,77 +30,84 @@ interface Props {
...
-        logger.error(e.message);
-        return Promise.reject();
+
+  const { bridgedMessages } = getSubstrateDynamicNames(chain);
+
+  const isReady = !!(isApiReady && api.query[bridgedMessages] && chain);
+
+  useEffect(() => {
+    let unsubscribeOutboundLanes: Promise<VoidFn> | null = null;
+
```
tomusdrw, line 58: Please simplify the `let statement is completelty redundant if you early-exit.

src/hooks/useMessagesLane.ts
tomusdrw: Okay, I'm super confused now. We do use `TypeScript`, right? If the compiler does not complaint about incorrect types then something is really wrong here.

Yes, if the function suggests the return type should be `void` - that's fine. The only difference is that you need to introduce an `async` block if that makes the code simpler and more readable compared to nested `then-catch`.
Please find a helper function here to make sure the types are correct:
```js
function makeVoid(fn: () => Promise<void>): () => void {
  return () => {
     fn().catch(e => logger.error('e'));
  };
}
```

You can use this function everywhere like that:
```js
useEffect(makeVoid(async () => {
  /// whatever
}));
```
and and awesome consequence is that you don't even need `try-catch` block, cause it logs the error and if you don't see the code becoming cleaner if we do this, then the rest of the discussion is not worth it and we can just stay with `then-catch` if you prefer.

src/hooks/tests/useBridgedBlocks.test.ts
hbulgarini: damn ... my bad ... this should come back. thanks for catching

src/hooks/useMessagesLane.ts
hbulgarini: i like what you propose , but i had to adjust makeVoid a little to:

```
function makeVoid(fn: () => Promise<(() => void) | undefined>): () => void {
  return () => {
    fn().catch((e) => logger.error(e));
  };
}
```
In any case i receive the following eslint warning:

```
Accepts a function that contains imperative, possibly effectful code.

@param effect — Imperative function that can return a cleanup function

@param deps — If present, effect will only activate if the values in the list change.

@version — 16.8.0

@see — https ://reactjs.org/docs/hooks-reference.html#useeffect

React Hook useEffect received a function whose dependencies are unknown. Pass an inline function instead.eslintreact-hooks/exhaustive-deps
```

hbulgarini: i like what you propose , but i had to adjust makeVoid a little to:

```
function makeVoid(fn: () => Promise<(() => void) | undefined>): () => void {
  return () => {
    fn().catch((e) => logger.error(e));
  };
}
```

In any case i receive the following eslint warning:

```
Accepts a function that contains imperative, possibly effectful code.

@param effect — Imperative function that can return a cleanup function

@param deps — If present, effect will only activate if the values in the list change.

@version — 16.8.0

@see — https ://reactjs.org/docs/hooks-reference.html#useeffect

React Hook useEffect received a function whose dependencies are unknown. Pass an inline function instead.eslintreact-hooks/exhaustive-deps
```

hbulgarini: Please check this proposal: https://github.com/paritytech/parity-bridges-ui/pull/181

src/util/chainsUtils.ts
hbulgarini: deal.

src/hooks/tests/useBridgedBlocks.test.ts
tomusdrw: still here

src/hooks/useBlocksInfo.ts
tomusdrw: still not addressed even with a comment

src/hooks/useBlocksInfo.ts
tomusdrw: ditto

src/hooks/useBridgedBlocks.ts
tomusdrw: not addressed

tomusdrw (approved):
There is still a few issues that are not addressed. The commented code could be easily found if you self-reviewed the PR first. I don't want to block any further work though, so please proceed.

src/hooks/useBridgedBlocks.ts
hbulgarini: i think this discussion got outdated with PR #181

src/hooks/tests/useBridgedBlocks.test.ts
hbulgarini: This comment got outdated after PR #181

src/hooks/useBridgedBlocks.ts
hbulgarini: outdated.

src/hooks/useBlocksInfo.ts
hbulgarini: Comment got outdated after merging #181

src/contexts/OnChainValuesContextProvider.tsx
hbulgarini: > The commented code could be easily found if you self-reviewed the PR first. I don't want to block any further work though, so please proceed.

This issue was solved on PR #181  which overwrite the outdated testing code for `useBlockInfo.test.ts` as i basically had to rewrote it because it has now a complete different behavior.

hbulgarini:
> There is still a few issues that are not addressed. The commented code could be easily found if you self-reviewed the PR first. I don't want to block any further work though, so please proceed.

This issue was solved on PR #181 which overwrite the outdated testing code for `useBlockInfo.test.ts` as i basically had to rewrote it because it has now a complete different behavior.

src/hooks/useApiSubscription.ts
```diff
@@ -0,0 +1,42 @@
...
+export function useApiSubscription(fn: () => UnsubscribePromise, isReady: boolean): void {
+  useEffect(() => {
+    if (!isReady) {
+      return;
+    }
+
+    try {
+      const unsub = fn();
+      return () => {
+        isReady &&
```
tomusdrw, line 30:
```suggestion
```
just re-read the code and this is not necessary - it will always be `true`.

src/hooks/useBalance.ts
```diff
@@ -41,13 +39,13 @@ const initValues = {
 };
 
 const useBalance = (api: ApiPromise, address: string, providedSi: boolean = false): State => {
-  const { dispatchMessage } = useUpdateMessageContext();
   const [state, setState] = useMountedState<State>(initValues);
 
   useEffect((): (() => void) => {
-    const getBalance = async (api: ApiPromise, address: string, setState: any): Promise<VoidFn> => {
+    let unsubscribe: Promise<VoidFn>;
```
tomusdrw, line 45: Can we use `useApiSubscription` here?

tomusdrw (approved):

src/hooks/useBalance.ts
hbulgarini: of course ... let me open a new PR only for this so we don't get keep making this PR bigger.

PR #180, opened 2021-05-26
Set up project for Puppeteer and 1st happy path test

wirednkod:
Setup:
- in browser automation E2E test with puppeteer.
- single React component test (without caring about children) with puppeteer
- Create E2E happy path test for submitting a transaction.

Make sure that in your `.env.local` you have adjusted:
```
REACT_APP_KEYRING_DEV_LOAD_ACCOUNTS=true
```
This is for making the app - not to require the extension and have the dev accounts active (for dev/test purposes)

In 2 terminals, run on first one the UI (`yarn start`), and  on the second one run `yarn run test:e2e-alone`. You will see the automated test for happy path of submitting a transaction to execute.


.env
```diff
@@ -5,4 +5,4 @@ REACT_APP_CHAIN_2_CUSTOM_TYPES_URL=https://raw.githubusercontent.com/paritytech/
 REACT_APP_CHAIN_2_SUBSTRATE_PROVIDER=wss://wss.millau.brucke.link
 
 REACT_APP_LANE_ID=0x00000000
-REACT_APP_KEYRING_DEV_LOAD_ACCOUNTS=false
\ No newline at end of file
+REACT_APP_KEYRING_DEV_LOAD_ACCOUNTS=false
```
hbulgarini, line 8: please remove this line.

wirednkod: I did not add this line - why would I remove it?

wirednkod: If I do the app want start and an error will appear:
```
Error: Missing REACT_APP_KEYRING_DEV_LOAD_ACCOUNTS variable
```

e2e/app.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+    await expect(page).toMatch('google');
+  });
+});
+
+describe('<App />', () => {
+  beforeAll(async () => {
+    await page.goto(SERVER_URL, { waitUntil: 'domcontentloaded' });
+  }, JEST_TIMEOUT);
+
+  //   it(
```
hbulgarini, line 33: i would remove these comments.

e2e/global-setup.js
```diff
@@ -0,0 +1,22 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+const { setup: setupPuppeteer } = require('jest-environment-puppeteer')
+
+module.exports = async function globalSetup(globalConfig) {
+  await setupPuppeteer(globalConfig);
+  console.log('log setup()');
```
hbulgarini, line 20: is this console.log needed?

e2e/global-teardown.js
```diff
@@ -0,0 +1,22 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+const { teardown: teardownPuppeteer } = require('jest-environment-puppeteer');
+
+module.exports = async function globalTeardown(globalConfig) {
+  // Your global teardown
+  console.log('log teardown()');
```
hbulgarini, line 20: same: is this console.log needed?

e2e/puppeteer_standalone.js
```diff
@@ -0,0 +1,71 @@
...
+const SERVER_URL = 'http://localhost:3001';
+
+const chromeOptions = {
+  executablePath:process.env.chrome,
+  headless: false
+};
+
+(async function main() {
+  try {
+    // const pathToExtension = require('path').join(__dirname);
```
hbulgarini, line 26: Please remove this comments.

hbulgarini (approved):
LGTM, i have tested it and is doing the job nicely. I have just left minor comments regarding code comments which i believe that should be removed.

e2e/app.test.tsx
wirednkod: That was note due to "pending" things to be done (if we want this solution and continue testing with it)

e2e/puppeteer_standalone.js
wirednkod: That was note due to "pending" things to be done (if we want this solution and continue testing with it)

e2e/global-teardown.js
wirednkod: That was note due to "pending" things to be done (if we want this solution and continue testing with it)

e2e/global-setup.js
wirednkod: That was note due to "pending" things to be done (if we want this solution and continue testing with it)

e2e/app.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+// @ts-ignore due to isolatedModules flag - no import so this needed
+describe('Google', () => {
+  beforeAll(async () => {
+    await page.goto('https://google.com', { waitUntil: 'domcontentloaded' });
+  });
+
+  it('sanity check, test Google server by checking "google" text on page', async () => {
+    await expect(page).toMatch('google');
+  });
+});
```
tomusdrw, line 26: I guess we don't need this? :)

e2e/puppeteer_standalone.js
```diff
@@ -0,0 +1,69 @@
...
+//
+// Parity Bridges UI is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+const puppeteer = require('puppeteer');
+const SERVER_URL = 'http://localhost:3001';
```
tomusdrw, line 17: We can read this from `e2e/jest.config.js` to avoid duplication?

e2e/puppeteer_standalone.js
```diff
@@ -0,0 +1,69 @@
...
+};
+
+(async function main() {
+  try {
+    const browser = await puppeteer.launch(chromeOptions);
+    const page = await browser.newPage();
+    page.emulate({
+      viewport: {
+        width: 1280,
+        height: 1024
```
tomusdrw, line 31:
```suggestion
        width: 1920,
        height: 1080
```
It's not 2004 anymore :)

e2e/puppeteer_standalone.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+    await page.waitForTimeout(500);
+    await page.focus('#test_amount_send').then(() => console.log('-- Focus on amount input field.'));
+    await page.keyboard.type('10').then(() => console.log('-- add 10 MLAU'));;
+    await page.waitForTimeout(500);
+    await page.focus('#receiver_input').then(() => console.log('-- Focus on receiver input field'));
+    await page.keyboard.type('74YBVK9EJ4uSzokqewqmBcecAWq6tosB2xujX5Lf8jKg2gze').then(() => console.log('-- Add a receiver address'));
+    await page.waitForTimeout(500);
+    await page.click('#test-button-submit').then(() => console.log('-- Click send button to initiate the transaction'));
+    await page.waitForXPath('//div[contains(text(), "Transaction was broadcasted")]').then(() => console.log('-- Transaction was broadcasted.'));
```
tomusdrw, line 55: Isn't this kind of selectors a bit too generic?

e2e/puppeteer_standalone.js
```diff
@@ -0,0 +1,69 @@
...
+      userAgent: ''
+    });
+    await page.goto(SERVER_URL, { waitUntil: 'domcontentloaded' });
+
+    // Run happy path test for submitting a transaction
+    await page.waitForSelector('#test_sender_component');
+    await page.waitForTimeout(1000);
+    await page.click('#test_sender_component').then(() => console.log('-- Open Sender dropdown'));
+    await page.waitForTimeout(1000);
+    const [aliceOption] = await page.$x("//div[contains(., '5sauUX')]");
```
tomusdrw, line 42: I feel this kind of selector is a bit too generic, like it can pretty much match any div, no?

src/components/ExtensionAccountCheck.tsx
```diff
@@ -29,19 +29,21 @@ const ExtensionAccountCheck = ({ component }: Props): JSX.Element => {
   const { extensionExists, accountExists } = useKeyringContext();
   // TODO #176: Move this to a more generic error-show component
   const { sourceReady, targetReady } = useLoadingApi();
+  const isDevelopment = process.env.REACT_APP_KEYRING_DEV_LOAD_ACCOUNTS === 'true';
```
tomusdrw, line 32: Let's make a separate flag for this? Now it will stop displaying the extension prompt on our current deployment, since we do load the dev accounts.

src/components/ExtensionAccountCheck.tsx
```diff
@@ -29,19 +29,21 @@ const ExtensionAccountCheck = ({ component }: Props): JSX.Element => {
...
+    if (!extensionExists) {
+      msg = 'Connect to a wallet. Install polkadotjs extension';
+    } else if (!accountExists) {
+      msg = 'There are no accounts in the extension. Please create one';
+      // TODO #176: Move this to a more generic error-show component
+    } else if (!sourceReady || !targetReady) {
+      msg = `${!sourceReady ? statusFunc('Source', sourceReady) : ''} ${
+        !targetReady ? statusFunc('Target', targetReady) : ''
+      }`;
+    }
```
tomusdrw, line 45:
```suggestion
  if (isDevelopment) {
    return component;
  }
  
  if (!extensionExists) {
    msg = 'Connect to a wallet. Install polkadotjs extension';
  } else if (!accountExists) {
    msg = 'There are no accounts in the extension. Please create one';
    // TODO #176: Move this to a more generic error-show component
  } else if (!sourceReady || !targetReady) {
    msg = `${!sourceReady ? statusFunc('Source', sourceReady) : ''} ${
      !targetReady ? statusFunc('Target', targetReady) : ''
     }`;
  }
```

src/components/ReceiverInput.tsx
```diff
@@ -138,6 +138,7 @@ function ReceiverInput({ setError }: Props) {
       <Box display="flex" alignItems="center">
         <AccountIdenticon address={addressInput} formatFound={formatFound} />
         <InputBase
+          id="receiver_input"
```
tomusdrw, line 141: Let's use `_` or `-` and `test-` prefix consistently.
```suggestion
          id="test-receiver-input"
```

src/components/Sender.tsx
```diff
@@ -136,6 +136,7 @@ const Sender = () => {
   return (
     <>
       <Select
+        id="test_sender_component"
```
tomusdrw, line 139:
```suggestion
        id="test-sender-component"
```

src/components/Transfer.tsx
```diff
@@ -91,6 +91,7 @@ function Transfer() {
     <>
       <Box mb={2}>
         <TextField
+          id="test_amount_send"
```
tomusdrw, line 94:
```suggestion
          id="test-amount-send"
```

src/contexts/KeyringContextProvider.tsx
```diff
@@ -42,14 +42,13 @@ export function KeyringContextProvider(props: KeyringContextProviderProps): Reac
   const [extensionExists, setExtensionExists] = useState<boolean>(false);
   const [accountExists, setAccountExists] = useState<boolean>(false);
 
-  const isDevelopment = Boolean(process.env.REACT_APP_KEYRING_DEV_LOAD_ACCOUNTS);
-
+  const isDevelopment = process.env.REACT_APP_KEYRING_DEV_LOAD_ACCOUNTS === 'true';
```
tomusdrw, line 45: Same here, I'd prefer a separate flag for that.

tomusdrw (approved):
Looks good! CI integration will be the tough part, but I guess we can get some help from the CI team. Or perhaps there is a GH action for that already?

e2e/puppeteer_standalone.js
wirednkod: Can you elaborate please?

e2e/puppeteer_standalone.js
wirednkod: This is true. Fortunately it is expected (to what we have - or plan to have afaik) to have only 1 div with the address of Alice visible once the dropdown is clicked. Unfortunately, due to MaterialUI there is not much else that we can do. I have tried adding unique identifiers on the dropdown to identify the specific ones - or we could insert into the "contains" the exact address of Alice to match exact.

In any case this is still an MVP solution for E2E testing and it will and should be further improved.

e2e/puppeteer_standalone.js
tomusdrw: Same as with `alice` - like it can match anything on the page, even if it's invisible, hence I think it's a bit fragile.

We can maybe add a test-specific, invisible component which displays current transaction status and match on this? Obviously polluting production code with test code is not a very good practice, but from my experience it's much more practical than attempting to match exactly what the user sees - note the test here is meant to make sure that the transaction is being processed correctly, not necessarily that the right messages are being displayed.
The main problem in the past I had with E2E tests was that they are rarely reliable and people don't really know how to debug them, so minimizing flakiness is my priority here, hence proposing this separate test-specific component.

But we don't have to fix it now, just wanted to plant a seed. :)

e2e/puppeteer_standalone.js
tomusdrw: Can we restrict this `div` to some (Grand)parent component? Like `page.$x("//#sender-account div[contains...]")`? At least we will not look on the entire page but within the sender account component.

wirednkod: As we discussed - lets just let this for now until we finalize on how we are going to use the puppeteer (CI or... )

src/contexts/KeyringContextProvider.tsx
hbulgarini: @tomusdrw could you please elaborate about what a `separate flag for that` means?

hbulgarini: i think i get it ... can we use NODE_ENV in .env file?

tomusdrw: No, `NODE_ENV` is not the right thing to use here, we need a separate env variable that is specific to either E2E tests or just disabling extension requirement.

tomusdrw (approved):
LGTM! Could you please update `README.md` file to include:
- Instructions & requirements how to run e2e tests.
- Update on required environment variables.

PR #183, opened 2021-06-07
Adding sub directories to hook directories

hbulgarini:
This PR seems pretty big but is simple in a nutshell. The purpose of it is to create sub directories on `hook`directory. 
The criteria is split by concerns with the following list:

- hooks
     - accounts
     - chain
     - connections
     - react
     - subscriptions
     - transactions


tomusdrw (approved):

wirednkod (approved):
lgtm (though Im not sure why there is this reorganization needed)

Issue #176, opened 2021-05-18
Design error components


wirednkod:
We need to have specific error show component that will show all the different errors that can come for a process.

Description updated by @hbulgarini

goldsteinsveta:
There are three types of errors (and error components) that should be designed and implemented, related to:
- App and connection
- User input
- Submitted extrinsic processing (hard to detect, ideally should be prevented at user input stage)

goldsteinsveta:
@wirednkod, @hbulgarini, @tomusdrw I've added error [page in the design doc](https://www.figma.com/file/OqRXj7Th4uZhSWjNQn8CmH/?node-id=1395%3A30293). Screenshots from there:

App error:

<img width="640" alt="Screenshot 2021-07-20 at 17 50 13" src="https://user-images.githubusercontent.com/8782596/126356789-d0707384-2e2e-4493-987f-f3220a79fdb7.png">

Input error:

<img width="640" alt="Screenshot 2021-07-20 at 17 55 49" src="https://user-images.githubusercontent.com/8782596/126356838-e7e1885d-076c-4b9e-bad8-195e71392996.png">

Submitted extrinsic / subscription error:

<img width="640" alt="Screenshot 2021-07-20 at 17 54 02" src="https://user-images.githubusercontent.com/8782596/126356979-4e8c4fdd-1778-4a3d-8608-a94e5bc74838.png">

PR #186, opened 2021-06-11
Wrapping some api calls and moving estimateFee and get payload to transaction reducer.

hbulgarini:


src/hooks/transactions/useEstimateFee.ts
```diff
@@ -0,0 +1,68 @@
...
+export const useEstimateFee = () => {
+  const { createType, stateCall } = useApiCallsContext();
+  const laneId = useLaneId();
+  const {
+    sourceChainDetails: { chain: sourceChain },
+    targetChainDetails: { chain: targetChain }
+  } = useSourceTarget();
+  const { dispatchTransaction } = useUpdateTransactionContext();
+  const { estimatedFeeMethodName } = getSubstrateDynamicNames(targetChain);
+  const { payload } = useTransactionContext();
```
tomusdrw, line 34: I think it would be much cleaner to have all parameters passed as an argument to `getEstimateFeeCall`. Now on the component level you don't really know when to update `useEffect` and it will quickly get messy.

Basically IMHO this hook should only be responsible for calling the asynchronous method, but should not depend on any existing state.

src/hooks/transactions/usePayload.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export const usePayload = () => {
+  const { createType } = useApiCallsContext();
+  const {
+    sourceChainDetails: { chain: sourceChain }
+  } = useSourceTarget();
+  const { dispatchTransaction } = useUpdateTransactionContext();
+  const { account } = useAccountContext();
+
+  const getPayload = useCallback(
```
tomusdrw, line 34: This function should not be called `getPayload`, because it's not returning anything. It should be called `updatePayload`, but IMHO this should even be a hook at all - it would be much easier to make it a regular helper funciton that looks like this:
```
function getPayload(createType, chain, account) {
 ...
 return payload
}

/// Usage
const { createType } = useApiCallsContext();
const {
    sourceChainDetails: { chain: sourceChain }
 } = useSourceTarget();
try {
  const payload = getPayload(createType, sourceChain, account);
  dispatchTransaction(TransactionActionCreators.setPayload(payload));
} catch (e) {
  dispatchTransaction(TransactionActionCreators.setTransactionError(error, 'getPayload'));
}
```
The more we extract to regular non-React dependent functions the easier the code is to understand. You separate library logic from your domain-logic of the application.
IMHO this shouldn't really even be done by a component/hook - this is part of the data model really. Every time `call` changes the `payload` should be constructed right away.
Like these two things are tightly coupled, so we should be always updating `{call, payload}` together and never first `call` and then count on some hooks being present to auto-update the `payload`. It can either be done inside a reducer (but yeah, with current design you will run into issues of: "I need all of these parameters that are currently hooks") or simply on the call site that updates `call` in local/global state.

Some other issues I've noticed with the code:
- `createType` is something coming from polkadot.js/api - how does it correlate to `sourceChain`?
- It seems to me that this code would simply be broken if we use `createType` from `useApiCallsContext` and `targetChain`.
- This indicates some issue with API design - why these two things are separated? How do I get `createType` of the other chain?

src/hooks/transactions/useTransactionPreparation.ts
```diff
@@ -33,89 +29,31 @@ interface Props {
...
-
-      const estimatedFeeCall = await sourceApi.rpc.state.call<Codec>(estimatedFeeMethodName, messageFeeType.toHex());
-
-      // @ts-ignore
-      const estimatedFeeType = sourceApi.registry.createType('Option<Balance>', estimatedFeeCall);
-      const estimatedFee = estimatedFeeType.toString();
-
-      dispatchTransaction(TransactionActionCreators.setEstimateFee(estimatedFee));
+    const asyncCalculateFee = async () => {
+      await calculateFee();
```
tomusdrw, line 46: This is not supposed to be an async function. Also why does it calculate fee only once? It seems to me that passing all parameters to this function would make the entire flow much cleaner and easier to understand. We definitely want to re-calculate the fee each time `payload` changes.

src/reducers/transactionReducer.ts
```diff
@@ -46,6 +46,8 @@ export default function transactionReducer(state: TransactionState, action: Tran
   switch (action.type) {
     case TransactionActionTypes.SET_ESTIMATED_FEE:
       return { ...state, estimatedFee: action.payload.estimatedFee };
+    case TransactionActionTypes.SET_CALCULATING_FEE:
+      return { ...state, isCalculatingFee: action.payload.isCalculatingFee };
```
tomusdrw, line 50: This doesn't make much sense IMHO and that's why I'll keep arguing that reducers in their current form are just boilerplate code (we would achieve exactly the same with `useState` using much less code). 
IMO the whole point of reducer actions is not to expose single-value parameters in a KVDB form, but rather having actions that are specific to your entire data model and ensure that some parameters are updated in unison. Example:
```js
/// Actions
function startEstimateFee() {}
function finishEstimateFee(error: null | string, data: Fee | null) {}

/// Reducers
switch (action.type) {
   case TransactionActionTypes.START_ESTIMATE_FEE:
       return { 
         ...state,
         isCalculatingFee: true,
         estimatedFee: null,
         estimatedFeeError: null
       };
   case TransactionActionTypes.FINISH_ESTIMATE_FEE:
       return {
         ...state,
         isCalculatingFee: false,
         estimatedFee: action.payload.error ? null : action.payload.data,
         estimatedFeeError: action.payload.error,
       };
}
```
This way you can always be sure about some invariants about the set of parameters in your data model. And on the call site you have much less actions to dispatch - basically you separate the logic behind parameters from the place in the code that triggers these updates.

src/reducers/transactionReducer.ts
```diff
@@ -58,6 +60,10 @@ export default function transactionReducer(state: TransactionState, action: Tran
       return { ...state, derivedReceiverAccount: action.payload.derivedReceiverAccount };
     case TransactionActionTypes.SET_GENERIC_RECEIVER_ACCOUNT:
       return { ...state, genericReceiverAccount: action.payload.genericReceiverAccount };
+    case TransactionActionTypes.SET_PAYLOAD:
+      return { ...state, payload: action.payload.payload };
+    case TransactionActionTypes.SET_ERROR:
+      return { ...state, error: action.payload.error };
```
tomusdrw, line 66: This sounds super generic - what kind of error is this? This reducer is doing a lot of things and it's really not clear which part this error is related to. If it's generic to *all* possible parameters then I'd argue it's not really a good choice, cause we most likely want a bit more fine-grained error handling.

src/hooks/transactions/useEstimateFee.ts
```diff
@@ -0,0 +1,68 @@
...
+
+      // @ts-ignore
+      const estimatedFeeType = createType(sourceChain, 'Option<Balance>', estimatedFeeCall);
+      const estimatedFee = estimatedFeeType.toString();
+      dispatchTransaction(TransactionActionCreators.setEstimateFee(estimatedFee));
+    } catch (error) {
+      dispatchTransaction(TransactionActionCreators.setTransactionError(error, 'esstimateFee'));
+    } finally {
+      dispatchTransaction(TransactionActionCreators.setCalculatingFee(false));
+    }
```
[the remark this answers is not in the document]
hbulgarini: I really tried to make the `useApiGenericCall` hook to handle reducers dispatchers calls as you propose but in the end i end up with something like:

```
const setGlobalIsLoading = useCallback( (value)=> dispatchTransaction(Action.setIsLoading(value)), [dispatchTransaction] );

useApiGenericCall(call, {setGlobalIsLoading});
```

I see a couple of issues with this:

1) Adds more boilerplate since we have to wrap dispatchTransaction in useCallback. Without useCallback, the hook gets infinitive re render because of the different functions reference in each render. Also we would face this issue with using a local state of a context.  Using `dispatch` functions have a great benefit that is forgetting about the functions references in renders , so you avoid the useCallback.

2) the useApiGenericCall adds additional re renderings during the initial rendering because of it's initial state.

3) Makes useApiGenericCall quite messy imho.

I have shared more information here: https://gist.github.com/tomusdrw/6be7adb067ba44920931d8146f7467f7#gistcomment-3777056

src/reducers/transactionReducer.ts
hbulgarini: > This doesn't make much sense IMHO and that's why I'll keep arguing that reducers in their current form are just boilerplate code (we would achieve exactly the same with useState using much less code).

No. You would have to make sure that every call is wrapped in a useCallback. With dispatchers you don't face that issue.

> IMO the whole point of reducer actions is not to expose single-value parameters in a KVDB form, but rather having actions that are specific to your entire data model and ensure that some parameters are updated in unison.

Very good. Let me see how i can try to switch to this that makes a lot of sense.

src/hooks/chain/useChainGetters.ts
```diff
@@ -15,35 +15,64 @@
...
+            api: targetApi,
+            isApiReady: targetIsApiReady,
+            substrateValues: getSubstrateDynamicNames(sourceChain)
+          };
+        case GENERIC:
+          return {
+            ss58Format: GENERIC_SUBSTRATE_PREFIX,
+            api: {} as ApiPromise,
+            isApiReady: false,
+            substrateValues: {} as SubstrateDynamicNames
```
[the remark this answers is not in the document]
hbulgarini: you are right. 

> Perhaps best to split ss58Format to it's own function and leave the rest here?

Yes!

src/hooks/api/useApiCalls.ts
```diff
@@ -0,0 +1,51 @@
...
+      const { api } = getChainValues(chain);
+
+      const params = [methodName, data];
+      if (at) {
+        params.push(at);
+      }
+      // @ts-ignore
+      return api.rpc.state.call<Codec>(...params);
+    },
+    [getChainValues]
```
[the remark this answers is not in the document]
hbulgarini: right , the getValuesByChain is wrapped already in a useCallback.

src/hooks/chain/useChainGetters.ts
```diff
@@ -15,35 +15,64 @@
...
+          return {
+            ss58Format: targetSS58Format,
+            api: targetApi,
+            isApiReady: targetIsApiReady,
+            substrateValues: getSubstrateDynamicNames(sourceChain)
+          };
+        case GENERIC:
+          return {
+            ss58Format: GENERIC_SUBSTRATE_PREFIX,
+            api: {} as ApiPromise,
```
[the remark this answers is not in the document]
hbulgarini: This is gone , code removed .

src/hooks/transactions/useEstimateFee.ts
hbulgarini: https://github.com/paritytech/parity-bridges-ui/pull/186#discussion_r650903055

src/reducers/transactionReducer.ts
hbulgarini: I like more the name `reset` rather than `start`, because then i can `reset` the values on different part of application if needded.

src/reducers/transactionReducer.ts
hbulgarini: ok, i have followed your recommendation.

Then we need to work on a specific error component to display them all accordingly. 
The GH issue was already created: #176

src/components/ReceiverInput.tsx
```diff
@@ -82,7 +82,8 @@ function ReceiverInput({ setError }: Props) {
     dispatchTransaction(TransactionActionCreators.setGenericAccount(null));
```
hbulgarini, line 82: This group of dispatchers will fly away in PR: https://github.com/paritytech/parity-bridges-ui/pull/189/files#diff-c41ea0157ecc87ff655874eaec216c1c80b01f0546bcb6d66cd44565aeae6a0cR71

src/hooks/transactions/useEstimateFee.ts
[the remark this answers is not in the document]
hbulgarini: Regarding this file i opted for removing it from this PR since it was not used and it was created more as a boilerplate. 
Nikos in his draft PR #190 is actually using it so i prefer using his PR for including this function.

@wirednkod please be aware of this disscussion.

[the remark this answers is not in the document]
tomusdrw: > Adds more boilerplate

Wait, how removing the entire async code handling vs adding one `useCallback` is more boilerplate?

>  Without useCallback, the hook gets infinitive re render because of the different functions reference in each render.

Which hook? I don't see any `useEffect` here that could be triggered.

> Using dispatch functions have a great benefit that is forgetting about the functions references in renders , so you avoid the useCallback.

I'm not sure I understand this point. You need to pass `dispatch` to every `useEffect` anyway?

> the useApiGenericCall adds additional re renderings during the initial rendering because of it's initial state.

Well, the we can add initial state as a parameter? Also I don't think optimising re-renderings is a good metric currently. We are tackling more severe issues here like subscription leaks and incorrect async code handling (for instance using old values). If someone looses significant amount of money due to incorrect transaction being sent from the UI I'm pretty sure they would be convinced that it's because we wanted to avoid few re-renderings.

> Makes useApiGenericCall quite messy imho.

Yes, that's a fair point. We can consider having a separate hook-based async handler and dispatch-based async handler then? It's still an improvement because we only have 2 places in code which deal with async code instead of having it duplicated in random hooks/components.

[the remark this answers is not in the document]
hbulgarini: > Yes, that's a fair point. We can consider having a separate hook-based async handler and dispatch-based async handler then? It's still an improvement because we only have 2 places in code which deal with async code instead of having it duplicated in random hooks/components.

All your comments are valid , but i feel that it would be more effective to discuss on a code rather than explaining "if i do this, then happen this". I will try to create the equivalent useApiGenericCall for dispatchers then

src/hooks/transactions/useEstimateFee.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import { useCallback } from 'react';
+import { useApiCallsContext } from '../../contexts/ApiCallsContextProvider';
+import { useSourceTarget } from '../../contexts/SourceTargetContextProvider';
+import { TransactionActionCreators } from '../../actions/transactionActions';
+import { useUpdateTransactionContext, useTransactionContext } from '../../contexts/TransactionContext';
+import { getSubstrateDynamicNames } from '../../util/getSubstrateDynamicNames';
+import useLaneId from '../chain/useLaneId';
+
+export const useEstimateFee = () => {
```
wirednkod, line 25: I would alter the useEstimateFee to follow the pattern mentioned [here](https://gist.github.com/tomusdrw/6be7adb067ba44920931d8146f7467f7#gistcomment-3769751) and a real-code example [here](https://github.com/paritytech/parity-bridges-ui/pull/190/files#diff-7ad0d602f18b557f8ef933910d583aee088359750a74fa8e12e69e0095177e12).

To explain: This hook returns a function - It can be instead converted in 2 different functions. The `getEstimateFeeCall` and a hook that will include the `getEstimatedFeeCall` (of course you will need to bring back the useGetApi hook that was removed) (see examples above). 

This is a proposal for breaking this hook to smaller layers - removing the dispatchers from the function logic (`layers 2-3`) - but again this (as is) can be assumed as `Layer 4` (see [here](https://gist.github.com/tomusdrw/6be7adb067ba44920931d8146f7467f7#gistcomment-3777140) for layers)

wirednkod (commented):
Overall lgtm - some comments about things we already discussed on the 1 to 1 call. Nothing blocking though for merging. I would add *TODOs* for next step fixing

src/hooks/chain/useChainGetters.ts
```diff
@@ -14,34 +14,68 @@
...
       case GENERIC:
-        return { ss58Format: GENERIC_SUBSTRATE_PREFIX, api: {} as ApiPromise };
+        return GENERIC_SUBSTRATE_PREFIX;
+      case targetChain:
+        return targetSS58Format;
+      case sourceChain:
+        return sourceSS58Format;
       default:
-        throw new Error(`Unknown type: ${chain}`);
+        return GENERIC_SUBSTRATE_PREFIX;
```
tomusdrw, line 78: I think it's safer to throw an error here, cause AFAIU we should never hit this line really.

src/hooks/transactions/useEstimateFee.ts
```diff
@@ -0,0 +1,66 @@
...
+    targetChainDetails: { chain: targetChain }
+  } = useSourceTarget();
+  const { dispatchTransaction } = useUpdateTransactionContext();
+  const { estimatedFeeMethodName } = getSubstrateDynamicNames(targetChain);
+  const { payload } = useTransactionContext();
+
+  const getEstimateFeeCall = useCallback(async () => {
+    if (!payload) {
+      return;
+    }
```
tomusdrw, line 39: So seems that in case the payload ends up being empty we will still keep previously calculated fee in the state.

src/hooks/transactions/usePayload.ts
```diff
@@ -0,0 +1,68 @@
...
+          spec_version: 1,
+          weight
+        };
+        // @ts-ignore
+        const payloadType = createType(sourceChain, 'OutboundPayload', payload);
+        logger.info(`OutboundPayload: ${JSON.stringify(payload)}`);
+        logger.info(`OutboundPayload.toHex(): ${payloadType.toHex()}`);
+        dispatchTransaction(TransactionActionCreators.setPayload(payload, null));
+      } catch (error) {
+        dispatchTransaction(TransactionActionCreators.setPayload(null, error));
```
tomusdrw, line 59: The usual convention with callbacks is that the error comes as the first argument.

src/hooks/transactions/useTransactionPreparation.ts
tomusdrw: Yes, you are right. And that's the exact reason why the code is hard to grasp. It's confusing to see a function that takes no parameters at all, but keeps changing every time some other random parameters change. The only way to figure out what will trigger fee calculation is to follow the code path and that's just entering a rabbit hole.

Also again, why is this function `async`? It should be just a regular function, even if it was `async` there is no point in handling it here cause we don't even expect it to return anything nor we handle errors.

src/hooks/transactions/useTransactionPreparation.ts
```diff
@@ -33,89 +29,31 @@ interface Props {
...
-
-  const [payload, setPayload] = useState<null | {}>(null);
   const { call, weight } = useTransactionType({ input, type, weightInput });
 
-  const { dispatchTransaction } = useUpdateTransactionContext();
-  const { estimatedFeeMethodName } = getSubstrateDynamicNames(targetChain);
+  const calculateFee = useEstimateFee();
+  const updatePayload = usePayload();
 
   useEffect(() => {
```
tomusdrw, line 44:
```suggestion
  useEffect(() => {
    if (areApiReady) { calculateFee() }
  }, [areApiReady, calculateFee]);
```

src/hooks/transactions/useTransactionPreparation.ts
```diff
@@ -33,89 +29,31 @@ interface Props {
...
-    logger.info(`OutboundPayload: ${JSON.stringify(payload)}`);
-    logger.info(`OutboundPayload.toHex(): ${payloadType.toHex()}`);
-    setPayload(payload);
-  }, [account, call, isValidCall, type, weight, sourceApi.registry]);
-
-  return {
-    payload
-  };
+    updatePayload(call, weight);
+  }, [account, call, isValidCall, type, weight, sourceChain, updatePayload]);
```
tomusdrw, line 58: This is extremely confusing and another source of bugs. Why are all of these deps here? If they are really needed that's something seriously worrisome, but I hope they are just forgotten during refactoring - again if that's the case I really think this kind of things should be caught during self-review, which I kindly ask you to do.

```suggestion
  }, [call, isValidCall, weight, updatePayload]);
```

src/reducers/transactionReducer.ts
```diff
@@ -45,7 +45,29 @@ const createTransaction = (state: TransactionState, initialTransaction: Transact
...
+        ...state,
+        estimatedFee: action.payload.estimatedFeeError ? null : action.payload.estimatedFee,
+        estimatedFeeError: action.payload.estimatedFeeError
+      };
+    case TransactionActionTypes.RESET_ESTIMATED_FEE:
+      return {
+        ...state,
+        estimatedFee: null,
+        estimatedFeeError: null
+      };
```
tomusdrw, line 58: Couldn't this be done via `setEstimateFee(null, null)`? Like we can have a separate action creator for that but we don't need a separate action and reducer branch.

tomusdrw: That's said, it's imho already much, much better than independent parameters. We could keep this action (but renamed), because what I feel is missing is the `isEstimatedFeeLoading` indicator. Now you don't really know if the `fee` is `null` because it's supposed to be `null` (due to some other things missing) or if it's actually loading.

src/hooks/transactions/useTransactionPreparation.ts
hbulgarini: you are right on both ... also it is not needed to have an async call.

src/hooks/transactions/useEstimateFee.ts
hbulgarini: Right, this problem has gone with the implementation of `useDispatchGenericCall`

src/reducers/transactionReducer.ts
hbulgarini: Please check after this commit: https://github.com/paritytech/parity-bridges-ui/pull/186/commits/2fc673b15196be26e3810302514d37d4c7a1e046

hbulgarini: >That's said, it's imho already much, much better than independent parameters. We could keep this action (but renamed), because what I feel is missing is the isEstimatedFeeLoading indicator. Now you don't really know if the fee is null because it's supposed to be null (due to some other things missing) or if it's actually loading.

I'm  afraid that i'm not following you on this one. Anyway i have changed the `setEstimateFee(null, null)` as you recommended so maybe this comment got outdated already.

src/hooks/transactions/useEstimateFee.ts
hbulgarini: well in a way it is reusable because it is being used not only in Transfer component but in Remark and CustomCall.
Anyway , i have set the payload as input parameter as you recommended.

src/hooks/transactions/useTransactionPreparation.ts
hbulgarini: Code updated.

src/hooks/transactions/useTransactionPreparation.ts
hbulgarini: Code updated. Looks better now.

src/reducers/transactionReducer.ts
hbulgarini: In a nutshell , i have removed the loading status and when call starts is calling:  setEstimateFee(null, null)

src/hooks/transactions/useEstimateFee.ts
hbulgarini: hey ... please check the code now. Is not exactly the pattern you have shared , but is much similar.

src/hooks/transactions/useEstimateFee.ts
```diff
@@ -0,0 +1,73 @@
...
+  );
+
+  const dispatch = useCallback(
+    (error: string, data: any) => dispatchTransaction(TransactionActionCreators.setEstimatedFee(error, data)),
+    [dispatchTransaction]
+  );
+
+  const { execute: calculateEstimateFee } = useDispatchGenericCall({
+    call: estimateFeeCallback,
+    shouldExecute: Boolean(payload),
```
tomusdrw, line 66: This means that (given current implementation) if `payload` suddenly becomes falsy we will keep the previous estimated fee in the state. I think it would be more sensible to reset the fee too, but I'm not super strong on this.

src/hooks/transactions/usePayload.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  const dispatch = useCallback(
+    (error: string, data: any) => dispatchTransaction(TransactionActionCreators.setPayload(error, data)),
+    [dispatchTransaction]
+  );
+
+  const { execute } = useDispatchGenericCall({
+    call: payloadCallback,
+    dispatch,
+    shouldExecute: Boolean(account && call && weight)
```
tomusdrw, line 65: Same here - if any of these is empty, we will end up with the old payload in the state. I'm assuming it's not incorrect now, but afaict the way you are overcoming not displaying the old payload/fee in case they are not valid is to repeat exact same conditions elsewere.

IMHO having a clear data model, where either all data is complete and valid or there is no data at all is much easier to reason about. Simply if you want to display `payload` and it's not there, you don't display it instead of adding extra checks about `account/call/weight` etc.

src/hooks/transactions/useTransactionPreparation.ts
```diff
@@ -33,89 +27,21 @@ interface Props {
...
-    dispatchTransaction,
-    estimatedFeeMethodName,
-    laneId,
-    payload,
-    sourceApi.registry,
-    sourceApi.rpc.state,
-    targetChain
-  ]);
+    calculateFee(payload);
+  }, [calculateFee, payload]);
```
tomusdrw, line 39: Thanks! This make much more sense to me now and makes it easier to follow why and when things change.

src/reducers/transactionReducer.ts
tomusdrw: So you seem to assume that the `async` call we are making is instant to the user (i.e. below 200ms). This is true on local development, but is not true in general.

I think we should have a loading state flag for every asynchronously updated parameter, so that we can display this to the user.
Now you are setting the fee to `null`- that's cool, the user will see that it's not there, however on the data model level you can't distinguish these two situations:
1. The fee is not there, because the input is somehow invalid and it will never be updated.
2. The fee is not there, because it's being calculated - the user should stay still because it will be updated shortly.

I really recommend testing the UI after enabling network throttling in the Chrome Dev Tools. It's not only about people on slow connections, but long term it's also about light(-er) clients that might take a while to process complex queries (like fee estimation).

tomusdrw (approved):
lgtm! There are still few comments, but I trust you'll address them correctly.

src/reducers/transactionReducer.ts
hbulgarini: @tomusdrw  i would like to clarify this before merging anything. I'm feeling quite confused now about the loaders states because in previous comments it was requested to not use a separate loader variable in the reducer:

https://github.com/paritytech/parity-bridges-ui/pull/186#discussion_r650798957

https://github.com/paritytech/parity-bridges-ui/pull/186#discussion_r650801306

https://github.com/paritytech/parity-bridges-ui/pull/186#discussion_r654432343

My first approach was to use a dedicated variable for loading states as all other libraries / code do , but based on the discussions shared above i changed the code. If you asked me , my first and preferred approach was always to use dedicated loading variables for every call ( and in this case it will be in the reducer since it is a global value )

src/hooks/api/useDispatchGenericCall.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const useDispatchGenericCall = ({ call, dispatch, shouldExecute = true }: DispatchGenericCallInput) => {
+  const execute = useCallback(
+    async (...params: any[]) => {
+      let data;
+      let error;
+      try {
+        if (!shouldExecute) {
+          return null;
+        }
```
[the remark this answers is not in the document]
hbulgarini: The purpose of this check is to not trigger extra actions calls at the beginning when values are null. So without this we would be triggering unnecessary updates to the reducer with null values.

src/hooks/chain/useSendMessage.ts
```diff
@@ -53,100 +54,119 @@ function useSendMessage({ isRunning, isValidCall, setIsRunning, input, type, wei
...
+        const options: Partial<SignerOptions> = {
+          nonce: -1
+        };
+        let sourceAccount: string | KeyringPair = account;
+        if (account.meta.isInjected) {
+          const injector = await web3FromSource(account.meta.source as string);
+          options.signer = injector.signer;
+          sourceAccount = account.address;
+        }
+        const unsub = await bridgeMessage.signAndSend(sourceAccount, { ...options }, ({ events = [], status }) => {
```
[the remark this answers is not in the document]
hbulgarini: I would really move this to a reducer variable so everything gets more consistent ...

src/hooks/api/useDispatchGenericCall.ts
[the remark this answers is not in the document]
hbulgarini: Anyway, what you make me think now is that if there is an error in the payload and it gets null or something, then the payload would not be updated to `null` as it should ... i think that i prefer having additional re-renders at the beginning rather than not having the payload updated correctly if there is an error.

src/hooks/transactions/useEstimateFee.ts
hbulgarini: https://github.com/paritytech/parity-bridges-ui/pull/186#discussion_r656196486

src/reducers/transactionReducer.ts
hbulgarini: What i can do is to put back specific loading variables for each action but make the dispatch function to be in charge of updating for example: 

`const setEstimatedFee = (estimatedFeeError: string | null, estimatingFeeLoading: boolean, estimatedFee: string | null) => {
  return {
    payload: { estimatedFee, estimatedFeeError, estimatingFeeLoading },
    type: TransactionActionTypes.SET_ESTIMATED_FEE
  };
};`

then on the `useDispatchGenericCall` i could use:

`  const execute = useCallback(
    async (...params: any[]) => {
      let data;
      let error;
      try {
        dispatch(null, true, null);
        data = await call(...params);
        dispatch(null, false, data);

      } catch (e) {
        error = e;
        dispatch(e, false,null);
      }
      return { data, error };
    },
    [call, dispatch]
  );`

src/hooks/chain/useSendMessage.ts
```diff
@@ -53,100 +54,119 @@ function useSendMessage({ isRunning, isValidCall, setIsRunning, input, type, wei
...
-    } finally {
-      setIsRunning(false);
-    }
-  };
+  }, [account, isRunning, makeCall, setIsRunning]);
 
   const { areApiReady } = useLoadingApi();
 
-  const isButtonDisabled = () => {
+  const isButtonDisabled = useCallback(() => {
```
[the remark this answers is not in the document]
hbulgarini: you are 100% right

src/reducers/transactionReducer.ts
tomusdrw: I think you are confusing two things:
1. Can we display the fee - i.e. is there a meaningful value in `estimateFee` variable.
2. Is the fee currently being loaded - i.e. `isFeeLoading === true`.

In the linked comments you've been using `isFeeLoading`:
1. either together with `estimateFee` check, which is superfluous if we know that every time `isFeeLoading === true` then `estimateFee === null`.
2. or incorrectly instead of `estimateFee === null` check (i.e. `estimateFee` can be `null` not only when it's loading).

Basically you want to prevent something from happening if the fee is not there (be it either display, or kicking off some extra calculations). Note that the loading state is not really relevant, cause it's only one case when fee is `null`, maybe there was some other error (i.e. missing parameters, API not available, etc) that makes the fee unavailable even if it's not loading.

The `isFeeLoading` variable should be used to display loading indicator to the user in case we know that the fee is being refreshed, but it should be ONLY used for that purpose - it's a pure user experience related state variable, everywhere else you should be using `estimateFee === null` cause it covers more broad spectrum of cases (missing data, errors, etc).


Also note that I was against introducing a separate `action` to update `isFeeLoading`, but not about having this value at all. Basically because the value of this variable is tightly coupled with other fee-related state variables I was suggesting that we should have two actions:
1. We are starting to make an asynchronous request to calculate the fee. (`startEstimateFee` or sth like that)
2. The asynchronous request to calculate the fee has finished (either error, timeout, or a success) (`finishEstimateFee` or sth like that)

And the first action should be responsible for `null`ifying the values and setting `isFeeLoading` to `true`, while the second action should be responsbile for flipping `isFeeLoading` back to false and setting either error or successful result to the state.

tomusdrw: > What i can do is to put back specific loading variables for each action but make the dispatch function to be in charge of updating for example:

By adding `isFeeLoading` parameter to the action creator you are exposing the internal details of state management to the component level and risk state inconsistency (what if a component sets `data` to something not `null`, but sets `loading` to `true`?).

Making reducer actions more "domain specific" is the way to go - that's the main distinction between regular `useState`, where you basically have no central state management, but you expose all details to the component level - which is actually much better (less boilerplate) in case of simple state. Reducers really shine in complex state management, but in such case you also need to have domain specific actions, and not simply `setters` for values that the reducer state holds.

tomusdrw: But as I said earlier in some comment, I feel long-term consistency will be much more valueable than simply addressing this comments in one PR. Like if you fundamentally don't think that domain-specific actions are easier to reason about and you prefer the setter approach, let's do it consistently everywhere, I'm not going to die on this hill, but just telling you from my experience that state consistency and simplicity of interactions makes the code easier to reason about and debug.

src/hooks/chain/useSendMessage.ts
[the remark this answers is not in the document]
hbulgarini: by the way, it is working as expected with the current code. There is no need to add additional `setIsRunning` ...

At the beginning of the exported function `sendMessageLane` the first step is to set the `setIsRunning` to `true`.
Then the  async `makeCall` function contains the `finally` statement that set the running state to `false` again. 

As i mentioned before i would work on this on a separate PR.

PR #189, opened 2021-06-15
Moving receiver logic to reducer.

hbulgarini:
This PR depends on #186 and it is actually pointed to it. Once #186 gets merged this will automatically pointed to master branch.

Closes: #160 

- Moved set receiver logic living in the component ( with generic / companion ) to transaction reducer.
- Implementation of `reset()` action which lets to reset all the transaction state in one action and get rid of all the different resets.
- Simplified receiver input component into more presentational approach.
- Simplified `useReceiver` hook.


src/hooks/transactions/useReceiver.ts
```diff
@@ -14,43 +14,110 @@
...
+        setReceiver(receiver);
+        return;
+      }
+
+      dispatchTransaction(
+        TransactionActionCreators.setValidationError(`Unsupported address SS58 prefix: ${formatFound}`)
+      );
+    },
+    [dispatchTransaction, reset, setReceiver, setUnformattedReceiver, sourceChain, targetChain, validateAccount]
+  );
```
tomusdrw, line 114: It's a bit hard to tell what the specific goal of the PR currently is cause it's still a draft, so I believe there is more work expected to be done here, but IMHO this entire function should simply be part of the reducer.

I don't see a reason why we send multiple actions (`reset` and then something else) in pretty much one function frame - this to me sounds like a reducer anti-pattern (mentioned that in some other PR - the actions should be as course-grained and domain-specific as possible).

`onReceiverChange` callback should literally just send one action `setUnformattedReceiver` and the rest (validation, format detection logic, etc, should be part of the reducer). I guarantee that this will make the code much easier to understand.

As mentioned earlier, using reducers without placing domain logic there (and instead keeping it in components/hooks) is misusing reducers and just adds unnecessary boilerplate of dispatching transactions - instead top-level `useState` with values&setters passed through context would be much simpler.

Overall if you are not intending to move logic to reducers at all, then let's keep the code as-is, cause having some imho half-baked solutions (where some stuff happens in the reducer and some logic is in hooks) is going to complicate the code instead of improving it.

src/reducers/transactionReducer.ts
```diff
@@ -53,21 +53,34 @@ export default function transactionReducer(state: TransactionState, action: Tran
...
+    case TransactionActionTypes.RESET:
+      return {
+        ...state,
+        derivedReceiverAccount: null,
+        estimatedFee: null,
+        estimatedFeeError: null,
+        genericReceiverAccount: null,
+        receiverAddress: null,
+        unformattedReceiverAddress: null,
+        validationError: null,
```
tomusdrw, line 80: `validationError` in the context of `transactionReducer` suggests that this is some kind of transaction error. But afaict it is only related to receiver address.
IMHO this is really important, cause naming things properly makes the code much more future proof and easier to understand for future (both internal and external) developers.

src/actions/transactionActions.ts
```diff
@@ -18,17 +18,20 @@ import { TransactionStatusType, UpdatedTransactionStatusType } from '../types/tr
 
 enum TransactionActionTypes {
   RESET_ESTIMATED_FEE = 'RESET_ESTIMATED_FEE',
-  CLEAR_ESTIMATED_FEE = 'CLEAR_ESTIMATED_FEE',
   SET_RECEIVER_ADDRESS = 'SET_RECEIVER_ADDRESS',
+  SET_ESTIMATED_FEE = 'SET_ESTIMATED_FEE',
```
tomusdrw, line 22: Why do we have `set` and `reset` now? Is there a difference between doing `setEstimatedFee(null)` and `resetEstimatedFee`? Introducing this kind of ambiguity is the reason why the code is perceived as "difficult" (note that we already had a very confusing `RESET/CLEAR` ambiguity earlier.

src/actions/transactionActions.ts
```diff
@@ -18,17 +18,20 @@ import { TransactionStatusType, UpdatedTransactionStatusType } from '../types/tr
...
   UPDATE_CURRENT_TRANSACTION_STATUS = 'UPDATE_CURRENT_TRANSACTION_STATUS',
   SET_TRANSACTION_COMPLETED = 'SET_TRANSACTION_COMPLETED',
   SET_RECEIVER_VALIDATION = 'SET_RECEIVER_VALIDATION',
   SET_DERIVED_RECEIVER_ACCOUNT = 'SET_DERIVED_RECEIVER_ACCOUNT',
   SET_GENERIC_RECEIVER_ACCOUNT = 'SET_GENERIC_RECEIVER_ACCOUNT',
+  SET_PAYLOAD_ERROR = 'SET_PAYLOAD_ERROR',
+  SET_VALIDATION_ERROR = 'SET_VALIDATION_ERROR',
   SET_PAYLOAD = 'SET_PAYLOAD',
-  CLEAR_PAYLOAD = 'CLEAR_PAYLOAD'
+  RESET_PAYLOAD = 'RESET_PAYLOAD',
```
tomusdrw, line 33: Again, we have 3 methods that deal with the payload, and it's really not clear what they are supposed to do. Also the ordering of actions in this file seems a bit random. Why is `VALIDATION_ERROR` closer to `PAYLOAD` than `PAYLOAD_ERROR`?

tomusdrw (commented):
I'd like to ask you to please try to explain the context, goal/purpose and some reasoning in every PR description.

Right now this PR is marked as a draft, but I'm requested to review it and I have no idea which parts are finished or where is it heading.

hbulgarini:
> I'd like to ask you to please try to explain the context, goal/purpose and some reasoning in every PR description.
> 
> Right now this PR is marked as a draft, but I'm requested to review it and I have no idea which parts are finished or where is it heading.

Yes, my bad setting you as reviewer while PR is in draft. I thought notifications were not going to be sent to review.
I will contact you to let you know when you can review it.

src/reducers/transactionReducer.ts
hbulgarini: guilty ,  i'm not the best one naming variables :)

src/actions/transactionActions.ts
hbulgarini: i disagree on this for a couple of reasons:

1) i think that term `RESET` is quite descriptive about what the action is doing.. something that for example `setEstimatedFee(null)` is not so since you need to understand that estimatedFee is null hence is reseted.
2) there are some cases where the payload is an object with a couple of properties or a few parameters so is much cleaner to have a `resetSomething()` without any needed parameters than `setSomething(param1,param2,param3)`

src/actions/transactionActions.ts
hbulgarini: `VALIDATION_ERROR` is not related to payload ... is related to address validation. Anyway i accepted my fault and i renamed to `addressValidationError`. Also agree that order should be consistent across concerns.

src/actions/transactionActions.ts
hbulgarini: We don't anymore .  I have just followed your recommendation about doing: `setEstimatedFee(null,null)`

hbulgarini: Anyway i have created a new reset() action which is resetting all the corresponding states.

src/hooks/transactions/useReceiver.ts
hbulgarini: I think this comment is outdated now since the approach of the PR changed drastically.

tomusdrw (approved):

src/hooks/accounts/useGenericAccount.ts
```diff
@@ -0,0 +1,88 @@
...
+  }, [dispatchTransaction]);
+
+  const setAddressAsTarget = useCallback(
+    (address: string, addressKind: AddressKind) => {
+      if (selected) {
+        looseHelperAccount();
+        return;
+      }
+      setSelected(addressKind);
+      dispatchTransaction(TransactionActionCreators.setReceiverAddress(address));
```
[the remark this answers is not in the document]
hbulgarini: I see your point but on the other hand, extracting the `selected` logic to the hook makes the `GenericAccount` component fully presentational.

I will rename the function so we have agreement of both sides.

PR #190, opened 2021-06-15
Move apis to 'library'

wirednkod:
The idea of this PR is to showcase how the API calls can be moved to hooks/functions at "top" level of the application in a "library" of reusable code. 
The overall idea is to eventually split the app in layers (2 -> 5) and probably having layers 2->4 as a reusable "library" for more than 1 apps,  as mentioned [here](https://gist.github.com/tomusdrw/6be7adb067ba44920931d8146f7467f7#gistcomment-3777140) and can be seen below:

> **Layer 1: API Layer** -> PolkadotJS API
> **Layer 2: Bridges API specific Layer** -> Bridges API that uses specific needed things from polkadotJS API (not all API calls from polkadot JS are needed
> **Layer 3: Bridges Simple/Specific Business logic** -> Functions/hooks that are using the Bridges API for 1 VERY specific reusable action (get the chain, estimate fee etc etc)
> **Layer 4: Bridges Advanced/Complicated Logic** -> Functions/hooks that can use any of Layers 2, 3 to make complicated logic (send a transaction and estimate fee and do something else at the same time)
> **Layer 5: Bridges or other Bridges-related UI** -> UI components that can use any of Layers 2 - 4, to re-create or reuse Bridges functionalities (simple or complicated)


src/api/getBridgedBlocks.ts
```diff
@@ -0,0 +1,44 @@
...
+
+interface DataType {
+  api: ApiPromise;
+  apiMethod: string;
+  separator: string;
+  setter: any;
+  arg1?: any;
+}
+
+export const getBridgedBlocks = async ({ api, apiMethod, separator, setter, arg1 }: DataType): Promise<VoidFn> => {
```
tomusdrw, line 28: Why couldn't this be two separate methods with proper typing and argument names? `arg1` seems a bit like an overkill.

src/api/getBridgedBlocks.ts
```diff
@@ -0,0 +1,44 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+import { ApiPromise } from '@polkadot/api';
+import { VoidFn } from '@polkadot/api/types';
+import { useGetApi, ReturnApi } from './useGetApi';
+
+interface DataType {
+  api: ApiPromise;
+  apiMethod: string;
+  separator: string;
```
tomusdrw, line 23: If this is expected to be only two values maybe better do:
```suggestion
  separator: 'bestFinalized' | 'importedHeaders';
```
?

tomusdrw: Or even better let's just have two separate functions?

src/hooks/subscriptions/useMessagesLane.ts
```diff
@@ -41,33 +41,31 @@ const useMessagesLane = ({ isApiReady, api, chain }: SubscriptionInput): Output
...
-        const latest_generated_nonce = res.get('latest_generated_nonce').toString();
-        const latest_received_nonce = res.get('latest_received_nonce').toString();
-        const pendingMessages = new BN(latest_generated_nonce).sub(new BN(latest_received_nonce));
-
-        setOutboudLanes({
-          latestReceivedNonce: latest_received_nonce.toString(),
-          pendingMessages: pendingMessages.isNeg() ? '0' : pendingMessages.toString(),
-          totalMessages: latest_generated_nonce
-        });
+      getLaneData({
```
tomusdrw, line 46: Enclosing this code into functions would make sense if we avoided touch `api` directly, but few lines above we still do `api.query[bridgedMessages]` for instance.

I think ideally we wouldn't even avoid calling into `getSubstrateDynamicNames` here. Basically the entire:
1. polkadot.js/api logic
2. any runtime-specific calls
would be abstracted away for these kind of components.

Of course it's easier said than done, so what I'm saying here is that this is not incorrect, but I'd like to challenge you to go a bit beyond this.

src/api/getBridgedBlocks.ts
wirednkod: Im trying to figure out if there is a way of a reusable hook/function that will bind Api usage with ApiSubscriptions together. Noted your comment though

src/api/getBridgedBlocks.ts
wirednkod: Im trying to figure out if there is a way of a reusable hook/function that will bind Api usage with ApiSubscriptions together. I want to be reusable but not too generic that will make typings disappear (that would be a NO)

hbulgarini: what about if you create a third hook called `useApiCallSubscribed` or `useApiCallWithSubscription` that wraps the hooks `useGetApi` and `useApiSubscription`?

```
function useApiCallSubscribed(getBlock, ...args){
   const {executeCb, data, error, isLoading} = useGetApi(getBlock, ...args):
   useApiSubscription(executeCb, isReady)

   return {data, error, isLoading}
}
```

hbulgarini: IMHO i think that having two separate files makes the things easier to understand since you wouldn't need a separator either an `if` block logic.

src/hooks/subscriptions/useMessagesLane.ts
hbulgarini: Supporting what tomek said , in PR #186 there is this file that is already taking into account what he propose: https://github.com/paritytech/parity-bridges-ui/pull/186/files#diff-7de2939806355f65b85bb93c1220b85ffe6a7f55fb0eb9aef1a7f7295aabdd8b

src/api/getBridgedBlocks.ts
wirednkod: > what about if you create a third hook called `useApiCallSubscribed` or `useApiCallWithSubscription` that wraps the hooks `useGetApi` and `useApiSubscription`?
> 
> ```
> function useApiCallSubscribed(getBlock, ...args){
>    const {executeCb, data, error, isLoading} = useGetApi(getBlock, ...args):
>    useApiSubscription(executeCb, isReady)
> 
>    return {data, error, isLoading}
> }
> ```

I have already created that since yesterday (unfortunately for some reasons there are some errors found (even though I cannot find a leak)). I will push the code anyway so you can see what I mean.

src/hooks/subscriptions/useMessagesLane.ts
wirednkod: > but few lines above we still do api.query[bridgedMessages] for instance.

As far as I understand the `api.query[bridgedMessages` is actually not necessary at that line. I will push an update now that fixes it and even maybe accomplishes the challenge - but there is a glitch that I do not like so much

src/api/ApiSubscription.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
```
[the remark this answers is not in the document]
wirednkod: This is still draft - some files are there and will be removed when its ready for review. I do not want to alter files that are used from other parts of the code - just to test my solution

wirednkod:
https://github.com/paritytech/parity-bridges-ui/pull/190#discussion_r652892191

That is what I work on - but its quite a challenge

src/api/getBridgedBlocks.ts
wirednkod: [This](https://github.com/paritytech/parity-bridges-ui/blob/nik-fix-apis-on-root/src/api/useApiCallAndSubscriptions.ts) is the newly created one that takes care of everything. Again though - still draft and still working on it

wirednkod: @tomusdrw if function is separated into 2, then there is no way to "automate" the useGetBridges hook in order to be called from useEffects etc. This actually breaks the concept introduced in `idea.ts`. Of course it can be done but it is a different approach

wirednkod:
As agreed with @hbulgarini in order to avoid code conflicts - we will keep this PR open as draft in order to cherry-pick ideas from here, and continue with PR #186 as an overall first-step solution.

wirednkod:
I think this can close now

PR #194, opened 2021-06-21
Adding warning on onmounted component state

### LINKED ISSUE #141: Warn in case we keep updating state on unmounted component (leak detection)

hbulgarini:
Closes #141

### DIFF
--- a/src/hooks/react/useMountedState.ts
+++ b/src/hooks/react/useMountedState.ts
@@ -14,8 +14,9 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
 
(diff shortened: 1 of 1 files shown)


wirednkod (approved):
lgtm

tomusdrw (approved):

Issue #191, opened 2021-06-16
Create E2E testing for receiver logic.


hbulgarini:
Test 1: Should provide source address type and should show companion.
Test 2: Should provide generic address type and should show two options and choose one.

Issue #192, opened 2021-06-16
Integrate E2E to CI.


hbulgarini:
Integrate E2E testing to our CI pipeline.

hbulgarini:
Requested to CI team: https://github.com/paritytech/ci_cd/issues/153

PR #196, opened 2021-06-22
Create E2E testing for receiver logic.

hbulgarini:
Related to #191

This PR extends current E2E testing with the following enhancements

There are now 4 testing asserts:

Test 1: Should provide native source address type and should show native address in receiver field.
Test 2: Should provide companion source address type and should show companion address in receiver field.
Test 2: Should provide generic address type and should show two options,  choose native and get native receiver field.
Test 4: Should execute end to end testing of transaction. This includes the following checks:

- Waits for transaction broadcasted message.
- Check that all steps get completed.
- Checks that balance is updated accordingly.


e2e/app.test.js
```diff
@@ -0,0 +1,274 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
```
wirednkod, line 1: I do not understand why this test exist? It launches a browser and runs tests just like the puppeteer E2E:alone test does.
IMO we should have 3 tests.
1) functionality tests (you already have build those)
2) E2E test (1 happy path test that shows that flow works - already built those)
3) component tests (these does not have to open a browser in order to run)

The way this is implemented IMO is not correct as it takes way too much time to run (open browser, wait, run, etc etc) and it should be able to run also in CI (which this kind of tests are not able i think)

package.json
```diff
@@ -50,8 +50,8 @@
     "prestart": "yarn downloadCustomTypes",
     "start": "PORT=3001 react-app-rewired start",
     "build": "react-app-rewired build",
-    "test:e2e": "jest -c e2e/jest.config.js",
-    "test:e2e-w": "jest -c e2e/jest.config.js --watch",
+    "test:e2e": "jest -c e2e/jest.config.js --detectOpenHandles",
```
wirednkod, line 53: I think `--detectOpenHandles` should be used only when we want to identify leaks (and Jest informs about them anyway) not in each run

e2e/app.test.js
hbulgarini: Well i think that it does not have to be so drastic... the additional tests added do not add more than 30 seconds or similar so even if it adds minutes to the process is something we can deal with more taking into account that the gain is significant in terms of consistency and quality.

Then I agree with your three suggestions but the purpose of this PR is to extend and cover all the critical path of the transfer feature and the complex logic around the receiver field that needs to work perfectly and this tests validate that.

The reason for including the tests on this jest file is because we can combine puppeter tools with powerful jest assertions which i think it's the best of two worlds.

Finally , the next step is to include these tests to the CI so we can run these tests in every deploy of master branch: #192

wirednkod: IMO all test should not take more than 20 seconds. In Substrate connect, covering 2 major projects with approx 80-90% coverage of unit tests takes 20 seconds. 
In addition i do not understand how exactly these tests will run in CI

hbulgarini: Ok fair enough ... 

The goal for the tests are described in the description of this PR:

Test 1: Should provide native source address type and should show native address in receiver field.
Test 2: Should provide companion source address type and should show companion address in receiver field.
Test 2: Should provide generic address type and should show two options, choose native and get native receiver field.
Test 4: Should execute end to end testing of transaction. This includes the following checks:

wirednkod: I know. I guess what i am trying to say is that these tests can run without even initiating a browser as far as I know

e2e/jest.config.js
```diff
@@ -18,9 +18,9 @@ module.exports = {
   preset: 'jest-puppeteer',
   globals: {
     SERVER_URL: 'http://localhost:3001',
-    JEST_TIMEOUT: 10000
+    JEST_TIMEOUT: 100000
```
tomusdrw, line 21: Is it because the 4th test takes so long?

package.json
tomusdrw: What's the drawback of using it always?

src/components/Icons.tsx
```diff
@@ -44,15 +44,15 @@ interface IconApiStatusProps {
 export const IconTxStatus = ({ status }: IconTxStatusProps) => {
   switch (status) {
     case TransactionStatusEnum.COMPLETED:
-      return <CheckCircleOutlineIcon />;
+      return <CheckCircleOutlineIcon id="check-circle-icon" />;
```
tomusdrw, line 47:
```suggestion
      return <CheckCircleOutlineIcon id="test-transaction-status-completed" />;
```
What's the point of bounding this to a concrete component type? 

I'd also distinguish for-tests ids by using `test` prefix everywhere.

e2e/app.test.js
tomusdrw: We shouldn't be adding any more E2E tests for now, but the ones that are here can stay. The focus should be more on unit testing, until we get some major features (like a new tab). 

Running it on the CI is a priority (at least after merging to master and deploying).

> I guess what i am trying to say is that these tests can run without even initiating a browser as far as I know

What's the alternative? Does pupetteer have some built-in browser?

e2e/app.test.js
```diff
@@ -0,0 +1,274 @@
...
+};
+
+const timeout = 500000;
+
+const ids = {
+  native: '#test-native-input',
+  companion: '#test-companion-input',
+  sender: '#test_sender_component',
+  transferButton: '#test-button-submit',
+  senderComponent: '#test_sender_component'
```
tomusdrw, line 54: Consistency please!!! Either `_` or `-` everywhere.

package.json
wirednkod: 2 drawbacks:
a) It implies `--runInBand`, making tests run serially.
b) It has a significant performance penalty

Documentation proposes that this should only be used for debugging.

e2e/app.test.js
wirednkod: > We shouldn't be adding any more E2E tests for now, but the ones that are here can stay. 

I agree. As soon as the work is done - there is no need for removing or anything.

> Running it on the CI is a priority

Im not sure how one can run GUI tests in Github CI - I do not say that it cannot happen - I just do not know how (never done it)

> Does puppetteer have some built-in browser

Puppeteer's browsers can be downloaded along with it (meaning you can add chromium and/or firefox with the installation of the package and run the tests with the downloaded browsers (cross-browsing testing). 
The configuration I did, does not include any browser and it is assumed that chrome is installed with it.

On top of that, puppeteer can run headless browsers (without the need of any GUI) and that is my point above. Lets run the E2E or component tests, but not with full browsers and without such a waiting time. 

I would suggest - since new components implementation start - to pause unit (component tests for now) and implement them whenever a new component is created (e.g. build a new dropdown -> write also unit/component tests for it on the way).

hbulgarini: @wirednkod the browser setup is something i plan to work on #192, this PR only includes the new tests described before.

e2e/app.test.js
hbulgarini: omg!

package.json
hbulgarini: Right. It is not needed .. i have removed the parameter.

e2e/jest.config.js
hbulgarini: not really, it can remain as it was before. I rolled back the change.

src/components/Icons.tsx
hbulgarini: I use this component to validate in the UI if the step was completed or not.

tomusdrw: > I use this component to validate in the UI if the step was completed or not.

I understand that. My point is that you are relying on a very specific component to be displayed in something that is a high-level test. The point is that if we decide to change the icon you will need to update the `id` and all references because the id is so tightly coupled with the component being displayed.
Also imagine that we want to write a new test (or assertion). To understand what the test is doing, you need to understand what is `test-check-circle-icon` component and in what circumstances it's display (pretty much the only way to do it so to go to the component and read the code), after doing that you will actually know that: "Aha, that component is displayed when the transaction is completed, so let me use the id.".

So long story short, by using `ids` like that:
1. You are tightly coupling specific design choice (circle icon component) with high-level application logic (transaction is completed). Changing designs will cause extra changes in completely unrelated parts of code.
2. You are creating additional mental overhead to understand what "circle icon component" actually means (transaction is completed).

tomusdrw (approved):

src/components/Icons.tsx
hbulgarini: Ok, fair enough ... i will switch it to `test-transaction-step-completed` then

PR #197, opened 2021-06-22
Dockerize bridges UI

### LINKED ISSUE #87: Add Dockerfile and publish images to dockerhub.
A special docker-optimized version should be built and pushed to dockerhub.

Ideally if the version was network agnostic (see #86) and it should only expect two nodes running on `localhost` (or somehow configurable in docker (BUILD_ARG maybe?)).

By running it locally and forwarding ports you should be able to have it connected to chains running on the host machine.

wirednkod:
Fixes #87

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,4 +91,4 @@ docker run --rm -it -p 8080:80 parity-bridges-ui:dev
 ```
 
 (optional) The `--env-file` param can be added in the command above if you intend to alter the default [ENV params](https://github.com/paritytech/parity-bridges-ui/blob/master/.env).
-(e.g. `docker run --rm -it --env-file .env -p 8080:80 parity-bridges-ui:dev` )
+(e.g. `docker build --env-file .env -t parity-bridges-ui:dev ` )
--- a/README.md
+++ b/README.md
@@ -89,6 +89,3 @@ Now that image is built, container can start with the following command, which w
 ```
 docker run --rm -it -p 8080:80 parity-bridges-ui:dev
 ```
-
-(optional) The `--env-file` param can be added in the command above if you intend to alter the default [ENV params](https://github.com/paritytech/parity-bridges-ui/blob/master/.env).
-(e.g. `docker build --env-file .env -t parity-bridges-ui:dev ` )
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ _NOTE: alter url once this is moved_
 
 To build the image run the:
 ```
-sudo docker build -t parity-bridges-ui:dev .
+docker build -t parity-bridges-ui:dev .
 ```
 
 Now that image is built, container can start with the following command, which will serve our app on port 8080.
--- a/README.md
+++ b/README.md
@@ -77,9 +77,6 @@ For additional information about the Bridges Project please refer to [parity-bri
 
 ## Docker
 
-Can be found in [dockerhub](https://hub.docker.com/repository/docker/wirednkod13/parity-bridges-ui).
-_NOTE: alter url once this is moved_
-
 To build the image run the:
 ```
 docker build -t parity-bridges-ui:dev .
--- a/README.md
+++ b/README.md
@@ -74,3 +74,20 @@ In case there is an unexpected issue with this process you can test this process
 ## Learn More
 
 For additional information about the Bridges Project please refer to [parity-bridges-common](https://github.com/paritytech/parity-bridges-common) repository.
+
+## Docker
+
+Can be found in [dockerhub]().
+
+To build the image run the:
+```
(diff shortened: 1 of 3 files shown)


dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# install app dependencies
+COPY package.json ./
+COPY yarn.lock ./
+RUN yarn
+
+# add app
+COPY . ./
+
+# start app
+CMD ["npm", "start"]
```
tomusdrw, line 22: This will start a development version, we rather want a production version, where we:
1. First build the static files.
2. Serve the static files using nginx.

wirednkod: > 2. Serve the static files using nginx.

Will the nginx be included inside the bridges-ui docker or someone should run it as an external docker conatiner?

README.md
```diff
@@ -74,3 +74,21 @@ In case there is an unexpected issue with this process you can test this process
...
+docker build -t parity-bridges-ui:dev .
+```
+
+Now that image is built, container can start with the following command, which will serve our app on port 8080.
+```
+docker run --rm -it -p 8080:80 parity-bridges-ui:dev
+```
+
+(optional) The `--env-file` param can be added in the command above if you intend to alter the default [ENV params](https://github.com/paritytech/parity-bridges-ui/blob/master/.env).
+(e.g. `docker run --rm -it --env-file .env -p 8080:80 parity-bridges-ui:dev` )
```
tomusdrw, line 94:
```suggestion
(e.g. `docker build --env-file .env -t parity-bridges-ui:dev ` )
```
I think the env file will not matter when the image is already built, `nginx` doesn't really care :)

dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+RUN apk update
+RUN apk add git
+
+# add `/parity-bridges-ui/node_modules/.bin` to $PATH
+ENV PATH /parity-bridges-ui/node_modules/.bin:$PATH
+
+# install app dependencies
+COPY . .
+RUN yarn && yarn build
```
tomusdrw, line 15: To speed up build times it's common to split depenedencies installation from actual build process, so that dependencies can be cached.
```
COPY package.json .
COPY yarn.lock .
RUN yarn
COPY . .
RUN yarn build
```

This should give much better times when re-building the image and there is no `package.json` changes.

dockerfile
```diff
@@ -0,0 +1,26 @@
...
+FROM node:14.17-alpine AS builder
+
+# set working directory
+WORKDIR /parity-bridges-ui
+
+RUN apk update
+RUN apk add git
+
+# add `/parity-bridges-ui/node_modules/.bin` to $PATH
+ENV PATH /parity-bridges-ui/node_modules/.bin:$PATH
```
tomusdrw, line 11: Why is this needed exactly? I don't think we us anything specific from that directory, do we?

dockerfile
```diff
@@ -0,0 +1,26 @@
...
+RUN yarn && yarn build
+
+# nginx state for serving content
+FROM nginx:alpine
+# Set working directory to nginx asset directory
+WORKDIR /usr/share/nginx/html
+# Remove default nginx static assets
+RUN rm -rf ./*
+# Copy static assets from builder stage
+COPY --from=builder /parity-bridges-ui/build .
```
tomusdrw, line 24: Isn't this going to create a `build` subdirectory instead of copying all the files?

dockerfile
wirednkod: I think that is stray from testing times

dockerfile
wirednkod: I think it just copies the static files to .
I have tested it and everything seemed to serve correctly

README.md
```diff
@@ -74,3 +74,18 @@ In case there is an unexpected issue with this process you can test this process
 ## Learn More
 
 For additional information about the Bridges Project please refer to [parity-bridges-common](https://github.com/paritytech/parity-bridges-common) repository.
+
```
hbulgarini, line 77: Personally i would not add this on the documentation until it is moved to `paritytech` hub .

wirednkod: This is parity open-source property and I provide the credentials. Why would not add that?

hbulgarini: What i'm pointing out is to avoid sharing the docker image that resides in your personal docker profile: `https://hub.docker.com/repository/docker/wirednkod13/parity-bridges-ui`

I would wait to announce this docker image for th UI once it resides in the `paritytech` hub.
Let's suppose that in the short term you make any changes on your profile that makes the image unavailable.

hbulgarini (approved):
lgtm!

tomusdrw (approved):

Issue #200, opened 2021-06-23
TransactionStatus card should have "Payload" and "Decoded Call" modes


goldsteinsveta:
<img width="1131" alt="Screenshot 2021-06-23 at 16 48 46" src="https://user-images.githubusercontent.com/8782596/123118478-f5b2d500-d442-11eb-9f7e-7f6e749d6d73.png">

Optional: Payload should be a default mode for CustomCall action

hbulgarini:
@goldsteinsveta  i found the library [react-json-view](https://github.com/mac-s-g/react-json-view) which empowers the JSON.stringify and it adds additional features like collapse, copy ,etc. Take into account that some properties from the payload contains quite a few elements so this tool helps to visually display it better using collapse feature. Here some screenshots:

![Screen Shot 2021-06-28 at 10 31 47](https://user-images.githubusercontent.com/12089572/123644933-3d977a80-d7fc-11eb-98fd-97b7a2378e34.png)
![Screen Shot 2021-06-28 at 10 31 47](https://user-images.githubusercontent.com/12089572/123645016-51db7780-d7fc-11eb-9e94-b1bf7c1f7764.png)
![Screen Shot 2021-06-28 at 10 31 55](https://user-images.githubusercontent.com/12089572/123645034-5738c200-d7fc-11eb-88c0-f20a406a6b8e.png)
![Screen Shot 2021-06-28 at 10 32 13](https://user-images.githubusercontent.com/12089572/123645050-5b64df80-d7fc-11eb-8fd5-9d8d8d82ea17.png)


what do you think?

tomusdrw:
We should keep it simpler (no toggles, imho no extra library needed) and:
1. `SourceAccount` should be formatted as hex.
2. `Call` should be decode using `target` chain custom types and displayed in a readable form, not a set of bytes.

hbulgarini:
> 1. `SourceAccount` should be formatted as hex.

Can i show the account value as:

origin: {
  "SourceAccount": "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY"
}

tomusdrw:
Yes, SS58 formatting with (source) network prefix would be best. We don't necessarily need the raw hex form, I didn't phrase that properly :)

hbulgarini:
> Yes, SS58 formatting with (source) network prefix would be best. We don't necessarily need the raw hex form, I didn't phrase that properly :)

Something like :

{
[...]

Issue #67, opened 2021-04-16
Replace semantic UI  for Material UI


hbulgarini:

goldsteinsveta:
I see you haven't closed this one with #69, so feel free to reassign this issue to me. I wouldn't close it at once though, but component by component remove styled-components/SUI dependencies everywhere

Issue #195, opened 2021-06-22
Refactor useSendMessage and useTransaction preparation to use wrapped calls and subscriptions.


hbulgarini:
useSendMessage and useTransactionPreparation are using raw calls to the api and we are not taking the benefit of the wrapped call work done before.

Also we should consider moving the `setIsRunning` local state to reducer logic.

PR #203, opened 2021-06-28
Adding payload and decoded tabs per transaction.

hbulgarini:
This PR adds the payload and decoded tabs per transaction detailed in #200.

Also i have included the json display proposal discussed on https://github.com/paritytech/parity-bridges-ui/issues/200#issuecomment-869690103 . In case @goldsteinsveta  or @tomusdrw do not agree with the proposal this PR can be cancelled and rebuilt accordingly. IMHO , the proposal adds value to the payload inspection and in case the raw payload is needed, the copy feature can be used.

EDIT: `semantic-ui-react` package was finally removed from the project. Related #67 .


wirednkod (commented):
Overall looks good - but the result is not as per #200 screenshots. There are details missing (suvh as icons and/or titles per tab)

src/components/TransactionSwitchTab.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+const TransactionSwitchTab = ({ transactionDisplayProps, children, payloadHex, TransactionDisplayPayload }: Props) => {
+  const [tab, setTab] = useState(SwitchTabType.RECEIPT);
+
+  const getColorByState = (value: string) => (value === tab ? 'primary' : 'secondary');
+
+  return (
+    <>
+      <Box mt={2}>
+        <ButtonSwitchMode color={getColorByState(SwitchTabType.PAYLOAD)} onClick={() => setTab(SwitchTabType.PAYLOAD)}>
```
tomusdrw, line 42: Every `onClick` should be `useCallback` really, but I can turn my blind eye on this.

src/hooks/transactions/useTransactionPayloadDisplay.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (!receiverAddress) {
+      setPayloadHex(null);
+      setTransactionDisplayPayload(null);
+    }
+    if (payload) {
+      //@ts-ignore
+      const payloadType = createType(sourceChain, 'OutboundPayload', payload);
+      setPayloadHex(payloadType.toHex());
+    }
+    //@ts-ignorex
```
tomusdrw, line 45:
```suggestion
    //@ts-ignore
```

src/util/transactionUtils.ts
```diff
@@ -14,8 +14,39 @@
...
+    TransactionDisplayPayload.origin = {
+      SourceAccount: toObject(origin.SourceAccount)
+    };
+
+    TransactionDisplayPayload.weight = weight;
+    TransactionDisplayPayload.spec_version = spec_version;
+    return TransactionDisplayPayload;
+  }
+  return null;
+}
```
tomusdrw, line 52:
```suggestion
export function getTransactionDisplayPayload(payload: Payload): TransactionDisplayPayload {
  const transactionDisplayPayload = {} as TransactionDisplayPayload;
  const { call, origin, spec_version, weight } = payload;
  transactionDisplayPayload.call = toObject(call);
  transactionDisplayPayload.origin = {
    SourceAccount: toObject(origin.SourceAccount)
  };

  transactionDisplayPayload.weight = weight;
  transactionDisplayPayload.spec_version = spec_version;
  return transactionDisplayPayload;
}
```

1. Keep the `null` check outside of the function (you're doing it anyway on the call site).
2. Use lower-case names for variables 😱

src/hooks/transactions/useTransactionPayloadDisplay.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  useEffect(() => {
+    if (!receiverAddress) {
+      setPayloadHex(null);
+      setTransactionDisplayPayload(null);
+    }
+    if (payload) {
+      //@ts-ignore
+      const payloadType = createType(sourceChain, 'OutboundPayload', payload);
+      setPayloadHex(payloadType.toHex());
```
tomusdrw, line 43: Incorrect. If `payload==null` the old hex will be left in state.

src/types/transactionTypes.ts
```diff
@@ -65,6 +78,8 @@ export interface TransactionState {
   payload: TransactionPayload | null;
   payloadError: string | null;
   estimatedFeeLoading: boolean;
+  payloadHex: string | null;
```
tomusdrw, line 81: group with other `payload` types.

src/types/transactionTypes.ts
```diff
@@ -65,6 +78,8 @@ export interface TransactionState {
...
+  TransactionDisplayPayload: TransactionDisplayPayload;
```
tomusdrw, line 82: lowercase
```suggestion
  transactionDisplayPayload: TransactionDisplayPayload;
```

src/util/transactionUtils.ts
```diff
@@ -14,8 +14,39 @@
...
+  TransactionDisplayPayload
+} from '../types/transactionTypes';
 
 export function isTransactionCompleted(transaction: TransactionStatusType): boolean {
   return transaction.status === TransactionStatusEnum.COMPLETED;
 }
+
+interface ToObject {
+  [propName: number]: number;
+}
```
tomusdrw, line 30: Why is this needed at all?

src/util/transactionUtils.ts
```diff
@@ -14,8 +14,39 @@
...
+
+function toObject(arr: Array<any>): Object {
+  const rv: ToObject = {};
+  for (let i = 0; i < arr.length; ++i) rv[i] = arr[i];
+  return rv;
+}
```
tomusdrw, line 36:
```suggestion
function toObject(arr: Array<any>): Object {
 return Object.fromEntries(Object.entries(arr));
}
```

I'd suggest to make it internal to `getTransactionDisplayPayload` - with shortr implementation we can even inline this.

tomusdrw (commented):
Can you please add a screenshot of what it looks like?

hbulgarini:
> #200

Yes! i need to add the header thanks!

hbulgarini:
> Can you please add a screenshot of what it looks like?

Screenshots are attached [here](https://github.com/paritytech/parity-bridges-ui/issues/200#issuecomment-869690103), do you need additional screenshots?

src/util/transactionUtils.ts
hbulgarini: it is not needed anymore after your proposal.

src/util/transactionUtils.ts
hbulgarini: 1. OK
2. Daaammmnnn 😱 , that was a mistake that happened after a replace all and i didn't realize

src/hooks/transactions/useTransactionPayloadDisplay.ts
hbulgarini: i don't know why this is even here.

src/hooks/transactions/useTransactionPayloadDisplay.ts
hbulgarini: correct.

src/components/TransactionSwitchTab.tsx
hbulgarini: no. it should be as you mention.

tomusdrw:
> Screenshots are attached here, do you need additional screenshots?

Yes. I'd like to see how it's integrated with the current UI, not only the payload part.

src/components/Transactions.tsx
```diff
@@ -34,6 +35,7 @@ const Transactions = ({ type, ...transactionDisplayProps }: Props) => {
   const { transactions } = useTransactionContext();
   const { dispatchTransaction } = useUpdateTransactionContext();
   const { dispatchMessage } = useUpdateMessageContext();
+  useResetTransactionState(type);
```
tomusdrw, line 38: Can you elaborate what this is about? What is `type` and why if we set it back to a previous value it should reset the transaction state?

src/contexts/TransactionContext.tsx
```diff
@@ -57,7 +57,9 @@ export function TransactionContextProvider(props: TransactionContextProviderProp
     payloadError: null,
     showBalance: false,
     formatFound: null,
-    estimatedFeeLoading: false
+    estimatedFeeLoading: false,
+    payloadHex: null,
```
tomusdrw, line 61: Sort all props please or at least move related stuff next to each other :)

src/hooks/transactions/useTransactionPayloadDisplay.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { TransactionDisplayPayload } from '../../types/transactionTypes';
+import { getTransactionDisplayPayload } from '../../util/transactionUtils';
+import useApiCalls from '../api/useApiCalls';
+import { useAccountContext } from '../../contexts/AccountContextProvider';
+
+const useTransactionPayloadDisplay = () => {
+  const { payload } = useTransactionContext();
+  const [payloadHex, setPayloadHex] = useState<string | null>(null);
+  const [transactionDisplayPayload, setTransactionDisplayPayload] = useState<TransactionDisplayPayload | null>(null);
```
tomusdrw, line 29: I feel these two states could be turned into one. We change them in unison anyway.

src/util/transactionUtils.ts
```diff
@@ -14,8 +14,46 @@
...
+  payload,
+  ss58Format,
+  account,
+  createType,
+  targetChain
+}: PayloadInput): TransactionDisplayPayload {
+  //@ts-ignore
+  const callType = createType(targetChain, 'BridgedOpaqueCall', payload.call);
+  //@ts-ignore
+  const call = createType(targetChain, 'Call', callType.toHex());
```
tomusdrw, line 47:
```suggestion
  const call = createType(targetChain, 'Call', payload.call);
```
I think this should work just fine, no?

tomusdrw (commented):
Awesome stuff! Thanks for the screenshots.

I think ideally if we could convert `callIndex` to an actual name of the pallet and dispatchable (`balances::transfer`), but let's make it a stretch goal in a separate PR :)

I also feel there is a bit too much happening in the JSON-display component. Is it styleabe? I'd honestly hide a bunch of things, like:
- number of items
- types
- "toggle triangles" (dunno the professional name :)) could be made lighter
- the vertical guides could be removed

CC @goldsteinsveta wdyt?

hbulgarini:
@goldsteinsveta  i had to do a couple of UX change in order to be able to show a the payload of a second transaction that was not confirmed yet:  The`TransactionStatusMocked`component ([code](https://github.com/paritytech/parity-bridges-ui/blob/hb-add-payload-decoded-tab/src/components/Transactions.tsx#L42))  used to get unmounted after the first transaction so as a result the payload currently being built could not be checked until the second transaction was confirmed. I thought that the easiest solution until we fully implement #108.

Also once the transaction starts executing the receiver form gets completed reseted, which i think for now is the best.
Again in #108 the idea would be to lock the form .

src/util/transactionUtils.ts
hbulgarini: i thought the same, but the api complained when it tried to built the type in that way.

src/components/Transactions.tsx
hbulgarini: Of course ... basically when i switch actions tabs ... let's say from "Transfer" to "CustomCall" the payload and estimatedFee needs to be reseted otherwise the tabs would show the values calculated in the previous tab.

Anyway it is not working 100% as expected since for this specific scenario i have an issue:

1) Fill the form for a Transaction tab.
2) Switch action to Custom Call

Estimated Fee remains the same as the one calculated from the previous tab. This is only happen for this scenario, the other switch do not face this issue. As i considered that this issue was out of the scope of this PR i have logged the following issues in order to track this:

https://github.com/paritytech/parity-bridges-ui/issues/195
https://github.com/paritytech/parity-bridges-ui/issues/209
https://github.com/paritytech/parity-bridges-ui/issues/207

src/contexts/TransactionContext.tsx
hbulgarini: sure ... i will also do it on the `transactionTypes`

src/util/transactionUtils.ts
tomusdrw: Maybe `payload.call` needs to be converted to hex first? I'm fine with this staying as-is, but it seems redundant ;)

hbulgarini: >  I'm fine with this staying as-is, but it seems redundant ;)

I know ! i believe the same as you but for some reason that i don't understand why it does not work. This is the error i get:
`Error: createType(Call):: findMetaCall: Unable to find Call with index 0x9490/[148,144]`

src/types/transactionTypes.ts
```diff
@@ -33,6 +33,11 @@ export enum TransactionStatusEnum {
   FAILED = 'FAILED'
 }
 
+export enum SwitchTabType {
```
goldsteinsveta, line 36: naming: SwitchTabEnum?

src/components/TransactionPayload.tsx
```diff
@@ -0,0 +1,96 @@
...
+  return (
+    <Card elevation={transactionDisplayProps?.size === 'sm' ? 23 : 24} className={classes.card}>
+      {payloadHex && <TransactionHeader type={type} status={status} />}
+      {estimatedFeeLoading && (
+        <div className={classes.loader}>
+          <CircularProgress />
+        </div>
+      )}
+      {!estimatedFeeLoading && tab === SwitchTabType.PAYLOAD && payloadHex}
+      {!estimatedFeeLoading && tab === SwitchTabType.DECODED && transactionDisplayPayload && (
```
goldsteinsveta, line 89: just curious, why do you need this `&& transactionDisplayPayload` condition?

src/components/TransactionPayload.tsx
```diff
@@ -0,0 +1,96 @@
...
+        <ReactJson src={transactionDisplayPayload} enableClipboard name={false} />
```
goldsteinsveta, line 90: typography nit: could you please wrap this in `<Typography variant="subtitle2" component="span"/>`. Thanks!

goldsteinsveta (approved):
This is beautiful!

src/types/transactionTypes.ts
hbulgarini: sure!

src/components/TransactionPayload.tsx
hbulgarini: It was there because of ReactJson component's src prop demands an object type and do not allow undefined.

src/components/TransactionPayload.tsx
hbulgarini: of course!

PR #204, opened 2021-06-29
Fix large amount bug

### LINKED ISSUE #111: Large balance in `amount` crashes the app

wirednkod:
Fix issue that is depicted on #111 
The amount is passed to dispatcher (global state) across rest of inputs (estimatedFee etc) instead of being a local state.
This way it can be handled across multiple components (show error, reactions of UI etc).

The 2 highest/lowest levels of token were commented out as 1e21 and more, is too big amount for JS to handle correctly (even UI wise).

Fix tests and added extra validations to ensure that there are no cases missing.

Fixed issue mentioned by Hector [here](https://github.com/paritytech/parity-bridges-ui/issues/111#issuecomment-866167093)

Fixes #111


src/hooks/transactions/useTransactionType.ts
```diff
@@ -53,11 +55,14 @@ export default function useTransactionType({ input, type, weightInput }: Props):
     weight: null
   });
 
+  const planck = 10 ** targetApi.registry.chainDecimals[0];
+  const [resInp, msg] = evalUnits((parseFloat(input) * planck).toString());
```
tomusdrw, line 59: This `parseFloat` and multiplication should be done in big number context already (same inside `evalUnits`). While it most likely doesn't crash it will loose precision here. Try something like `1.00...001` with large amount of zeros:)

src/hooks/transactions/useTransactionType.ts
```diff
@@ -53,11 +55,14 @@ export default function useTransactionType({ input, type, weightInput }: Props):
...
+  const trnsfInput = resInp?.toString() || '0';
```
tomusdrw, line 60:
```suggestion
  const transferInput = resInp?.toString() || '0';
```
gbbrsh

src/util/evalUnits.ts
```diff
@@ -14,8 +14,10 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
 const si = [
-  { value: -1e24, symbol: 'y' },
-  { value: -1e21, symbol: 'z' },
+  // Deactivate these huge nums for now
+  // > than 1e21 is too big for JS to handle
+  // { value: -1e24, symbol: 'y' },
```
tomusdrw, line 19: Can we use big num instead?

src/util/tests/evalUnits.test.ts
```diff
@@ -91,4 +85,10 @@ describe('Tests suite - evalUnits', () => {
     expect(actualValue).toBeFalsy;
     expect(msg).toBe(EvalMessages.GIBBERISH);
   });
+
+  it('Should accept as input a HUGE number and return error message', () => {
```
tomusdrw, line 89: I'd rather see this handled correctly, but yeah, better than crashing :)

src/util/evalUnits.ts
wirednkod: Alter all parts of code and now we use BN

src/util/tests/evalUnits.test.ts
wirednkod: You will now see it handled correctly

src/hooks/transactions/useTransactionType.ts
```diff
@@ -53,11 +56,14 @@ export default function useTransactionType({ input, type, weightInput }: Props):
     weight: null
   });
 
+  const planck = 10 ** targetApi.registry.chainDecimals[0];
+  const [resInp, msg] = evalUnits(new BN(input).mul(new BN(planck)).toString());
+  const transferInput = resInp?.toString() || '0';
```
hbulgarini, line 61: I think this logic can be included in the reducer and spawn the formatted value across all the dependent components. You would have to add to the `setTransferAmount` action the `targetApi.registry.chainDecimals[0]` so you can use it in the reducer through the action payload. Also there you could omit all your validations like `transferInput || "0"` in components call:

```
       const {chainDecimals,input} = action.payload;
       const planck = 10 ** chainDecimals;
       const [resInp, msg] = evalUnits(new BN(input).mul(new BN(planck)).toString());
       const transferInput = resInp?.toString() || '0';
      return {
        ...state,
        transferAmount: transferInput,
        transferAmountError: action.payload.transferAmountError
      };
}
```

btw ... are you using `msg` that comes from `evalUnits` ?

src/components/Transfer.tsx
```diff
@@ -47,54 +50,51 @@ const useStyles = makeStyles((theme) => ({
...
 
   // To extract estimated fee logic to specific component. Issue #171
   useEffect((): void => {
     estimatedFee &&
-      actualInput &&
-      setAmountNotCorrect(new BN(balance.free).sub(new BN(actualInput).add(new BN(estimatedFee))).toNumber() < 0);
-  }, [actualInput, estimatedFee, balance, isApiReady]);
+      transferAmount &&
+      setAmountNotCorrect(new BN(balance.free).sub(new BN(transferAmount).add(new BN(estimatedFee))).isNeg());
+  }, [transferAmount, estimatedFee, balance, isApiReady]);
```
hbulgarini, line 89: Since we have now "smarter" reducers that handle more complex logic, this useEffect could be removed and let the `setTransferAmount` action to take care of it. Then you can use `transferAmountError` to set whatever error message you would like to share with the user.

src/components/Transfer.tsx
```diff
@@ -47,54 +50,51 @@ const useStyles = makeStyles((theme) => ({
...
+    dispatchTransaction(
+      TransactionActionCreators.setTransferAmount(null, actualValue !== null ? actualValue?.toString() : '')
+    );
   };
 
   useEffect((): void => {
-    isRunning && setTransferInput('');
-  }, [isRunning]);
+    isRunning && dispatchTransaction(TransactionActionCreators.setTransferAmount(null, ''));
+  }, [dispatchTransaction, isRunning]);
```
hbulgarini, line 82: It is a shame that we still rely on a local state for determining if the execution is running or not, but if that status would be in the reducer then this useEffect might be removed as well. To be considered during this: #208

src/reducers/transactionReducer.ts
```diff
@@ -165,6 +165,12 @@ export default function transactionReducer(state: TransactionState, action: Tran
         estimatedFeeError: action.payload.estimatedFeeError,
         estimatedFeeLoading: action.payload.estimatedFeeLoading
       };
+    case TransactionActionTypes.SET_TRANSFER_AMOUNT:
+      return {
+        ...state,
+        transferAmount: action.payload.transferAmount,
+        transferAmountError: action.payload.transferAmountError
+      };
```
hbulgarini, line 173: https://github.com/paritytech/parity-bridges-ui/pull/204/files#r661025605

hbulgarini (changes requested):

src/hooks/transactions/useTransactionType.ts
wirednkod: This logic cannot be included in the reducer as this will make alter the purpose of `case TransactionTypes.TRANSFER:` inside of useTransactionType and a need will be risen for it to be removed, thus decoupling the cases to more than one files.

src/components/Transfer.tsx
wirednkod: See above.

src/util/evalUnits.ts
```diff
@@ -40,11 +44,25 @@ const alphaInts = /^[0-9]*[a-zA-Z]{1}$/;
...
+const transformShorthandToBN = (input: string): BN | null => {
+  // find the character from the alphanumerics
+  const charPart = input.replace(/[0-9.,]/g, '');
+  // find the value from the si list
+  const siVal = si.find((s) => s.symbol === charPart);
+  // unfortunately in the case of shorthand parseFloat has to be used
+  // as BN library does not support decimals
+  const numbered = parseFloat(input.replace(/[,]/g, '.'));
+  // get only the numeric parts of input
+  return siVal ? new BN(numbered * siVal.value) : null;
```
tomusdrw, line 63: The multiplication here can still overflow the JS type size unfortunately. I think we will need to complicate this a bit more to make it fully fool proof. I'd suggest something like:
1. Detect if there are any `,` and `.`
2. If not - parse using BN instead of `parseFloat` and perform multiplication on BN.
3.  Otherwise:
4. Try to `parseFloat` and compare `parseFloat(input).toString() === input` to avoid trimming any data - if that's not correct then display error about unsupported notation. (Note that the `toString` will turn to scientific notation at some point but IMHO it's fine).

Basically the premise for this is that we either want to parse whatever the user types exactly as intended or display an error. Note that decimals are most likely going to be used for small values. Like I don't expect anyone to write something like `10000000000000000.1 M` or `1.00000000000000000001 E`, but if they do we should make it clear that it's not supported and they need to use non-float notation for this.

src/util/tests/evalUnits.test.ts
```diff
@@ -18,77 +18,79 @@ import { evalUnits, EvalMessages } from '../evalUnits';
...
     const [actualValue, msg] = evalUnits('good23');
     expect(actualValue).toBeFalsy;
     expect(msg).toBe(EvalMessages.GIBBERISH);
   });
 
   it('Should accept as input double decimal symbols (1,23.445k) and return error message', () => {
-    const [actualValue, msg] = evalUnits('1,23.445k');
-    expect(actualValue).toBeFalsy;
+    const [, msg] = evalUnits('1,23.445k');
+    // expect(actualValue).toBeFalsy;
```
tomusdrw, line 88: .

tomusdrw (approved):

src/hooks/transactions/useTransactionType.ts
hbulgarini: Sorry , i'm not following you ... could you please elaborate more the restriction for moving this logic to the reducer?

wirednkod: Each case of the switch is dependent on the input that arrives in the useTransactionType.
Based on the type (Transfer, Call, Remark), weightInput and input (string) actions are made.
By altering the ones you mention above the `TransactionActionTypes.SET_TRANSFER_AMOUNT` actions are irrelevant to the incoming input and now solely depend to the transferAmount.

You probably have a point here - transferring the above inside the clever dispatcher - but I would prefer to implement that in a different issue, and keep this only for the bug fixing. Objections?

hbulgarini: No objections! is fine to restrict the scope so PR does not get huge.

I think this can be included in https://github.com/paritytech/parity-bridges-ui/issues/195

src/util/evalUnits.ts
wirednkod: Issue was actually crazy:
```
  // This may seem crazy but 1e21 and 1e24 were actually passed to BN like that and making it break.
  // (while 1e18 is transforming to format 1000000000)
  ```

File: src/components/Transfer.tsx [hunk 11]
```diff
@@ -47,54 +50,60 @@ const useStyles = makeStyles((theme) => ({
...
     estimatedFee &&
-      actualInput &&
-      setAmountNotCorrect(new BN(balance.free).sub(new BN(actualInput).add(new BN(estimatedFee))).toNumber() < 0);
-  }, [actualInput, estimatedFee, balance, isApiReady]);
+      transferAmount &&
+      setAmountNotCorrect(
+        new BN(balance.free)
+          // TODO (nik): EstimatedFee should align with base unit token
+          // (see github issue https://github.com/paritytech/parity-bridges-ui/issues/210)
+          .sub(new BN(parseFloat(transferAmount || '0') * 10 ** api.registry.chainDecimals[0]))
```
tomusdrw, line 94: I don't like that we have `transfer` calculation in multiple places. This should be done only once and should be part of the data model imho. Why can't we move that to the reducer?

src/util/evalUnits.ts
```diff
@@ -29,8 +33,10 @@ const si = [
   { value: 1e12, symbol: 'T' },
   { value: 1e15, symbol: 'P' },
   { value: 1e18, symbol: 'E' },
-  { value: 1e21, symbol: 'Z' },
-  { value: 1e24, symbol: 'Y' }
+  // This may seem crazy but 1e21 and 1e24 were actually passed to BN like that and making it break
+  // (while 1e18 is transforming to format seen below)
+  { value: 1000000000000, symbol: 'Z' },
+  { value: 1000000000000000, symbol: 'Y' }
```
tomusdrw, line 39: Hmm, sorry but I'm pretty sure this is plain wrong. This has way too little zeros (that number is less than `1e18`) and should be left out as `1e21`.  I don't understand the reason for changing it in the first place.

src/hooks/transactions/useTransactionType.ts
```diff
@@ -69,12 +70,15 @@ export default function useTransactionType({ input, type, weightInput }: Props):
             break;
           case TransactionTypes.TRANSFER:
             if (receiverAddress) {
-              call = (await targetApi.tx.balances.transfer(receiverAddress, input)).toU8a();
+              const planck = 10 ** targetApi.registry.chainDecimals[0];
+              const [resInp] = evalUnits((parseFloat(input) * planck).toString());
+              const transferInput = resInp?.toString() || '0';
+              call = (await targetApi.tx.balances.transfer(receiverAddress, transferInput)).toU8a();
```
tomusdrw, line 76: Why do we calculate it here again? There should be only one place where this is performed.

src/util/evalUnits.ts
```diff
@@ -53,19 +81,26 @@ export enum EvalMessages {
...
-    return siVal ? [numericPart * siVal.value, EvalMessages.SUCCESS] : [null, EvalMessages.SYMBOL_ERROR];
-  } else {
+    const result = parseFloat(input.replace(/[,]/g, '.'));
+    return [result, EvalMessages.SUCCESS];
+  }
+  if (alphaInts.test(input) || alphaFloats.test(input)) {
+    const containDecimal = !input.replace(/[,]/g, '.').includes('.');
+    return transformShorthandToBN(input, containDecimal);
+  }
+  if (new BN(input).gte(new BN(Number.MAX_SAFE_INTEGER))) {
```
tomusdrw, line 96: What's the purpose of thi scheck? The point of using `BN` is to exactly support numbers larger than `MAX_SAFE_INTEGER`.

src/util/evalUnits.ts
```diff
@@ -53,19 +81,26 @@ export enum EvalMessages {
...
-  } else if (alphaInts.test(input) || alphaFloats.test(input)) {
-    const numericPart = parseFloat(input.replace(/[,]/g, '.'));
-    const charPart = input.replace(/[0-9.,]/g, '');
-    const siVal = si.find((s) => s.symbol === charPart);
-    return siVal ? [numericPart * siVal.value, EvalMessages.SUCCESS] : [null, EvalMessages.SYMBOL_ERROR];
-  } else {
+    const result = parseFloat(input.replace(/[,]/g, '.'));
+    return [result, EvalMessages.SUCCESS];
+  }
+  if (alphaInts.test(input) || alphaFloats.test(input)) {
```
tomusdrw, line 92: Can you give some examples that don't pass this test and requires the fallback below?

src/util/evalUnits.ts
```diff
@@ -53,19 +81,26 @@ export enum EvalMessages {
...
   if (floats.test(input) || ints.test(input)) {
-    return [parseFloat(input.replace(/[,]/g, '.')), EvalMessages.SUCCESS];
-  } else if (alphaInts.test(input) || alphaFloats.test(input)) {
-    const numericPart = parseFloat(input.replace(/[,]/g, '.'));
-    const charPart = input.replace(/[0-9.,]/g, '');
-    const siVal = si.find((s) => s.symbol === charPart);
-    return siVal ? [numericPart * siVal.value, EvalMessages.SUCCESS] : [null, EvalMessages.SYMBOL_ERROR];
-  } else {
+    const result = parseFloat(input.replace(/[,]/g, '.'));
+    return [result, EvalMessages.SUCCESS];
```
tomusdrw, line 90: What if it's `0.0`? We should rather return `EvalMessages.ZERO`
What if it's negative?

src/util/evalUnits.ts
```diff
@@ -40,11 +46,33 @@ const alphaInts = /^[0-9]*[a-zA-Z]{1}$/;
...
   SYMBOL_ERROR = 'Provided symbol is not correct',
   GENERAL_ERROR = 'Check your input. Something went wrong'
 }
 
+const transformShorthandToBN = (input: string, multiplyWithBN?: boolean): [BN | null, EvalMessages] => {
+  // find the character from the alphanumerics
+  const charPart = input.replace(/[0-9.,]/g, '');
+  // find the value from the si list
+  const siVal = si.find((s) => s.symbol === charPart);
+  const numberPart = input.replace(/[a-zA-Z]/g, '');
```
tomusdrw, line 61: Isn't the number part simply everything except for `charPart`?

src/util/evalUnits.ts
```diff
@@ -53,19 +81,26 @@ export enum EvalMessages {
  * the first is the actual calculated number (or null if none) while
  * the second is the message that should appear in case of error
  */
-export function evalUnits(input: string): [number | null, string] {
+export function evalUnits(input: string): [BN | number | null, string] {
```
tomusdrw, line 84: This function has become extremely complex and I don't think I fully get why. We should simply be always returning `BN`.

Take a look at my proposal:
```js
function evalUnits(input: string) -> [BN | null, string | null] {
   const symbol = input.replace(/[0-9,. ]/g, '');
   const siVal = si.find(s => s.symbol === symbol);
   if (symbol && !siVal) {
      return [ null, 'Invalid symbol' ];
   }

   const numberStr = input.replace(symbol, '').replace(',', '.);
   const asFloat = parseFloat(numberStr);
   if (asFloat.toString() === numberStr) {
      // now that's the tricky part - how do we convert a float into a BN?
      // let's move the comma as far as we can to the right :)
      const [decPart, fracPart] = numberStr.split('.');
      const decimals = fracPart.length;
      const exp = new BN(10).pow(new BN(decimals));
      const number = new BN(decPart).mul(exp).add(new BN(fracPart));
      const withSi = number.mul(siVal.value).div(exp);
      return [ withSi, null ];
   }

   // we can't parse the float correctly - perhaps it's above the precision. Let's simply try BN.
   try { 
      const bn = new BN(numberStr);
      return [ bn.mul(siVal.value), null];
   } catch (e) {
      return [null, `Invalid number: ${e}`]
   }
}
```

Note that we should keep `siVal.value` as `BN` already. Best to do `new BN(10).pow(new BN(21))` (instead of `1e21`) because parsing large floats fails as you've noticed.

tomusdrw (changes requested):

src/components/Transfer.tsx
hbulgarini: @wirednkod , tomek has the same comment as i did. 

Contrary to what agreed yesterday, i think this PR should include the logic being performed at the reducer level and not spread in different locations as we have now.

hbulgarini: Also if this logic is moved to the reducer, we can get rid of this useEffect.

hbulgarini (changes requested):

src/util/evalUnits.ts
wirednkod: Yeah you are right on that. My bad. Will convert all to `new BN().pow(new BN())` as you propose [here](https://github.com/paritytech/parity-bridges-ui/pull/204#discussion_r662406151).
Makes much more sense

src/util/evalUnits.ts
wirednkod: Was trying to make the function more details with specific errors and not generic (e.g. you dont separate Negative errors with e.g. zero) - but I get the point that it became very complicated for no actual reason.

src/util/evalUnits.ts
wirednkod: Fix negativity with regex validation. User will not be able to add anymore - symbol

src/util/evalUnits.ts
wirednkod: Code should never reach there. It was either a `"never reachable" return` or add to return types `undefined` in order to satisfy TSlint.
I will alter it and leave last option the ZERO one.

src/util/evalUnits.ts
wirednkod: This COULD be fixed if the final outcome would ALWAYS be multiplied with the chain.decimals (which I will give it a try)

src/util/evalUnits.ts
```diff
@@ -53,19 +52,37 @@ export enum EvalMessages {
...
+  const symbol = input.replace(/[0-9.,]/g, '');
+  // find the value from the si list
+  const siVal = si.find((s) => s.symbol === symbol);
+  const numberStr = input.replace(symbol, '').replace(',', '.');
+  let numeric: BN = new BN(0);
+  if (siVal) {
+    let action: number | BN;
+    const containDecimal = input.replace(/[,]/g, '.').includes('.');
+    if (containDecimal) {
+      action = parseFloat(numberStr.replace(/[,]/g, '.')) * 10 ** chainDecimals;
```
tomusdrw, line 71: The point of my proposal (and comparing parsed float with the input) was to avoid parsing user inputs incorrectly. Your approach can be still:
1. silently lossy (`parseFloat('100000000000000000.1') == 100000000000000000`)
2. or can end up failing to be converted to `BN` (try `1000` with `18` decimals: `(1000.0 * 10**18).toString ==  '1e+21'`).

src/util/evalUnits.ts
```diff
@@ -53,19 +52,37 @@ export enum EvalMessages {
...
+    } else {
+      action = new BN(numberStr).mul(new BN(10).pow(new BN(chainDecimals)));
+    }
+    numeric = new BN(action).mul(siVal.value);
+
+    // Temporary workaround so that UI will not break
+    if (numeric.bitLength() > 64) {
```
tomusdrw, line 78: Why is this needed? We do support chains with `128` bits of balance :)

src/util/evalUnits.ts
```diff
@@ -13,36 +13,35 @@
...
-  { value: 1, symbol: '' },
-  { value: 1e3, symbol: 'k' },
-  { value: 1e6, symbol: 'M' },
-  { value: 1e9, symbol: 'G' },
-  { value: 1e12, symbol: 'T' },
-  { value: 1e15, symbol: 'P' },
-  { value: 1e18, symbol: 'E' },
-  { value: 1e21, symbol: 'Z' },
-  { value: 1e24, symbol: 'Y' }
+  { value: new BN(1), symbol: '' },
```
tomusdrw, line 22: Hmm, I think it would be cool to support milli units. I know it's not possible to store in `BN` type, but perhaps we could have something like: `{ negExp: true, exp: 3 }` and convert to BN only after we multiply by decimals (in case of `negExp` you divide by `new BN(10).pow(new BN(exp))` in case of `negExp == false` you multiply).

tomusdrw (commented):
Still some issues. Maybe we can work on the suite of tests first, so that we agree on what exactly should be supported?

src/util/evalUnits.ts
wirednkod: as said [here](https://matrix.to/#/!QDITCUQOfCSGTtPxkj:matrix.parity.io/$brl9NG87YDEdskWPyvSivDBrxPuq3Y4rc0hnpO_czPg?via=matrix.parity.io):

```
Unhandled Rejection (Error): createType(Call):: Call: failed decoding balances.transfer:: Struct: failed on args: {"dest":"LookupSource","value":"Compact<Balance>"}:: Struct: failed on value: Compact<Balance>:: Balance: Input too large. Found input with 80 bits, expected 64
```
but this time this happen at the point where polkadot API is called:
```
call = (await targetApi.tx.balances.transfer(receiverAddress, transferAmount || new BN(0))).toU8a();
```
based on numeric.bitLength() the MAX token shorthand we can get is G.
If API is not called (not select for example something from the dropdown) everything works fine.

tomusdrw: But this stuff depends on chain-specific parameters (like `Balance` type size and number of decimals). This restriction seems artificial.
I'm fine doing something like:
```
function isAmountSupported(parsedBalance: BN): bool {
  try {
      api.createType('Balance', parsedBalance);
      return true;
  } catch (e) {
      return false;
  }
```

to check if given balance is goign to be supported by the chain or not.

src/components/Transfer.tsx
```diff
@@ -47,59 +49,60 @@ const useStyles = makeStyles((theme) => ({
...
-
-  // To extract estimated fee logic to specific component. Issue #171
   useEffect((): void => {
     estimatedFee &&
-      actualInput &&
-      setAmountNotCorrect(new BN(balance.free).sub(new BN(actualInput).add(new BN(estimatedFee))).toNumber() < 0);
-  }, [actualInput, estimatedFee, balance, isApiReady]);
+      transferAmount &&
+      setAmountNotCorrect(new BN(balance.free).sub(transferAmount).add(new BN(estimatedFee)).isNeg());
+  }, [transferAmount, estimatedFee, api.registry.chainDecimals, balance.free]);
```
hbulgarini, line 92: can't we move this to the reducer as well? if we can achieve this we would be pretty close to the presentational component approach we are looking for components.

Also can't this message be included in the `helperText` ?

hbulgarini (changes requested):

src/util/evalUnits.ts
wirednkod: I have implemented this as per your suggestion with a twist - meaning that first I multiply (in every case) with chainDecimals, and then div (in case of using neg) with the siVal.value or (in case of normal) multiply with the siVal.value.
There is still though a chance that there will be loss on conversion.
For example if the user types in `0.0000035m` that will actually send `3` or in another scenario if user types `1000000.3f` then this will send `1`.

What can be done in order to "catch" this is to create a string with calculation in floats and BN, compare the outcome and if it is the same, then send it along (but then again - we may gave conversion issues with the floats - right?)

What do you thing @tomusdrw ?

hbulgarini (approved):
I think this PR looks quite good enough and a lot of feedback was provided/addressed already.

I would like to merge this one and keep any other enhancements in following PR.

src/util/evalUnits.ts
```diff
@@ -53,19 +63,51 @@ export enum EvalMessages {
...
-export function evalUnits(input: string): [number | null, string] {
-  if (!floats.test(input) && !ints.test(input) && !alphaInts.test(input) && !alphaFloats.test(input)) {
+export function evalUnits(input: string, chainDecimals: number): [BN | null, string] {
+  //sanitize input to remove + char if exists
+  input = input && input.replace('+', '');
+  if (
+    !new RegExp(floats).test(input) &&
+    !new RegExp(ints).test(input) &&
+    !new RegExp(alphaInts).test(input) &&
+    !new RegExp(alphaFloats).test(input)
```
tomusdrw, line 73: Instead of creating a new `RegExp` every time, you could just make the consts above the function a `RegExp` right away.

src/util/evalUnits.ts
```diff
@@ -53,19 +63,51 @@ export enum EvalMessages {
...
+      } else {
+        numeric = new BN(new BN(numberStr).mul(new BN(10).pow(new BN(chainDecimals)))).div(siVal.value);
+      }
+    }
+    if (numeric.eq(new BN(0))) {
+      return [null, EvalMessages.ZERO];
+    }
+    return [numeric, EvalMessages.SUCCESS];
   }
+  return [null, EvalMessages.SYMBOL_ERROR];
```
tomusdrw, line 112: I find this much easier to follow and it avoids a lot of repetition.
 
```suggestion
  if (!siVal) {
    return [null, EvalMessages.SYMBOL_ERROR];
  }
  const decimalsBn = new BN(10).pow(new BN(chainDecimals));
  const containDecimal = numberStr.includes('.');
  if (containDecimal) {
      const [decPart, fracPart] = numberStr.split('.');
      const fracDecimals = fracPart.length;
      const fracExp = new BN(10).pow(new BN(fracDecimals));
      numeric = new BN(decPart).mul(exp).add(new BN(fracPart)).mul(decimalsBn).div(fracExp);
    } else {
      numeric = new BN(numberStr).mul(decimalsBn);
    }
    
    // [ToDr] I honestly hate that logic of detecting the symbol by lower/uppercase, but can live with it and won't complain any more ;)
    const isMulSymbol = symbol.toUpperCase() === symbol || symbol === 'k';
    if (isMulSymbol) {
        numeric = numeric.mul(siVal.value);
    } else {
        numeric = numeric.div(siVal.value);
    }
    
    if (numeric.eq(new BN(0))) {
      return [null, EvalMessages.ZERO];
    }
    
    return [numeric, EvalMessages.SUCCESS];
  }

```

src/util/evalUnits.ts
tomusdrw: I think the cases you've mentioned are fine since that's simply below what is possible to be sent (i.e. it's not our mistake of parsing that incorrectly - it's just the user is trying to send a value that is impossible to send).

I think with payload/amount confirmation it should be fine to ignore that edge case. If you really want to, we could also display a warning by doing something like:
```js
const number = ...;
if (number !== number.div(siValue.value).mul(siValue.value)) {
  return "Warning: truncated some of the value to transfer because it's below one planck." 
}
```

tomusdrw (changes requested):
@hbulgarini There is still incorrect behavior - I'm fine leaving smaller improvements for follow up PRs, but I don't think we should be merging something that is known to be broken.

src/components/Transfer.tsx
```diff
@@ -47,69 +48,64 @@ const useStyles = makeStyles((theme) => ({
...
           }}
         />
       </Box>
       <Receiver />
-      <ButtonSubmit disabled={isButtonDisabled() || amountNotCorrect} onClick={sendLaneMessage}>
+      <ButtonSubmit disabled={isButtonDisabled() || !!transferAmountError} onClick={sendLaneMessage}>
         Send bridge transfer from {sourceChainDetails.chain} to {targetChainDetails.chain}
       </ButtonSubmit>
-      {amountNotCorrect ? (
+      {transferAmountError === 'Not enough funds for this transaction.' ? (
```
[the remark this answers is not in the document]
hbulgarini: oh yes .. he has a good point that i missed. This should be moved to a constant on `transactionTypes.ts`

src/reducers/transactionReducer.ts
```diff
@@ -165,6 +167,23 @@ export default function transactionReducer(state: TransactionState, action: Tran
...
+    case TransactionActionTypes.SET_TRANSFER_AMOUNT: {
+      const { transferAmount, balance, chainDecimals } = action.payload;
+      let [actualValue, message] = evalUnits(transferAmount, chainDecimals);
+      if (
+        state.estimatedFee &&
+        actualValue &&
+        new BN(balance).sub(actualValue).add(new BN(state.estimatedFee)).isNeg()
+      ) {
+        actualValue = null;
+        message = 'Not enough funds for this transaction.';
```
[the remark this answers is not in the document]
hbulgarini: I recommended to move this to the reducer in order to try to avoid logic on the component level.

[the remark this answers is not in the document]
tomusdrw: Well, then it's broken because balance is not being updated (not part of the reducer) and if we want to do it properly we will again run into a situation where displaying a particular component is necessary for the reducer logic to work correctly (cause that component will have `useEffect` that will keep updating balance in the *global* reducer).

src/components/Transfer.tsx
```diff
@@ -47,69 +48,64 @@ const useStyles = makeStyles((theme) => ({
...
-      setActualInput(actualValue && actualValue * planck);
-    }
-    setTransferInput(event.target.value);
+    const actualValue = event.target.value;
+    setInput(actualValue);
+    dispatchTransaction(
+      TransactionActionCreators.setTransferAmount(
+        actualValue !== null ? actualValue?.toString() : '',
+        balance.free,
+        api.registry.chainDecimals[0]
```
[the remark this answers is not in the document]
wirednkod: This was inside a useEffect and I was asked to move it into the reducer (where we cannot move the hooks). Should I revert it back now to the useEffect? Am I missing something?

src/util/evalUnits.ts
```diff
@@ -13,33 +13,43 @@
...
+  .join(', ')
+  .replace(', ,', ',');
+const floats = '^[+]?[0-9]*[.,]{1}[0-9]*$';
+const ints = '^[+]?[0-9]+$';
+const alphaFloats = '^[+]?[0-9]*[.,]{1}[0-9]*[' + allowedSymbols + ']{1}$';
+const alphaInts = '^[+]?[0-9]*[' + allowedSymbols + ']{1}$';
 
 export enum EvalMessages {
   GIBBERISH = 'Input is not correct. Use numbers, floats or expression (e.g. 1k, 1.3m)',
+  ZERO = 'You cannot send 0 funds',
```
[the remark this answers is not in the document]
wirednkod: Technically you can yes - but is there a scenario that we should allow user to send 0 funds? should I remove this check?

src/util/evalUnits.ts
wirednkod: This code will not work. In addition I think it is better to compare on an existing entity than introducing a new one just to compare on it - but I'll proceed with that even though I do not agree (and you should always complain about what you believe is correct)

src/util/evalUnits.ts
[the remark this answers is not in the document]
tomusdrw: No, I was just commenting on the quite "strong" statement that you are making here. It's probably possible to do a PHD on displaying error messages to the user but I usually tend to be more in favour of soft, positive and helpful language instead of being strict and technical. I don't suggest any change, cause it would be out of context for this one PR - we tend to have very strict and technical error messages anyway, but for instance I'd be for something more like:
```js
GIBBERISH = "We can't understand the amount. Try typing just numbers with unit prefixes (e.g. 1k, 1.3m, etc)"
ZERO = "Seems you are trying to send 0, please check the amount."
SYMBOL_ERROR = "We don't know the symbol '{}', try using one of '{}'."
GENERAL_ERROR = "You've run into an unusual situation that we didn't handle correctly. Please consider reporting this as a bug if you think it should work". // This one is actually not used in the new code, so we should get rid of it (or can do it in `try-catch`).
```

src/util/evalUnits.ts
tomusdrw: > This code will not work

Care to elaborate?

> In addition I think it is better to compare on an existing entity than introducing a new one just to compare on it

Could you give an example? I don't think I get it.

src/components/Transfer.tsx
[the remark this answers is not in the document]
hbulgarini: i have to say this was my bad... i didn't have into account  the balance change event catched by the useEffect and as you mentioned moving that to the reducer is missing that part. I think we should revert that change sorry.

hbulgarini:
> m fine leaving smaller improvements for follow up PRs, but I don't think we should be merging something that is known to be broken.

At the time i put that comment in the approval i was not aware about the issue you find on the useEffect / reducer thing. As it is a bug introduced by my proposal , i requested @wirednkod  to rollback the chage.

hbulgarini (approved):

src/util/evalUnits.ts
wirednkod: > Care to elaborate?

Sure:
a) There is an `exp` leftover that will break the code (Im sure about that)
b)  you reverted `div(fracExp)` with the `mul(siVal.value)` which will lead to different results (we must FIRST multiply with chain decimals and THEN revert the powed frac decimals or else edge scenarios will not work. (Im not 100% sure about that)

>Could you give an example? I don't think I get it.

The symbols we KNOW that are Mul are all capitalcase except 1 (k).
So instead of adding 1 more variable to check against (isMil or isMul) one can check against standard expected values like the check I did before. As soon as the need (at least so far) does not require this extra new boolean to be used across many places, the check against capitalization comparison of self (and k) would be enough I think.

tomusdrw: > a) There is an exp leftover that will break the code (Im sure about that)

Sure, it was just a suggestion, I obviously didn't check the code just suggested something in Github editor.

> b) you reverted div(fracExp) with the mul(siVal.value) which will lead to different results (we must FIRST multiply with chain decimals and THEN revert the powed frac decimals or else edge scenarios will not work. (Im not 100% sure about that)

We still multiple by decimals and then divide by powered frac, it's ust we don't multiply/divide by `siVal` first because I don't think it's needed (yes, we loose cases where someone writes `0.00<chain-decimals-zeros>1 <largeprefix>`, but as mentioned in other comment it's not something we expected people to write. A limit of `chain-decimals` is a sane one).

> The symbols we KNOW that are Mul are all capitalcase except 1 (k).

This one is about code quality / clarity of intent. Since you've split the place where `siValues` are defined and the place where we have the logic that checks that you create an implicit assumption about that fact. Someone who would add a new unit (let's say `hecto`) will not see that the way we detect mul/div is implicitly based on the assumption about units existing during that particular time. Every kind of unclear assumption like that is not bad per-se, but it's not necessarily future proof - for some cases it makes sense and for some it will bite us in the future. Based on my experience things in the codebase change so frequently that I always try to avoid making any assumptions of this kind and if there are any I write them down in the form of tests or at least documentation.

wirednkod: Makes sense. Just when u propose code like that, github gives option of merging, and im always tempted to cllick that merge button. Are you ok with the latest changes? I pretty much followed what u proposed i think

tomusdrw (approved):

tomusdrw:
@wirednkod thanks a lot for bearing with me :) Looks awesome now. :rocket: :moon:

Issue #209, opened 2021-06-29
Optimize estimatedFee process to not calculate multiple times.


hbulgarini:
There is an inefficient and slow process that every time the input field receives a new value, the calculation process starts and it takes some time until is finished. 
The expected behavior should be that estimated Fee process should start when the element gets the onBlur event and not on the OnChange.

This bug depends on #195.

PR #211, opened 2021-06-30
Optimize api calls based on inputs.

hbulgarini:
Related with #209 

This PR optimizes the number of call to the API adding the onBlur event to the inputs.

EDIT: onBlur callback was dismissed and the optimization was implemented around debounce.


src/components/CustomCall.tsx
```diff
@@ -46,42 +50,61 @@ const CustomCall = () => {
...
   });
-  const onChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    decodePayload(event.target.value);
-    setCustomCallInput(event.target.value);
-  };
+  const onChange = useCallback((event: React.ChangeEvent<HTMLInputElement>) => {
+    setActualCustomCallInput(event.target.value);
+  }, []);
+
+  const decodePayload = useCallback(
```
tomusdrw, line 57: Why is it extracted and not simply part of `onBlur`?

tomusdrw (commented):
I thought about this a bit more and I think it might be pretty annoying to not see changes right away when you type.
Like what would happen if I do the following sequence:
1. Type 5 in the `amount`
2. Move focus to `recipient`
3. Go back to `amount` and add `000` at the end.
4. Hit "Send"

AFAICT the transaction will be sent with old `fee` computed when `5` was in the amount, cause there won't be enough time for the `onBlur` callback to be completed.
Another option is that clikcing "Send" would do nothing, cause `onBlur` we would start calculations and reset the `fee` to null, which would cause the button to be disabled for a brief period and the transaction would not be sent.

In any case this will create an inferior user experience. I think the right way to handle that is to use debounce and make it clear when "Send" button can be clicked (i.e. the most recent fee has been calculated). CC @goldsteinsveta

src/components/Transfer.tsx
```diff
@@ -61,6 +61,14 @@ function Transfer() {
   const { api, isApiReady } = sourceChainDetails.apiConnection;
   const balance = useBalance(api, account?.address || '');
 
+  const transformInput = (value: string) => {
+    const [actualValue, message] = evalUnits(value || '0');
+    setHelperText(message);
+    return actualValue && actualValue * planck;
+  };
+
+  const [transferInput, actualInput, setActualInput] = useDebounceState('', null, transformInput);
```
hbulgarini, line 70: Taking into account what @wirednkod is working on #204 where the input will be moved to the reducer logic , some of the features here might not be needed any more. The idea is to not use the `transformInput` callback and assign the same value to both states so the one going to the reducer will be processed there.

src/hooks/react/useDebounceState.ts
```diff
@@ -0,0 +1,52 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
```
hbulgarini, line 1: What do you think about this implementation @tomusdrw @wirednkod ? Do you get the idea of this hook?

src/components/Remark.tsx
```diff
@@ -21,30 +21,34 @@ import { useSourceTarget } from '../contexts/SourceTargetContextProvider';
...
   const { isButtonDisabled, sendLaneMessage } = useSendMessage({
-    input: remarkInput,
+    input: remarkActualInput,
     isRunning,
     setIsRunning,
     type: TransactionTypes.REMARK
   });
   const onChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    setRemarkInput(event.target.value);
+    setRemarkInput(event.target.value || initialRemark);
```
tomusdrw, line 43: No, please don't do that. It's annoying to have the input be modified randomly. What's wrong with an empty string? Just make it a placeholder.

src/hooks/react/useDebounceState.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const useDebounceState = (
+  initialValue: string | null,
+  wait?: null | number,
+  transformCallback?: Function
+): Output => {
+  const [value, setValue] = useState(initialValue);
+  const [debounced, setDebounced] = useState(value);
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
tomusdrw, line 30: Looks wrong.

src/hooks/react/useDebounceState.ts
```diff
@@ -0,0 +1,52 @@
...
+import { useCallback, useState } from 'react';
+import debounce from 'lodash/debounce';
+
+type Output = [string | null, string | null, Function];
+
+export const useDebounceState = (
+  initialValue: string | null,
+  wait?: null | number,
+  transformCallback?: Function
+): Output => {
```
tomusdrw, line 26: With so many optional values, I'd suggest to use configuration object instead of individial parameters.

src/hooks/react/useDebounceState.ts
```diff
@@ -0,0 +1,52 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+
+import { useCallback, useState } from 'react';
+import debounce from 'lodash/debounce';
+
+type Output = [string | null, string | null, Function];
+
+export const useDebounceState = (
+  initialValue: string | null,
```
tomusdrw, line 23: This should be a generic value instead of `string`.

src/hooks/react/useDebounceState.ts
```diff
@@ -0,0 +1,52 @@
...
+  wait?: null | number,
+  transformCallback?: Function
```
tomusdrw, line 25: I think we could be more explicit with the type here if we use generic types.
```suggestion
  transformCallback?: (T) => Y
```

src/hooks/transactions/useTransactionType.ts
```diff
@@ -29,9 +29,9 @@ interface TransactionFunction {
 }
 
 interface Props {
-  input: string;
+  input: string | null;
   type: string;
-  weightInput?: string;
+  weightInput?: string | null;
```
tomusdrw, line 34: It was nicer if the types were non-nullable. We can solve that by having the `useDebouncedState` always require the initial state or moving to  generic parameters.

src/components/Transfer.tsx
tomusdrw:
```suggestion
  const [currentInput, debouncedInput, setInput] = useDebounceState('', null, transformInput);
```
I'd rename the variables to emphasise it's debounced.

src/hooks/react/useDebounceState.ts
hbulgarini: what do you mean by a configuration object? as input or output?

src/hooks/react/useDebounceState.ts
hbulgarini: It's not i'm afraid ... if i don't disable the linter on that i receive the following message:

```
React Hook useCallback received a function whose dependencies are unknown. Pass an inline function instead
```

But the problem is that i need the debounce enclosing the whole function otherwise it does not work. Any ideas how to solve this without disabling eslint?

src/components/Remark.tsx
hbulgarini: ok.

src/components/Transfer.tsx
hbulgarini: sure.

src/hooks/react/useDebounceState.ts
hbulgarini: Please check now if it looks better ... regarding the output i want to leave it as array format similar to `useState`

src/hooks/react/useDebounceState.ts
tomusdrw: Well, `useCallback` and `debounce` sounds like a common problem, I found some solution [here](https://kyleshevlin.com/debounce-and-throttle-callbacks-with-react-hooks) do you think that could work?

tomusdrw (commented):
Looks good, except for the disabled linter.

src/hooks/react/useDebounceState.ts
wirednkod: I wanted to say `use useMemo` but link of Tomasz gave that away - and it would sound like I copied it from there - so I say nothing.

src/hooks/react/useDebounceState.ts
```diff
@@ -0,0 +1,54 @@
...
+// Parity Bridges UI is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+
+import { useCallback, useState } from 'react';
+import debounce from 'lodash/debounce';
```
wirednkod, line 18: just `import { debounce } from 'lodash';` will do the job

wirednkod (approved):
lgtm - I would wait for #204 to be merged first in order to avoid the upcoming conflicts

src/hooks/react/useDebounceState.ts
hbulgarini: I think we discussed it before , but doing that you are importing the whole lodash package which is quite big.

tomusdrw: @hbulgarini @wirednkod please compare numbers not opinions. I agree with Nikos here, simple import is nicer and I find it hard to believe that the code cannot be tree-shaken efficiently. However if you can prove to me that the artifacts are *significantly* larger if we do that I'm happy to rewrite pretty much everything to the approach you propose.

src/hooks/react/useDebounceState.ts
hbulgarini: works like charm ... thanks both!

hbulgarini:
> lgtm - I would wait for #204 to be merged first in order to avoid the upcoming conflicts

Of course!

src/hooks/react/useDebounceState.ts
tomusdrw: @hbulgarini thanks! I was skeptical because the article is quite old (May 2018), and run tests myself. We only have 3 instances of `lodash` usage currently and all of them is `isEmpty`. One in `src/App.tsx` is not using the syntax you've mentioned - fixing this (as done in https://github.com/paritytech/parity-bridges-ui/pull/214) reduces the build artifacts by 23KBs. I stand corrected.

tomusdrw (approved):

src/hooks/react/useDebounceState.ts
hbulgarini: cool!

wirednkod: I know that @tomusdrw will call me "too opinionated" but _just For Our Information (including some good will)_, this is a characteristic of lodash that - afaiu - [will not longer be supported on v5 as per official docs](https://lodash.com/per-method-packages). In addition - as I do not like CRA so much and prefer other bundlers - webpack and parcel transpiles the above mentioned tree shaking absolutely perfectly while CRA does not

😃

src/components/Transfer.tsx
```diff
@@ -62,44 +61,53 @@ function Transfer() {
...
     transactionRunning && setInput('');
-  }, [transactionRunning]);
+  }, [setInput, transactionRunning]);
 
   useEffect((): void => {
     estimatedFee &&
       transferAmount &&
-      setAmountNotCorrect(new BN(balance.free).sub(transferAmount).add(new BN(estimatedFee)).isNeg());
-  }, [transferAmount, estimatedFee, api.registry.chainDecimals, balance.free]);
+      setAmountNotCorrect(new BN(balance.free).sub(new BN(transferAmount).add(new BN(estimatedFee))).toNumber() < 0);
```
tomusdrw, line 101: This should not have been changed it basically reverts all the hard work Nikos put into this in #204

hbulgarini:
@tomusdrw @wirednkod , sorry guys i would need you to review the PR again since the sync with master it extended the scope.

src/components/Transfer.tsx
hbulgarini: ouh! that came bacause a conflict resolution that got the old code.

src/components/Transfer.tsx
```diff
@@ -111,20 +119,20 @@ function Transfer() {
...
         </Alert>
-      ) : estimatedFeeLoading ? (
+      ) : estimatedFeeLoading && transferAmount && receiverAddress ? (
         <Typography variant="body1" color="secondary">
           Estimated source Fee loading...
         </Typography>
       ) : (
         <Typography variant="body1" color="secondary">
-          {receiverAddress && estimatedFee && `Estimated source Fee: ${estimatedFee}`}
+          {receiverAddress && estimatedFee && transferAmount && `Estimated source Fee: ${estimatedFee}`}
```
hbulgarini, line 135: i don't really like this validations for showing the `estimatedFee` ... i suspect merging this PR with the PR #218  will solve the need for so many checks.

src/hooks/transactions/usePayload.ts
```diff
@@ -38,27 +39,32 @@ export const usePayload = ({ call, weight }: Input) => {
...
-    };
-    // @ts-ignore
-    const payloadType = createType(sourceChain, 'OutboundPayload', payload);
-    logger.info(`OutboundPayload: ${JSON.stringify(payload)}`);
-    logger.info(`OutboundPayload.toHex(): ${payloadType.toHex()}`);
-    return payload;
-  }, [account, call, createType, sourceChain, targetApi, weight]);
+      // @ts-ignore
+      const payloadType = createType(sourceChain, 'OutboundPayload', payload);
+      logger.info(`OutboundPayload: ${JSON.stringify(payload)}`);
```
wirednkod, line 62: I would conver these to debug in order to avoid having them in production

src/contexts/TransactionContext.tsx
```diff
@@ -65,6 +69,10 @@ export function TransactionContextProvider(props: TransactionContextProviderProp
     payloadHex: null
   });
 
+  useEffect((): void => {
```
hbulgarini, line 72: @tomusdrw  can we consider this kind of effects part of the data model update you were talking about? i definitely understand the idea but in react terms i think this maybe the place where this kind of logic should reside.

hbulgarini:
@tomusdrw  is there anything pending on this PR for your review?

src/components/CustomCall.tsx
```diff
@@ -38,47 +42,50 @@ const CustomCall = () => {
...
+    (event: React.ChangeEvent<HTMLInputElement>) => {
+      const input = event.target.value;
+      try {
+        setError(null);
+        setCustomCallInput(input);
+        //@ts-ignore
+        const call = createType(targetChain, 'Call', input);
+        setDecoded(JSON.stringify(call, null, 4));
+      } catch (e) {
+        setError('Wrong call provided');
```
tomusdrw, line 60: Might be good to at least log the original error to the console.
```suggestion
      } catch (e) {
        logger.error('Wrong call', e);
        setError('Wrong call provided');
```

src/contexts/TransactionContext.tsx
tomusdrw: Yes, that's fine.

src/contexts/TransactionContext.tsx
```diff
@@ -65,6 +69,10 @@ export function TransactionContextProvider(props: TransactionContextProviderProp
...
+    account && dispatchTransaction(TransactionActionCreators.setSenderAccount(account.address));
```
tomusdrw, line 73:
```suggestion
    dispatchTransaction(TransactionActionCreators.setSenderAccount(account ? account.address : null));
```

If `account` is set back to `null` (or in general something undefined) we probably want the reducer to be updated as well.

src/hooks/transactions/usePayload.ts
tomusdrw: Why would that be an issue? How do we distinguish between "production" and "development"?

tomusdrw (approved):

src/contexts/TransactionContext.tsx
hbulgarini: at the moment this scenario is not possible since you choose an account you don't have the possibility to clean the account, but as i suppose this will be possible in the refactored Sender component this might be valid.

tomusdrw: The fact that something is not **currently** clickable (or we believe it is not) does not justify taking shortcuts in code correctness.
1. Setting it to `null` is strictly better than not doing anything.
2. The code change is trivial
3. Assumptions like that quickly get outdated.
Hence I'd be always more in favour "future proof" behaviors.

Thanks for addressing that!

hbulgarini: of course, that's why i implemented your proposal without any questioning.

PR #212, opened 2021-06-30
Reading spec_version from target api instead of hardcode the value.

### LINKED ISSUE #122: `spec_version` should be fetch from chain
Currently `spec_version` when constructing `MessagePayload` is hardcoded to `1`. Instead it should be fetch from target chain API (runtime version).

hbulgarini:
Closes #122 .

### DIFF
--- a/src/hooks/transactions/usePayload.ts
+++ b/src/hooks/transactions/usePayload.ts
@@ -34,14 +34,14 @@ export const usePayload = ({ call, weight }: Input) => {
   const {
     sourceChainDetails: { chain: sourceChain },
     targetChainDetails: {
-      apiConnection: { api: targetApi, isApiReady: isTargetApiReady }
+      apiConnection: { api: targetApi }
     }
   } = useSourceTarget();
   const { dispatchTransaction } = useUpdateTransactionContext();
   const { account } = useAccountContext();
 
   const payloadCallback = useCallback(() => {
-    if (!account || !call || !weight || !isTargetApiReady) {
+    if (!account || !call || !weight) {
       return null;
     }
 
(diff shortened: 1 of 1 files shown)


src/hooks/transactions/usePayload.ts
```diff
@@ -32,30 +32,33 @@ interface Input {
...
+    targetChainDetails: {
+      apiConnection: { api: targetApi, isApiReady: isTargetApiReady }
+    }
   } = useSourceTarget();
   const { dispatchTransaction } = useUpdateTransactionContext();
   const { account } = useAccountContext();
 
   const payloadCallback = useCallback(() => {
-    if (!account || !call || !weight) {
+    if (!account || !call || !weight || !isTargetApiReady) {
```
tomusdrw, line 44: Do we really need the `isTargetApiReady` check? I mean - what is `targetApi.consts.system.version.specVersion` supposed to contain if the API is not ready?

tomusdrw (approved):

src/hooks/transactions/usePayload.ts
hbulgarini: my understanding is that all child objects for the api are undefined until the api is ready.
@wirednkod  what do you think?

wirednkod: If api is not ready you ll probably get an error saying `cannot find const of undefined`

hbulgarini: cool .. then is needed. Thanks!

tomusdrw: No, not really. It will error out when trying to set up the hook, because you pass `specVersion` in the dependency list.

PR #216, opened 2021-07-06
Validate SS58 prefix length without parsing an error message

### LINKED ISSUE #115: Avoid parsing error messages from `checkAddress`
Currently to determine account prefix we parse error returned from `checkAddress`:
https://github.com/paritytech/parity-bridges-ui/blob/0d64607baab6c8d75f0bc66cc4c887258fcbd7bb/src/util/getReceiverAddress.ts#L31

Instead we should use something like `checkAddressChecksum` and simply decode the prefix.

adamzerella:
Been meaning to start working with Parity for awhile now, figured this issue was a good place to start! 😄 

Resolves #115


src/util/getReceiverAddress.ts
```diff
@@ -46,15 +45,14 @@ const getReceiverAddress = ({ targetChainDetails, sourceChainDetails, receiverAd
   };
 
   try {
-    const [validatedDerivedAcccount, rest] = checkAddress(receiverAddress, targetSS58Format);
+    const decodedReceiverAddress = base58Decode(receiverAddress);
+    const [isValidDerivedAcccount, , , ss58Decoded] = checkAddressChecksum(decodedReceiverAddress);
```
tomusdrw, line 49: :sob: my eyes hurt when I look at the `, , , ` to skip params, but I guess it's a common JS thing, isn't it?

Could we maybe add something like `isValidDerivedAccount, _length, _ss58Lengh, ss58Decoded` for readabiliy, i.e. some unused variables without making `eslint` yell at us?

tomusdrw (approved):
nice stuff! CC @hbulgarini

hbulgarini:
hi @adamzerella ! thanks for the taking the time to review this.

Please take into account is making the following test to fail: https://github.com/paritytech/parity-bridges-ui/runs/2998516129?check_suite_focus=true . So i assume you need to review if the output is exactly the same is it used to be.

hbulgarini (changes requested):
hi @adamzerella ! thanks for the taking the time to review this.

Please take into account is making the following test to fail: https://github.com/paritytech/parity-bridges-ui/runs/2998516129?check_suite_focus=true . So i assume you need to review if the output is exactly the same is it used to be.

src/util/getReceiverAddress.ts
adamzerella: Totally agree on these unused vars :( 

The only way to keep ESLint happy here is to at a `//eslint-disable ...` IF we want to keep the variables named but unused, otherwise we can modify the `.eslintrc` to support unused variables but...that also sucks 👎🏻

adamzerella:
> hi @adamzerella ! thanks for the taking the time to review this.
> 
> Please take into account this change is making the following test to fail: https://github.com/paritytech/parity-bridges-ui/runs/2998516129?check_suite_focus=true . So i assume you need to review if the output is exactly the same is it used to be.

I've modified the tests slightly here, don't get the wrong idea as I'm not trying to 'shut the test suite up' but I no longer think some of the test fixture states make sense, at least 2 of them no longer do

src/reducers/tests/transactionReducer.test.ts
```diff
@@ -53,18 +54,6 @@ describe('transactionReducer', () => {
       });
     });
 
-    it('should return the according transaction state for a generic address.', () => {
-      payload.unformattedReceiverAddress = '5H3ZryLmpNwrochemdVFTq9WMJW39NCo5HWFEwRtjbVtrThD';
-      const action = TransactionActionCreators.setReceiver(payload);
-      const result = transactionReducer(state, action);
-      expect(result).toEqual({
-        ...state,
-        genericReceiverAccount: '5H3ZryLmpNwrochemdVFTq9WMJW39NCo5HWFEwRtjbVtrThD',
```
adamzerella, line 62: The problem here is that the receiver address _is not_ a generic address, it's a valid address...

This function will throw a `INCORRECT_FORMAT` error given invalid address, maybe we assert that the function throws instead?

adamzerella: Hmm, on second though I may have misunderstood how these SS58 address' are decoded 🤔

src/util/getReceiverAddress.ts
```diff
@@ -46,15 +46,17 @@ const getReceiverAddress = ({ targetChainDetails, sourceChainDetails, receiverAd
   };
 
   try {
-    const [validatedDerivedAcccount, rest] = checkAddress(receiverAddress, targetSS58Format);
+    const decodedReceiverAddress = base58Decode(receiverAddress);
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [isValidDerivedAcccount, _length, _ss58DecodedLength, ss58Decoded] = checkAddressChecksum(
```
hbulgarini, line 51: i much prefer your initial approach of empty values rather than escaping this that shows the warning in the console.

hbulgarini (changes requested):
change looks good to me even if  `derivedReceiverAccount` state has a different value.

Please revert the warning as mentioned  in my comment and we are good to go!

src/util/getReceiverAddress.ts
adamzerella: Haha me too! Wasn't sure what to do with the comment earlier, is just one of those things that sucks in JS I suppose 😄

hbulgarini (approved):
lgtm! thanks!

src/util/getReceiverAddress.ts
```diff
@@ -46,15 +46,14 @@ const getReceiverAddress = ({ targetChainDetails, sourceChainDetails, receiverAd
...
-      return { address: receiverAddress, formatFound: targetChain };
-    }
+    const formatFound = getChainBySS58Prefix(ss58Decoded) || ss58Decoded;
 
-    const parts = rest?.split(',');
-    const prefix = parts![2].split(' ');
-    const formatFound = getChainBySS58Prefix(prefix[2]) || prefix[2];
+    if (isValidDerivedAcccount) {
+      return { address: receiverAddress, formatFound };
+    }
```
tomusdrw, line 56: I think this is where the issue with tests is. Note that previously `validatedDerivedAccount` as only `true` if `targetSS58Format` was matching the decoded format.
Currently I think we will get `true` if the address was decoded successfully, but we should compare the `ss58Decoded` format with the expected one and re-format if needed.

tomusdrw (changes requested):

hbulgarini (changes requested):
Tomek's comment makes sense. Even if the process works  with the change (checked on my own already step by step ) it would be nice to have the same expected output as before.

src/util/getReceiverAddress.ts
hbulgarini: @adamzerella how's everything going with this comment? do you need some help to address it? I think we are very close :)

adamzerella: Makes sense to me guys! Just trying to find the time to update the PR, aiming to push something later this week 🤞🏻

tomusdrw (approved):
Looks great, thanks!

hbulgarini (approved):
great!!! thank you Alan!

Issue #48, opened 2021-04-14
Put icons on all app


hbulgarini:

hbulgarini:
Code reference where the icons should be defined per each chain: https://github.com/paritytech/parity-bridges-ui/blob/master/src/configs/substrateProviders.ts

goldsteinsveta:
I close the issue as "all app" uses material UI icons. Unfortunately the linked config is 404, but as I understand this is about network logos. There's icon/ligatures font: https://github.com/goldsteinsveta/web3-regular that I'd use for chains with logos. Chains that are currently supported do not have associated logos at all.

PR #218, opened 2021-07-07
Fix estimated fee component and transform

### LINKED ISSUE #210: Estimated Fee is not calculated on base unit
When calculated estimated fee does not return on MLAU unit: (e.g. 0.00002) but multiplied by chain decimals.
That is an inconsistency with the input that user adds (Its in base unit) so user sees 1 MLAU as input but the estimated fee appears: `Estimated source Fee: 275002585`

wirednkod:
Fixes #210 #171


src/components/EstimatedFee.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const srcChainDecimals = sourceChainDetails.apiConnection.api.registry.chainDecimals[0];
+  const { chainTokens } = targetChainDetails.apiConnection.api.registry;
+
+  const [amount, setAmount] = useState<string>('0');
+
+  useEffect(() => {
+    !estimatedFeeLoading && setAmount(transformToBaseUnit(estimatedFee || '0', srcChainDecimals));
+  }, [estimatedFee, estimatedFeeLoading, srcChainDecimals]);
+
+  return (
```
hbulgarini, line 49: Why are we not uncluding the loading part in this component?

```
      {estimatedFeeLoading ? (
          Estimated source Fee loading...
        </Typography>
      ) : (
        <EstimatedFee />
      )}
```

I see this is quite the same over the other actions.

hbulgarini (changes requested):

src/components/EstimatedFee.tsx
wirednkod: Good catch - eluded my attention as Loading is not working at the moment. Will fix

src/components/EstimatedFee.tsx
```diff
@@ -49,9 +49,9 @@ export const EstimatedFee = (): React.ReactElement => {
   return (
     <div>
       <Typography variant="body1" color="secondary">
-        Estimated source Fee:
+        {`Estimated source Fee ${estimatedFeeLoading ? 'Loading' : ''}`}
```
tomusdrw, line 52:
```suggestion
        {`Estimated ${chain} Fee: ${estimatedFeeLoading ? '...' : ''}`}
```
Can we replace this to say the chain name (uppercase), have a double colon so that it's clear that it's a label for some value?

I'd also get rid of `Loading`. It's not really "loading" anything, it's being calculated and we are waiting for a response. Adding a word there is extra mental capacity imho, I'd just replace it with dots. This is totally my preference though so I'm not super strong on this.

src/components/EstimatedFee.tsx
```diff
@@ -49,9 +49,9 @@ export const EstimatedFee = (): React.ReactElement => {
...
         <span className={`${estimatedFeeLoading ? classes.blurred : ''} ${classes.amount}`}>{amount}</span>
-        {chainTokens}
+        {!estimatedFeeLoading && chainTokens}
```
tomusdrw, line 54: If we switch to `...` I'd also always display tokens here (unconditionally).

src/util/evalUnits.ts
```diff
@@ -89,3 +89,8 @@ export function evalUnits(input: string, chainDecimals: number): [BN | null, str
   }
   return [numeric, EvalMessages.SUCCESS];
 }
+
+export function transformToBaseUnit(estFee: string, chainDecimals: number): string {
+  const est = parseInt(estFee) / 10 ** chainDecimals;
+  return est.toString();
```
tomusdrw, line 95: Any chance we can do it using `BN` library? Or manually? It's basically about padding with zeros and potentially adding `0.` in front. The reason I ask for this is because if `chainDecimals` are large we will again run into float issues and weird  values being displayed.

Example:
```
(235 / 10 ** 9).toString() == '2.35e-7'
```

tomusdrw (approved):

src/util/evalUnits.ts
wirednkod: Im not sure that you will ever have 235 fee... In addition is there a chance that the chain will have way more than 9 decimals?

tomusdrw: It's not about our assumptions, but what Substrate is designed for. If we can make it always correct (and not only under some opinionated conditions) with a bit of more code, let's do it.

Ethereum has 18 decimals for instance and the fees are in `Gwei` so 20 Gwei should not be displayed as `(20 * 10**9/10**18).toString() === '2e-8'`

src/components/EstimatedFee.tsx
hbulgarini: I like using three dots instead of loading ....

Regarding the chain , what it would be better is to add the corresponding chain icon instead of the chain name.

hbulgarini: > Regarding the chain , what it would be better is to add the corresponding chain icon instead of the chain name.

I'm not saying to do that in this PR but as a refresher we have this issue: #48

src/components/EstimatedFee.tsx
hbulgarini: What do you mean?  not sure i get this comment.

wirednkod: I think Tomasz means that if we replace `Loading` with `...` There is no need to have a condition on wheather the chainTokens should appear or not (and IMO he is right)

src/components/EstimatedFee.tsx
wirednkod: > I'm not saying to do that in this PR but as a refresher we have this issue: #48

I agree you should reopen that issue and lets do that there

hbulgarini (approved):
This looks good to me!

tomusdrw (changes requested):
Requesting changes due to: https://github.com/paritytech/parity-bridges-ui/pull/218#discussion_r666091675

tomusdrw (approved):

src/components/EstimatedFee.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  useEffect(() => {
+    !estimatedFeeLoading && setAmount(transformToBaseUnit(estimatedFee || '0', srcChainDecimals));
+  }, [estimatedFee, estimatedFeeLoading, srcChainDecimals]);
+
+  return (
+    <div>
+      <Typography variant="body1" color="secondary">
+        {`Estimated ${sourceChainDetails.chain} fee ${estimatedFeeLoading ? '...' : ''}`}
+        <span className={classes.amount}>
```
goldsteinsveta, line 50: Style nit: pls replace `<span>` with `<Typography variant="subtitle2" component="span">`

src/components/EstimatedFee.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { makeStyles } from '@material-ui/core/styles';
+import { Typography } from '@material-ui/core';
+import React from 'react';
+import { useEffect } from 'react';
+import { useState } from 'react';
+import { useSourceTarget } from '../contexts/SourceTargetContextProvider';
+import { useTransactionContext } from '../contexts/TransactionContext';
+import { transformToBaseUnit } from '../util/evalUnits';
+
+const useStyles = makeStyles(() => ({
```
goldsteinsveta, line 26: You don't need these styles really

src/components/EstimatedFee.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const [amount, setAmount] = useState<string>('0');
+
+  useEffect(() => {
+    !estimatedFeeLoading && setAmount(transformToBaseUnit(estimatedFee || '0', srcChainDecimals));
+  }, [estimatedFee, estimatedFeeLoading, srcChainDecimals]);
+
+  return (
+    <div>
+      <Typography variant="body1" color="secondary">
+        {`Estimated ${sourceChainDetails.chain} fee ${estimatedFeeLoading ? '...' : ''}`}
```
goldsteinsveta, line 49: I'd suggest to `${estimatedFeeLoading ? '...' : ':'}`

goldsteinsveta (approved):

PR #219, opened 2021-07-07
Fix bug that was not activating the button when receiver was updated;…

### LINKED ISSUE #217: Issue with deactivated Send button
When the steps are done in the following order the Send button remains deactivated.
a) Choose account from Sender dropdown
b) add an amount
c) add a Receiver address

Result: Button remains deactivated. In order to activate again the amount needs to be altered AFTER both addresses are added.

Screenshot from 2021-07-07 19-23-41

wirednkod:
… Fixes #217


wirednkod:
@hbulgarini Can you please verify that the line I remove was erroneously there? 
Im not sure why it was there, but I believe it should not be there

hbulgarini (approved):

PR #222, opened 2021-07-09
Refactoring Transactions state - data model

### LINKED ISSUE #159: Refactor by adding OverrideSourceTarget around TransactionStatus component.

hbulgarini:
This PR is making an important refactor in regards to transactions management state. 

Closes #159.

Once the transaction is created and updated  with the block number on the `useSendMessage` hook ( soon to be accommodated under this architecture ), then all the updates and happens on the transaction context which flow the updates through to the corresponding child components.

This change follows the agreed architecture of a the state being handled by the data model ( in react terms  transaction reducer shared through transaction context ), so presentational components are only in charge of displaying the information dictated from this data model.


src/util/transactionUtils.ts
```diff
@@ -71,3 +77,211 @@ export function getTransactionDisplayPayload({
...
+      label: 'Finalize message',
+      status: completionStatus(messageFinalizedOnTarget)
+    },
+    {
+      id: 'test-step-confirm-delivery',
+      chainType: sourceChain,
+      label: 'Confirm delivery',
+      status: completionStatus(sourceConfirmationReceived)
+    }
+  ];
```
wirednkod, line 239:
```suggestion
  const updatedSteps = [
    step(1, sourceChain, completionStatus(!!transaction.block), transaction.block),
    step(2, sourceChain, completionStatus(sourceTransactionFinalized)),
    step(3, targetChain, completionStatus(blockFinalityRelayed)),
    step(4, targetChain, completionStatus(messageDelivered), onChainCompleted(messageDelivered) && deliveryBlock),
    step(5, targetChain, completionStatus(messageFinalizedOnTarget)),
    step(6, sourceChain, completionStatus(sourceConfirmationReceived))
  ];
```

src/util/transactionUtils.ts
```diff
@@ -71,3 +77,211 @@ export function getTransactionDisplayPayload({
...
+    label: 'Finalize message',
+    status: TransactionStatusEnum.NOT_STARTED
+  },
+  {
+    id: 'test-step-confirm-delivery',
+    chainType: sourceChain,
+    label: 'Confirm delivery',
+    status: TransactionStatusEnum.NOT_STARTED
+  }
+];
```
wirednkod, line 287:
```suggestion
export const createEmptySteps = (sourceChain: string, targetChain: string) => [
  step(1, sourceChain),
  step(2, sourceChain),
  step(3, targetChain),
  step(4, targetChain),
  step(5, targetChain),
  step(6, sourceChain)
];
```

src/util/transactionUtils.ts
```diff
@@ -71,3 +77,211 @@ export function getTransactionDisplayPayload({
   transactionDisplayPayload.spec_version = spec_version;
   return { transactionDisplayPayload, payloadHex };
 }
+
```
wirednkod, line 80: You can add the following array and function that will simplify a lot the `updatedSteps` and `createEmptySteps` (see below where updateSteps and startSteps are) :

```suggestion

const steps = [
  ['include-message-block', 'Include message in block'],
  ['finalized-block', 'Finalize block'],
  ['relay-block', 'Relay block'],
  ['deliver-message-block', 'Deliver message in target block'],
  ['finalized-message', 'Finalize message'],
  ['confirm-delivery', 'Confirm delivery']
];

const step = (step: number, chainType: string, status?: TransactionStatusEnum, labelOnChain?: any) => {
  const obj = {
    id: 'test-step-' + steps[step - 1][0],
    chainType,
    label: steps[step - 1][1],
    status: status || TransactionStatusEnum.NOT_STARTED
  };
  if (!isEmpty(labelOnChain)) {
    return { ...obj, labelOnChain };
  }
  return obj;
};

```

hbulgarini: wow classy! let me play with it ... i think this will go in thanks!

src/util/transactionUtils.ts
hbulgarini: i was thinking to do something like this , but as the PR scope got challenging i decided to leave it as it is and address it in next PR.
Since you took the time to solve this i will definitely accept this. Thanks!

wirednkod (approved):

src/hooks/transactions/useTransactionsStatus.tsx
```diff
@@ -0,0 +1,112 @@
...
+    const getTransactionStatus = async () => {
+      const updatedTransactions = await Promise.all(
+        transactions.map(async (transaction: TransactionStatusType) => {
+          if (transaction.status === TransactionStatusEnum.COMPLETED || transaction.evaluating) {
+            return transaction;
+          }
+          dispatchTransaction(
+            TransactionActionCreators.updateTransactionStatus(
+              {
+                evaluating: true
```
tomusdrw, line 54: How is this related to `transaction.evaluating`?

src/hooks/transactions/useTransactionsStatus.tsx
```diff
@@ -0,0 +1,112 @@
...
+    getTransactionStatus();
+  }, [
+    apiCalls,
+    currentSourceChain,
+    dispatchMessage,
+    dispatchTransaction,
+    getValuesByChain,
+    laneId,
+    subscriptions,
+    transactions
```
tomusdrw, line 110: This will keep re-evaluating the same transactions every time we just add/change anything in the array unless it's either complete or already evaluating. Isn't there a way to be a bit more fine-grained about it?

src/hooks/transactions/useTransactionsStatus.tsx
```diff
@@ -0,0 +1,112 @@
...
+          return updatedTransaction;
+        })
+      );
+
+      if (!isEqual(transactions, updatedTransactions)) {
+        dispatchTransaction(TransactionActionCreators.updateTransactionsStatus(updatedTransactions));
+      }
+    };
+
+    getTransactionStatus();
```
tomusdrw, line 101: async errors are being swallowed here + you risk dispatching actions after the hook is unmounted (since it's in the context, I guess that's not a real issue, but we could add a comment somewhere).

src/hooks/transactions/useTransactionsStatus.tsx
```diff
@@ -0,0 +1,112 @@
...
+              )
+            );
+          }
+
+          return updatedTransaction;
+        })
+      );
+
+      if (!isEqual(transactions, updatedTransactions)) {
+        dispatchTransaction(TransactionActionCreators.updateTransactionsStatus(updatedTransactions));
```
tomusdrw, line 97: How do we ensure that the update is processed in the correct order? Since this code is running async, the results might arrive in different time, hence an older update can update the newer one - since we update all transacations in one go (I'm assuming we override) - this can cause an old transaction status being displayed.

src/contexts/TransactionContext.tsx
```diff
@@ -69,12 +70,14 @@ export function TransactionContextProvider(props: TransactionContextProviderProp
     payloadHex: null
   });
 
+  useTransactionsStatus(transactionsState.transactions, dispatchTransaction);
```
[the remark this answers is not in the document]
hbulgarini: I have moved the hook to the directory `hooks/context` so at least at dir level, there is a relation between them.
Then regarding moving the `useReducer` to specific hook i'm not 100% convinced since now we have just one hook that updates the data model, but what happens if in the future we need to add more? which one should be the one defining the reducer?

src/hooks/transactions/useTransactionsStatus.tsx
hbulgarini: I tried to address both concerns:

1) In case there is any async error, an error message will be dispatched so the toast appears on the screen.
2) If due to really edged case the component gets unmounted while the evaluation is running, i make sure to set the transaction evaluation state to false again so the next cycle can re start the process.

src/hooks/transactions/useTransactionsStatus.tsx
hbulgarini: Right. I think it was not a way to ensure the right order as you mentioned, but that was not a critical issue because the update containing the updated values would happen quite fast.

Anyway, based on your comment i redefined the evaluation process by replacing the dedicated `evaluating` state of each transaction by a global `evaluatingTransactions` state. As soon as the first evaluation starts, this state is updated and as soon as all the evaluations finishes then this variable is updated again.
With this change we make sure that the updates will come in the right order since there will not be two checks at the same time because of this check. This might decrease a little bit the performance of the check, but i prefer having the right update process.

src/hooks/transactions/useTransactionsStatus.tsx
hbulgarini: Ideally the evaluation function should be executed mostly when there is an update on the subscriptions, so i'm wondering if i can catch the previous values of subscriptions ( using `usePrevious` ) and only run the function when the values are different. If we implement this check, we would be omitting the additional execution because of the previous transaction array update. What do you think?

src/hooks/transactions/useTransactionsStatus.tsx
hbulgarini: outdated. https://github.com/paritytech/parity-bridges-ui/pull/222#discussion_r672560246

src/hooks/transactions/useTransactionsStatus.tsx
tomusdrw: But you would need to do that for each transaction separately? I don't know how much extra complexity this would introduce - I'm fine with the current code as-is, was just wondering if there aren't maybe any more fine-grained ways. Ideally we would like only the transaction that actually is affected to be re-calculated, not all of them, but this shouldn't be an issue if we keep the transaction count reasonable.

src/hooks/context/useTransactionsStatus.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+          return updatedTransaction;
+        })
+      );
+
+      if (!isEqual(transactions, updatedTransactions)) {
+        dispatchTransaction(TransactionActionCreators.updateTransactionsStatus(updatedTransactions));
+      }
+
+      dispatchTransaction(TransactionActionCreators.setEvaluatingTransactionsStatus(false));
```
tomusdrw, line 88: This should be moved to `finally` block of the upper `try-catch` to make sure we set it to `false` in case of an error as well.

src/hooks/transactions/useTransactionsStatus.tsx
hbulgarini: > But you would need to do that for each transaction separately?

No. My idea is to not trigger the function `getTransactionStatus()` until a new subscription value comes in: 

```
const subscriptions = useSubscriptions();
const prevSubs = usePrevious(subscriptions);
...
useEffect( ()=>
...
const subChanged = subscriptions!== prevSubs;
if (!evaluatingTransactions && transactionsInProgress && subChanged) {
      try {
        getTransactionStatus();
      } catch (e) {
```

> Ideally we would like only the transaction that actually is affected to be re-calculated, not all of them.

As the evaluation process mainly resides on updated subscribed values i think there is no scenario when one transaction should be evaluated and the other does not.  So as soon as a new subscribed value arrives all transaction in progress status should be re evaluated.  The exception to this rule is when the transactions are completed , which is a case that is handled already.

Should we go with the subscriptions check? what do you think?

src/hooks/context/useTransactionsStatus.tsx
hbulgarini: I found out that moving the `setEvaluatingTransactionsStatus(false)` is incorrect because the  `getTransactionStatus` triggers an async function , but it is not waiting for the function to end , as a result `finally` block gets executed in the middle of the execution and mess up the process.
The solution to make sure that the evaluating will be managed correctly is by adding the `setEvaluatingTransactionsStatus(false)` call in the catch section.

By the way, the function `getTransactionStatus()` is called without the await because useEffect should not be async ( we discussed this before in previous PR)

src/hooks/transactions/useTransactionsStatus.tsx
hbulgarini: Well ... i have added the `const subChanged = subscriptions!== prevSubs;` and i have to say that is working great.
It reduces A LOT unnecessary the number of calls and it is only executed when new values are coming from subscriptions, which is an expected behavior.

src/hooks/context/useTransactionsStatus.tsx
tomusdrw: Then the whole `try-catch` is misleading and wrong then, because it does not handle all possible errors. We do need to handle all possible errors from async function, so please use `.catch` then, or wrap the whole content into async-aware `try-catch` (pretty sure we discussed error handling as well in previous PR).

src/hooks/context/useTransactionsStatus.tsx
```diff
@@ -0,0 +1,115 @@
...
+    const transactionsInProgress = transactions.find(({ status }) => status === TransactionStatusEnum.IN_PROGRESS);
+    const updatedSubscriptions = prevSubscriptions !== subscriptions;
+
+    if (!evaluatingTransactions && transactionsInProgress && updatedSubscriptions) {
+      try {
+        getTransactionStatus();
+      } catch (e) {
+        dispatchMessage(MessageActionsCreators.triggerErrorMessage({ message: e }));
+        dispatchTransaction(TransactionActionCreators.setEvaluatingTransactionsStatus(false));
+      }
```
tomusdrw, line 101: IMHO that would work really nicely, cause you separate the logic of `getTransactionStatus` from enacting the values to the reducer - i.e. there is a single co-located place where the updates happen, instead of having the scattered.

```suggestion
     dispatchTransaction(TransactionActionCreators.setEvaluatingTransactionsStatus(true));
     getTransactionStatus()
       .then(updatedTransactions => {
         if (!isEqual(transactions, updatedTransactions)) {
           dispatchTransaction(TransactionActionCreators.updateTransactionsStatus(updatedTransactions));
         }
         dispatchTransaction(TransactionActionCreators.setEvaluatingTransactionsStatus(false));
       })
       .catch( e => {
         dispatchMessage(MessageActionsCreators.triggerErrorMessage({ message: e }));
         dispatchTransaction(TransactionActionCreators.setEvaluatingTransactionsStatus(false));
       });
```

hbulgarini: i really like this approach. I will definitely implement it.

hbulgarini: @tomusdrw , Finally i didn't implement your proposal because i used the the new version of `genericCall` for managing the async call and reducer state management in a consistent and tested way. 
I think that closely we are reaching the final model.

src/reducers/tests/transactionReducerMocks.ts
```diff
@@ -32,6 +32,7 @@ export const state = {
   transactionDisplayPayload: {} as TransactionDisplayPayload,
   transactionRunning: false,
   transactionReadyToExecute: false,
+  evaludateTransactionStatusError: null,
```
tomusdrw, line 35:
```suggestion
  evaluateTransactionStatusError: null,
```

src/reducers/transactionReducer.ts
```diff
@@ -238,12 +238,20 @@ export default function transactionReducer(state: TransactionState, action: Tran
       return updateTransaction(state, action.payload);
     case TransactionActionTypes.SET_RECEIVER:
       return setReceiver(state, action.payload.receiverPayload);
-    case TransactionActionTypes.SET_TRANSACTION_RUNNING: {
+    case TransactionActionTypes.SET_TRANSACTION_RUNNING:
       return { ...state, transactionRunning: action.payload.transactionRunning };
+    case TransactionActionTypes.UPDATE_TRANSACTIONS_STATUS: {
+      const { evaludateTransactionStatusError, transactions, evaluatingTransactions } = action.payload;
```
tomusdrw, line 244:
```suggestion
      const { evaluateTransactionStatusError, transactions, evaluatingTransactions } = action.payload;
```

src/reducers/transactionReducer.ts
```diff
@@ -238,12 +238,20 @@ export default function transactionReducer(state: TransactionState, action: Tran
...
+      return {
+        ...state,
+        transactions: transactions && !evaludateTransactionStatusError ? transactions : state.transactions,
+        evaluatingTransactions,
+        evaludateTransactionStatusError
```
tomusdrw, line 249:
```suggestion
        evaluateTransactionStatusError
```

src/types/transactionTypes.ts
```diff
@@ -77,9 +78,11 @@ export interface TransactionState {
   derivedReceiverAccount: string | null;
   genericReceiverAccount: string | null;
   transactions: Array<TransactionStatusType>;
+  evaluatingTransactions: boolean;
   transactionDisplayPayload: TransactionDisplayPayload;
   transactionRunning: boolean;
   transactionReadyToExecute: boolean;
+  evaludateTransactionStatusError: string | null;
```
tomusdrw, line 85:
```suggestion
  evaluateTransactionStatusError: string | null;
```

tomusdrw (approved):

src/reducers/tests/transactionReducerMocks.ts
hbulgarini: "evaludate" :/

Issue #224, opened 2021-07-12
Refactor  selected sender account/companion account and their balance process.


hbulgarini:
At the moment , the logic for selected account/companion resides on a react hook at the component level. 

For sake of consistency based on the architecture agreed, all subscriptions should be handled in the data model for the accounts and update states changes through the account context.

In order to keep the changes progressively, this change will try to be split :

- [x] 1) Set companion account and subscribe to accounts balances in the data model.
- [x] 2) Replace `Account` component  in the Sender component for selected account/companion to use presentational component `AccountDisplay` ( currently Account is a wrapper that uses the useBalance hook which handles the subscription there ).
- [x]  3) Create a refresh process that every minute or 30 seconds reads the balances for all the accounts so the sender dropdown does not have to handle subscriptions. Also balances might be updated when the dropdown is opened.

hbulgarini:
Something that i don't have clear yet is if we should replace the balance process for the accounts listed in the dropdown. 
I think that subscribing all the existing accounts to their respective balance might be a bit overwhelming for the node.

We need to think in a way to move this logic to the data model as well but without affecting performance.

tomusdrw:
Balances of accounts in dropdown list are most likely good enough if they are just fetched periodically (say every minute) instead of being subscribed for. We can also do that only if the dropdown is actually open (i.e. after opening the dropdown there might be a small delay until you see balances). Only after the account is selected we should start a subscription to get balance updates as soon as possible.

Ideally (but that's really a stretch goal), we would send batch requests for accounts in groups of 10. (i.e. with 20 accounts we would just send two batch-RPC requests, each asking for 10 accounts - we don't even do that in parallel but rather in the order they appear on the dropdown).

hbulgarini:
> Balances of accounts in dropdown list are most likely good enough if they are just fetched periodically (say every minute) instead of being subscribed for. We can also do that only if the dropdown is actually open (i.e. after opening the dropdown there might be a small delay until you see balances). Only after the account is selected we should start a subscription to get balance updates as soon as possible.
> 
> Ideally (but that's really a stretch goal), we would send batch requests for accounts in groups of 10. (i.e. with 20 accounts we would just send two batch-RPC requests, each asking for 10 accounts - we don't even do that in parallel but rather in the order they appear on the dropdown).

[...]

PR #225, opened 2021-07-12
Add sender companion account and their balances to account's data model

hbulgarini:
First step for the work described in #224.

This PR is just creating the mechanism to add companion account derived process and the balance subscriptions to the data model only. In order to keep PR small , this PR is not including the actual use of this new states.

@goldsteinsveta  this change might simplify a little bit the logic for `Sender` component.
However i still don't have clear how to approach subscriptions for accounts listed in the dropdown. Being disscussed.


wirednkod (approved):
lgtm

src/hooks/accounts/useSourceAccountsBalances.ts
```diff
@@ -0,0 +1,53 @@
...
+    ss58Format: configs.ss58Format,
+    address: accountState.account?.address || '',
+    bridgeId: getBridgeId(configs, chainName)
+  };
+  const companionAccount = getDeriveAccount(toDerive);
+  const companionBalance = useBalance(targetApi, companionAccount || '', true);
+
+  useEffect(() => {
+    dispatchAccount(AccountActionCreators.setSenderBalances(accountBalance, companionBalance));
+    dispatchAccount(AccountActionCreators.setSenderCompanionAccount(companionAccount));
```
tomusdrw, line 49: This is always a smell to me. If things are always changed together it should be one reducer action.

tomusdrw: Alternatively maybe split this into two separate `useEffects`?

src/hooks/accounts/useSourceAccountsBalances.ts
```diff
@@ -0,0 +1,53 @@
...
+  } = useSourceTarget();
+
+  const accountBalance = useBalance(sourceApi, accountState.account?.address || '', true);
+
+  const toDerive = {
+    ss58Format: configs.ss58Format,
+    address: accountState.account?.address || '',
+    bridgeId: getBridgeId(configs, chainName)
+  };
+  const companionAccount = getDeriveAccount(toDerive);
```
tomusdrw, line 44: IMHO this should be part of the reducer logic and not the hook.

src/contexts/AccountContextProvider.tsx
```diff
@@ -46,19 +46,24 @@ export function useUpdateAccountContext() {
...
+  const [accountState, dispatchAccount] = useReducer(accountReducer, {
     account: null,
-    accounts: []
+    accounts: [],
+    companionAccount: null,
+    senderAccountBalance: null,
+    senderCompanionAccountBalance: null
   });
 
+  useSourceAccountsBalances(accountState, dispatchAccount);
```
tomusdrw, line 57: This is cool, but the only issue I have now is more about directory structure. I imagine that a generic `hooks` folder contains hooks that can be used in any place without any additional dependencies. Now, this particular hook can only be used in conjunction with a reducer and pretty much only in this one place - however by looking into `hooks` directory you won't be able to distinguish "regular hooks" and "helper/specialised hooks" - this long term will imho lead to confusion and "the code is complicated" feeling.

To overcome that we can either put such hooks in a different place, or have the hook `useReducer` internally. The latter is maybe a bit weird, cause if we happen to actually use that hook in other places than the context you will have a separate copy of the state, but well that might actually be a feature really.

src/util/getDeriveAccount.ts
```diff
@@ -26,6 +26,9 @@ interface Data {
 }
 
 export default function getDeriveAccount({ ss58Format = 42, bridgeId, address }: Data): string {
+  if (!address) {
+    return address;
+  }
```
tomusdrw, line 31: This is a bit weird, why special case empty string?

tomusdrw (approved):
lgtm! Some grumbles in the comments, but it's just old man's whining :)

src/hooks/accounts/useSourceAccountsBalances.ts
hbulgarini: no, this should be moved to reducer as you suggested.

src/hooks/accounts/useSourceAccountsBalances.ts
hbulgarini: totally, i think this came from moving the hooks.

src/contexts/AccountContextProvider.tsx
hbulgarini: it's a valid point which makes sense to me. I'm wondering if below the `hooks` directory, i could create a `context` sub folder and place this kind of hooks. Additionally i think it is interesting moving the `useReducer` to this specific dedicated hook.

src/util/getDeriveAccount.ts
hbulgarini: sorry , i'm not getting your question. Do you mean why i'm handling the empty address case?

hbulgarini:
> lgtm! Some grumbles in the comments, but it's just old man's whining :)

hell no!  it is great feedback that i will address right now thanks!

PR #226, opened 2021-07-13
Reading balances of selected sender account/companion account from context.

hbulgarini:
This PR is the second part of the data model explained in issue: #224.


src/components/Sender.tsx
```diff
@@ -125,6 +127,7 @@ const Sender = () => {
   const AccountSelected = () => {
     if (account) {
       const text = getName(account);
+      <AccountDisplay friendlyName={text} address={account.address} balance={senderAccountBalance?.formattedBalance} />;
```
wirednkod, line 130: Im pretty sure this is some kind of copy/paste issue? This code does not make any sense here

hbulgarini: ooouuhhh!!!

tomusdrw:
Please mark things that are not against `master` as draft, to avoid accidental merges.

src/components/Sender.tsx
```diff
@@ -149,12 +157,12 @@ const Sender = () => {
         {chains.map((chain) => renderAccounts(chain))}
       </Select>
       <div className={classes.accountCompanion}>
-        {derivedAccount ? (
-          <Account
+        {companionAccount && senderCompanionAccountBalance ? (
```
tomusdrw, line 160: `senderCompanionAccountBalance` can be `0`, but I guess since it's a `BN` it passes the check, right? That said I wonder if it wouldn't be better to explicitly compare the balance with `null`?

src/components/Sender.tsx
```diff
@@ -68,7 +69,8 @@ const useStyles = makeStyles((theme) => ({
 const Sender = () => {
   const classes = useStyles();
   const [chains, setChains] = useState<Array<string[]>>([]);
-  const { account, accounts, derivedAccount, setCurrentAccount } = useAccounts();
+  const { accounts, setCurrentAccount } = useAccounts();
+  const { account, companionAccount, senderAccountBalance, senderCompanionAccountBalance } = useAccountContext();
```
tomusdrw, line 73: It's a bit confusing to me to see two different hooks regarding the seemingly the same data. Like why do we `setCurrentAccount` using `useAccounts` but retrieve the `account` (assuming "current") with the other hook?

tomusdrw (approved):

src/components/Sender.tsx
hbulgarini: you are right saying that are two hooks, but the `useAccountContext` is just pulling the account context only. In our project all hooks having the convention: `use<concern>Context()` is just pulling the context . 
What it is confusing is that `accounts` are read from the `useAccounts` hook instead of the context. I have changed that.

src/components/Sender.tsx
hbulgarini: > if it wouldn't be better to explicitly compare the balance with null

sure!

PR #227, opened 2021-07-14
Move accounts to  accounts context

hbulgarini:
Previously, the accounts read from keyring context were processed in the hook at component level. This PR moves the logic to the accounts context.
As we are going to start reading the balances from the data model level , this change is necessary.

### DIFF
--- a/src/types/accountTypes.ts
+++ b/src/types/accountTypes.ts
@@ -15,21 +15,16 @@
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
 
 import type { KeyringPair } from '@polkadot/keyring/types';
-
 import { AccountActionsTypes } from '../actions/accountActions';
 
 export type Account = KeyringPair | null;
 
-export interface AccountContextType {
-  account: Account;
+export interface Payload {
+  [propName: string]: any;
 }
-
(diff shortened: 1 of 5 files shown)


src/types/accountTypes.ts
```diff
@@ -15,21 +15,16 @@
...
 import type { KeyringPair } from '@polkadot/keyring/types';
-
 import { AccountActionsTypes } from '../actions/accountActions';
 
 export type Account = KeyringPair | null;
 
-export interface AccountContextType {
-  account: Account;
+export interface Payload {
+  [propName: string]: any;
```
wirednkod, line 23: avoid using 'any'. Either add specific interface or go with unknown

wirednkod (changes requested):
lgtm (besides one minor comment).

Just a note, I think that the account/accounts can become confusing

hbulgarini:
> lgtm (besides one minor comment).
> 
> Just a note, I think that the account/accounts can become confusing

Understood. Any recommendations ?

src/types/accountTypes.ts
wirednkod: as agreed this is fine

wirednkod (approved):

PR #228, opened 2021-07-14
Alter eslintrc to check the REACT_APP_IS_DEVELOPMENT

### LINKED ISSUE #124: Improve development experience
Currently any formatting issue or unused variable causes a hard error which prevents the development version of UI to refresh.

This is extremely inefficient for regular exploration & development cycle.

These things should become `warnings` during development and hard errors only when building on CI.

wirednkod:
When REACT_APP_IS_DEVELOPMENT is set in true (in .env.local preferably) the CRA will not break if eslint rules are not folloed

Fixes #124;


hbulgarini (approved):
This looks good to me and is addressing the original issue.

tomusdrw (approved):

PR #229, opened 2021-07-15
Test CI with puppeteer

wirednkod:


.github/workflows/js.yml
```diff
@@ -27,7 +27,12 @@ jobs:
     - run: yarn
     - run: yarn downloadCustomTypes
     - run: yarn lint
+    # run normal tests
     - run: yarn test
+    # run ui puppeteer tests
+    - run: yarn test:e2e
+    # run end to end puppeteer test
+    - run: yarn test:e2e-alone
```
hbulgarini, line 35: i think this stand alone test is not necessary since `test:e2e` contains that test already.

hbulgarini (commented):
That's it? don't we need any container for running chromium or something like that? if it's so simple then let's do it!

wirednkod:
> I think we need to do more around this:
> 
> 1. just run the e2e test when is merged to master ( not on branch updates ).
> 2. i think something needs to be added so chromium is available during the execution. Check the GH action result:
> 
> ```
>     Failed to launch the browser process!
>     [1972:1972:0715/203636.071968:ERROR:browser_main_loop.cc(1393)] Unable to open X display.
> ```

This is not ready yet :) I'll add you as a reviewer when it is

tomusdrw:
@wirednkod does it work? Can we merge master here and review?

wirednkod:
> @wirednkod does it work? Can we merge master here and start the review?

No this is not working - I am trying to find why - but I have not so far (that is why it is still in draft)

hbulgarini (approved):
I think this PR is filling the needs pretty enough.

hbulgarini:
@wirednkod , i managed to make the e2e to work inside the GH action but i did a few changes to the deployment scope. 
The main one is that i have removed the matrix to work on node versions 14,15 and 16 due to a few reasons. First due to jobs run in parallel the build and deployment was executed three times (same with tests). So i decided to leave the same process to run under the v16 node version. 

What we would need to do is to do some corrective actions so this PR is not waiting for v14 and 15 to finish in order to get merged.

wirednkod:
> @wirednkod , i managed to make the e2e to work inside the GH action but i did a few changes to the deployment scope.
> The main one is that i have removed the matrix to work on node versions 14 and15 due to a few reasons. First because the jobs run in parallel the build and deployment was executed three times (same with tests). So i decided to leave the same process to run under the v16 node version.
> 
> What we would need to do is to do some corrective actions so this PR is not waiting for v14 and 15 to finish in order to get merged.

You should revert that. All flow should run in all matrix versions in order to ensure that the flow is working in all 3 versions (and yes build should take place 3 times in this case). The reason for having multiple versions is to ensure the support for active and current versions. If you really want to reduce build (which I do not understand the logic behind it) - you should remove v 15 (have a look [here](https://nodejs.org/en/about/releases/)) - but I would not remove 14 or 16.

hbulgarini:
> > @wirednkod , i managed to make the e2e to work inside the GH action but i did a few changes to the deployment scope.
> > The main one is that i have removed the matrix to work on node versions 14 and15 due to a few reasons. First because the jobs run in parallel the build and deployment was executed three times (same with tests). So i decided to leave the same process to run under the v16 node version.
> 
> You should revert that. All flow should run in all matrix versions in order to ensure that the flow is working in all 3 versions (and yes build should take place 3 times in this case). The reason for having multiple versions is to ensure the support for active and current versions. If you really want to reduce build (which I do not understand the logic behind it) - you should remove v 15 (have a look [here](https://nodejs.org/en/about/releases/)) - but I would not remove 14 or 16.
> 
> > What we would need to do is to do some corrective actions so this PR is not waiting for v14 and 15 to finish in order to get merged.
> 
> IMO we should wait for all CI steps to finish before merge.

I have reviewed your comment and i have added v14 and v15 for the test and lint steps only. Then e2e and build are only running for version 16 ( otherwise we would have unnecessary deploys ). The problem is that jobs `Build & Deploy (14.x)` & `Build & Deploy (15.x)` were removed but for some reason the PR expects to have the execution. @denisp-gh-admin , can you help us with this?

.github/workflows/publish-release-container.yml
```diff
@@ -0,0 +1,45 @@
+name: Build and Publish container for new releases
+
+on:
+  release:
+    types:
+      - published
```
TriplEight, line 6: this will happen only when you release it, is it all right?

.github/workflows/publish-release-container.yml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+
+    - uses:                       actions/checkout@v2
+
+    - name: Workaround rootless build
+    # temporary workaround for https://github.com/redhat-actions/buildah-build/issues/45
+      run: |
+        sudo apt-get install fuse-overlayfs
+        mkdir -vp ~/.config/containers
+        printf "[storage.options]\nmount_program=\"/usr/bin/fuse-overlayfs\"" > ~/.config/containers/storage.conf
```
TriplEight, line 21:
```suggestion
```
As I mentioned, please refer to a fixed job https://github.com/paritytech/substrate-api-sidecar/blob/master/.github/workflows/publish-release-container.yml

.github/workflows/publish-release-container.yml
```diff
@@ -0,0 +1,45 @@
...
+      with:
+        image:                    parity-bridges-ui
+        tags:                     ${{ github.event.release.tag_name }} latest
+        dockerfiles: |
+          ./Dockerfile
+        build-args: |
+          VERSION=${{ github.event.release.tag_name }}
+          VCS_REF=${{ github.ref }}
+          BUILD_DATE=${{ github.event.release.published_at }}
+    - name:                       Push image to docker.io
```
TriplEight, line 33:
```suggestion

    - name:                       Push image to docker.io
```
readability

.github/workflows/publish-release-container.yml
```diff
@@ -0,0 +1,45 @@
...
+      with:
+        registry:                 docker.io/paritytech
+        image:                    parity-bridges-ui
+        tags:                     ${{ github.event.release.tag_name }} latest
+        username:                 ${{ secrets.DOCKERHUB_USERNAME }}
+        password:                 ${{ secrets.DOCKERHUB_TOKEN }}
+
+    - name:                       Check the image
+      run: |
+        echo "New image has been pushed to ${{ steps.push-to-dockerhub.outputs.registry-path }}"
```
TriplEight, line 45:
```suggestion
        echo "New image has been pushed to ${{ steps.push-to-dockerhub.outputs.registry-path }}"

```
new line at the end of file

TriplEight (changes requested):
did you decide to build the image separately by now?

.github/workflows/js.yml
```diff
@@ -30,17 +28,34 @@ jobs:
...
-
+    - run: yarn build
+  integration:
+    name: Integration test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [16.x]
+    steps:
+    - name: Cancel Previous Runs
```
TriplEight, line 40:
```suggestion
    - name: Cancel Previous Runs

```
please add the spaces before the new job everywhere, it's easier to read

.github/workflows/js.yml
```diff
@@ -30,17 +28,34 @@ jobs:
...
+      uses: styfle/cancel-workflow-action@0.4.1
+      with:
+        access_token: ${{ github.token }}
+    - uses: actions/checkout@v2
```
TriplEight, line 44:
```suggestion

    - uses: actions/checkout@v2
```

.github/workflows/publish-release-container.yml
hbulgarini: yes, i think is the best way to go to publish the image when a new tag is created.

.github/workflows/publish-release-container.yml
hbulgarini: not sure i'm following this comment. What do you mean that i have to refer to a fixed job? Is this just a matter of remove that part of the code or should i move this code to the `js.yml`?

TriplEight: just apply the suggestion

TriplEight (approved):

PR #232, opened 2021-07-17
Create Api call for reading sender account balances.

hbulgarini:
At the moment when the sender dropdown is opened, the balances are read from the `useBalance` hook , opening a subscription per account.

This PR creates a specific call in the `useApiCall` that will preload the balances when application is starting and also they can be read as soon as the sender dropdown is opened. The result is stored in the reducer state called `displaySenderAccounts`.

Please take into account that at the moment the new values from the reducer state `displaySenderAccounts` are not being used since replacing these new values in the existing Sender component got a complexity that does not worth it since the whole component will be refactored.

This change is included in the issue: https://github.com/paritytech/parity-bridges-ui/issues/224


src/App.tsx
```diff
@@ -46,12 +46,12 @@ function App() {
 
   return (
     <SourceTargetContextProvider connections={connections}>
-      <ApiCallsContextProvider>
-        <SubscriptionsContextProvider>
-          <MessageContextProvider>
-            <SnackbarProvider>
-              <KeyringContextProvider>
-                <AccountContextProvider>
+      <KeyringContextProvider>
```
wirednkod, line 49: Is there a reason for this reorganising?

src/hooks/api/useApiCalls.ts
```diff
@@ -45,7 +56,76 @@ const useApiCalls = (): ApiCallsContextType => {
...
+          apiConnection: { api: sourceApi },
+          chain: sourceChain,
+          configs: sourceConfigs
+        } = sourceRole;
+        const {
+          apiConnection: { api: targetApi },
+          configs: targetConfigs
+        } = targetRole;
+
+        const accounts = await Promise.all(
```
wirednkod, line 88: Why you need an `await Promise.all` here?

src/types/accountTypes.ts
```diff
@@ -24,12 +25,36 @@ export interface Payload {
...
 }
 
+export type BalanceState = {
+  chainTokens: string;
+  formattedBalance: string;
+  free: Balance;
+};
+
+interface DisplayAccount {
+  account: {
```
wirednkod, line 35: make the address, balance, name a type in order to avoid repeating code

wirednkod (changes requested):

src/App.tsx
hbulgarini: Yes! the order matters because of the props chain through the high order components. So ApiCallsContextProvider needs to have the props ready from the KeyringContextProvider. Same with ApiCallsContextProvider -> AccountContextProvider

src/hooks/api/useApiCalls.ts
hbulgarini: because the iterator is returning an array of promises and what i want to is to fill the array `accounts` with the results. I have removed the `Promise.all` and i got the following typing issue:

```
 Type 'Promise<{ account: { address: string; balance: BalanceState; name: string; }; companionAccount: { address: string; balance: BalanceState; name: string; }; }>[]' is not assignable to type 'DisplayAccount[]'.
```

src/hooks/context/useAccountsContextSetUp.ts
```diff
@@ -51,8 +54,9 @@ const useAccountsContextSetUp = () => {
   useEffect(() => {
     if (keyringPairsReady && keyringPairs.length) {
       dispatchAccount(AccountActionCreators.setAccounts(keyringPairs));
+      updateSenderAccountsInformation(dispatchAccount);
```
tomusdrw, line 57: Seems it's never really refreshed after the initial call or am I missing something?

tomusdrw: I understand it's a work-in-progress but I'd appreciate if this was tracked some place so that we can figure out what is still missing and not to forget particular things.

hbulgarini: > Seems it's never really refreshed after the initial call or am I missing something?

Indeed. At the moment there is no refresh. The idea is that accounts are loaded at the really beginning so we have the values in the context in this initial phase of the refactor, so then we can focus on the best refresh process.

> I understand it's a work-in-progress but I'd appreciate if this was tracked some place so that we can figure out what is still missing and not to forget particular things.

OK. I will add a comment in the code.

tomusdrw: Awesome, thanks! I got confused because #224 was already closed, now it makes sense cause there is a list of tasks there.

wirednkod (approved):

src/hooks/context/useAccountsContextSetUp.ts
hbulgarini: I don't know how #224 got closed since it is clearly in progress. If i did it it was a mistake.

Issue #233, opened 2021-07-19
General display issues.


hbulgarini:
At the moment there a few display bugs / issues that degrades the UI quality:

1) Incorrect Estimated Fee when loading. After merging  #218 , when estimatedFee is re calculated it incorrectly shows: 

`Estimated Rialto fee ...3.706789324 MLAU`

when it should show:

`Estimated Rialto fee: ...`

2) UI jumping because of conditional renderings: 

- Warning label for incorrect amount.
- Estimated fee loading ( @goldsteinsveta : IMHO estimated fee should not be displayed if there is no estimated fee to show )

3) Autocomplete for inputs (should not show any recommendation)

PR #234, opened 2021-07-19
Solving display issues

hbulgarini:
Solves issues described in #233

### DIFF
--- a/src/components/EstimatedFee.tsx
+++ b/src/components/EstimatedFee.tsx
@@ -13,10 +13,10 @@
 //
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+
+import { useEffect, useState } from 'react';
 import { Typography, makeStyles } from '@material-ui/core';
 import React from 'react';
-import { useEffect } from 'react';
-import { useState } from 'react';
 import { useSourceTarget } from '../contexts/SourceTargetContextProvider';
 import { useTransactionContext } from '../contexts/TransactionContext';
 import { transformToBaseUnit } from '../util/evalUnits';
@@ -40,22 +40,12 @@ export const EstimatedFee = (): React.ReactElement => {
     !estimatedFeeLoading && setAmount(estimatedFee ? transformToBaseUnit(estimatedFee, srcChainDecimals) : null);
   }, [estimatedFee, estimatedFeeLoading, srcChainDecimals]);
 
-  if (!amount) {
-    return <div className={classes.container}></div>;
-  }
+  const feeLabel = `Estimated ${sourceChainDetails.chain} fee`;
 
   return (
     <div className={classes.container}>
       <Typography variant="body1" color="secondary">
-        Estimated {sourceChainDetails.chain} fee
-        {estimatedFeeLoading ? (
-          '...'
-        ) : (
-          <Typography variant="subtitle2" component="span">
-            {': '}
-            {amount} {chainTokens}
-          </Typography>
-        )}
+        {estimatedFeeLoading ? `${feeLabel}...` : amount ? `${feeLabel}: ${amount} ${chainTokens}` : null}
       </Typography>
     </div>
   );
--- a/src/components/EstimatedFee.tsx
+++ b/src/components/EstimatedFee.tsx
@@ -13,11 +13,13 @@
 //
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
-
-import React from 'react';
 import { Typography, makeStyles } from '@material-ui/core';
+import React from 'react';
+import { useEffect } from 'react';
+import { useState } from 'react';
 import { useSourceTarget } from '../contexts/SourceTargetContextProvider';
 import { useTransactionContext } from '../contexts/TransactionContext';
+import { transformToBaseUnit } from '../util/evalUnits';
 
 const useStyles = makeStyles(() => ({
   container: {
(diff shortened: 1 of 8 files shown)


src/reducers/transactionReducer.ts
```diff
@@ -160,16 +160,27 @@ const setReceiver = (state: TransactionState, payload: ReceiverPayload): Transac
...
+    case TransactionActionTypes.SET_ESTIMATED_FEE: {
+      const { estimatedFeeError, estimatedFeeLoading, srcChainDecimals } = action.payload;
+
+      let estimatedFee = null;
+      let estimatedFeeTransformed = null;
+
+      if (!estimatedFeeError && !estimatedFeeLoading && action.payload.estimatedFee) {
+        estimatedFee = action.payload.estimatedFee;
+        estimatedFeeTransformed = transformToBaseUnit(estimatedFee || '0', srcChainDecimals);
+      }
```
wirednkod, line 172:
```suggestion
      const { estimatedFeeError, estimatedFee, estimatedFeeLoading, srcChainDecimals } = action.payload;
      
      let estimatedFeeTransformed = null;

      if (!estimatedFeeError && !estimatedFeeLoading && estimatedFee) {
        estimatedFeeTransformed = transformToBaseUnit(estimatedFee || '0', srcChainDecimals);
      }
```

src/reducers/transactionReducer.ts
```diff
@@ -160,16 +160,27 @@ const setReceiver = (state: TransactionState, payload: ReceiverPayload): Transac
...
       return {
         ...state,
-        estimatedFee: action.payload.estimatedFeeError ? null : action.payload.estimatedFee,
-        estimatedFeeError: action.payload.estimatedFeeError,
-        estimatedFeeLoading: action.payload.estimatedFeeLoading,
-        transactionReadyToExecute: action.payload.estimatedFeeLoading
-          ? false
-          : isReadyToExecute(state) && action.payload.estimatedFee
+        estimatedFee,
+        estimatedFeeTransformed,
```
wirednkod, line 177: I do not think there is a need to carry across the dispatcher the estimatedFee transformed. As soon as there is the estimatedFee, it can be transformed wherever needed (and that is I think only in 1 place)

src/reducers/transactionReducer.ts
```diff
@@ -160,16 +160,27 @@ const setReceiver = (state: TransactionState, payload: ReceiverPayload): Transac
...
+        estimatedFeeError: estimatedFeeError,
+        estimatedFeeLoading: estimatedFeeLoading,
+        transactionReadyToExecute: estimatedFeeLoading ? false : isReadyToExecute(state) && estimatedFee
       };
+    }
+
     case TransactionActionTypes.SET_TRANSFER_AMOUNT: {
       const { transferAmount, chainDecimals } = action.payload;
```
wirednkod, line 185:
```suggestion
      const { transferAmount, chainDecimals, payload, payloadError } = action.payload;
```

src/reducers/transactionReducer.ts
```diff
@@ -186,7 +197,9 @@ export default function transactionReducer(state: TransactionState, action: Tran
       return {
         ...state,
         payload: action.payload.payloadError ? null : action.payload.payload,
```
wirednkod, line 199:
```suggestion
        payload: payloadError ? null : payload,
```

src/reducers/transactionReducer.ts
```diff
@@ -186,7 +197,9 @@ export default function transactionReducer(state: TransactionState, action: Tran
...
-        payloadError: action.payload.payloadError
+        payloadError: action.payload.payloadError,
```
wirednkod, line 200:
```suggestion
        payloadError,
```

src/components/EstimatedFee.tsx
```diff
@@ -13,33 +13,39 @@
...
 
   return (
-    <div>
+    <div className={classes.container}>
       <Typography variant="body1" color="secondary">
-        {`Estimated ${sourceChainDetails.chain} fee ${estimatedFeeLoading ? '...' : ':'}`}
-        <Typography variant="subtitle2" component="span">
-          {amount} {chainTokens}
-        </Typography>
+        Estimated {sourceChainDetails.chain} fee:{' '}
```
wirednkod, line 41: Im not sure why you revert [a proposal from Sveta](https://github.com/paritytech/parity-bridges-ui/pull/218#discussion_r671310165).

src/components/ReceiverInput.tsx
```diff
@@ -66,6 +66,7 @@ function ReceiverInput() {
           className={classes.address}
           fullWidth
           onChange={onReceiverChange}
+          autoComplete="off"
```
wirednkod, line 69: Its better to allo the user have autocomplete - this way if he uses a specific address again and again - he does not have to copy/paste non-stop

wirednkod (changes requested):

src/components/EstimatedFee.tsx
tomusdrw: My guess was that what @goldsteinsveta meant is:
```
"Estimated fee..." # when loading
```
and 
```
"Estimated fee: 5 RLT" # when ready
```

This one is a bit closer than previous, but we still have the extra double colon when loading. Also let's have Sveta confirm.

src/components/ReceiverInput.tsx
tomusdrw: IMHO this will be better addressed with a custom address book. Having a bunch of random (past; one-time) addresses being displayed isn't that helpful. We have identicons to also quickly verify the correctness of address without going character by character.

src/reducers/transactionReducer.ts
tomusdrw: Yeah, the `transformed` stuff seems purely like a display thing. Also "transformed" is super ambiguous it's not really clear what exactly is happening - seeing a variable like this in the data model seems like something that serves a purpose for exactly one specific component, hence it most likely does not really fit there.

src/components/ReceiverInput.tsx
hbulgarini: I agree with the address book. That would be the ideal solution, but for now i much prefer disabling the autocomplete because after a couple uses it gets really messy and annoying.

src/components/EstimatedFee.tsx
hbulgarini: > My guess was that what @goldsteinsveta meant is:
> 
> ```
> "Estimated fee..." # when loading
> ```
> 
> and
> 
> ```
> "Estimated fee: 5 RLT" # when ready
> ```
> 
> This one is a bit closer than previous, but we still have the extra double colon when loading. Also let's have Sveta confirm.

I changed to work in this way.

src/reducers/transactionReducer.ts
hbulgarini: Agree guys, it's too overwhelming for the reducer to keep this display value.

wirednkod (approved):

src/reducers/transactionReducer.ts
hbulgarini: addressed.

src/reducers/transactionReducer.ts
hbulgarini: addressed.

Issue #213, opened 2021-07-02
Internal Transfers


tomusdrw:
Currently the UI after selecting a Source Chain account allows you to transfer Target Chain tokens (e.g. Selecting Kusama account allows you to transfer DOT tokens).

Technically this is quite obvious, because that's exactly what the bridge is for. However from the user perspective it might be counter intuitive to not be able to do arbitrary transfers from the account they control.

We should consider extending UIs functionality to allow selecting any token (asset) that is held by the selected source account (on both networks).

Initially the focus should be just on native tokens (first instance of balances pallet). And after selecting the sender account, the user should be able to select the token they want to transfer.

Depending on the selected token (either Source Chain Token (SRC) or Target Chain Token (TAR)) the UI should compose a transaction that is:
1. In case of TAR - bridge transfer (call to bridge pallet on the source chain)
2. In case of SRC - internal transfer (call to balances pallet on the source chain)

This will *bridge* the functionality gap, and would allow the users to interact with the bridge seamlessly by simply performing transfers of some tokens that are in control of the selected account.

PR #236, opened 2021-07-20
Create API call for local transfers.

hbulgarini:
Related #213.

This PR includes the apiCall `localTransfer` implementation and it's corresponding hook `useLocalTransfer` which wraps all the necessary data for executing the local transfer.

As the idea is to submit incremental PR, the function is not being called at the moment , since we would need to have the transaction steps function to adapt the steps accordingly.

### DIFF
--- a/src/hooks/api/useApiCalls.ts
+++ b/src/hooks/api/useApiCalls.ts
@@ -19,8 +19,23 @@ import { Text, Bytes } from '@polkadot/types';
 import { Codec } from '@polkadot/types/types';
 import { ApiCallsContextType } from '../../types/apiCallsTypes';
 import useChainGetters from '../chain/useChainGetters';
+import { useSourceTarget } from '../../contexts/SourceTargetContextProvider';
+import { TransactionActionCreators } from '../../actions/transactionActions';
+import { web3FromSource } from '@polkadot/extension-dapp';
+import { KeyringPair } from '@polkadot/keyring/types';
+import { SignerOptions } from '@polkadot/api/types';
+import { TransactionDisplayPayload, TransactionStatusEnum, TransactionTypes } from '../../types/transactionTypes';
+import moment from 'moment';
+import { MessageActionsCreators } from '../../actions/messageActions';
+import logger from '../../util/logger';
 
 const useApiCalls = (): ApiCallsContextType => {
+  const {
+    sourceChainDetails: {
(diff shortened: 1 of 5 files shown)


src/hooks/api/useApiCalls.ts
```diff
@@ -45,7 +60,91 @@ const useApiCalls = (): ApiCallsContextType => {
...
+                      status: TransactionStatusEnum.COMPLETED
+                    },
+                    id
+                  )
+                );
+              })
+              .catch((e) => {
+                logger.error(e.message);
+                throw new Error('Issue reading block information.');
+              });
```
wirednkod, line 129: Since this is included in an async function - I would alter this to async/await

```suggestion
            try {
              const res = await sourceApi.rpc.chain.getBlock(status.asInBlock)
              const block = res.block.header.number.toString();
              dispatchTransaction(
                TransactionActionCreators.updateTransactionStatus(
                  {
                      block,
                      blockHash: status.asInBlock.toString(),
                      status: TransactionStatusEnum.COMPLETED
                    },
                    id
                 )
                );
              } catch (e) {
                logger.error(e.message);
                throw new Error('Issue reading block information.');
              };
```

hbulgarini: YES !

hbulgarini: YES!

src/hooks/api/useApiCalls.ts
```diff
@@ -45,7 +60,108 @@ const useApiCalls = (): ApiCallsContextType => {
...
+            );
+          }
+
+          if (status.isBroadcast) {
+            dispatchMessage(MessageActionsCreators.triggerInfoMessage({ message: 'Transaction was broadcasted' }));
+            dispatchTransaction(TransactionActionCreators.reset());
+          }
+
+          if (status.isInBlock) {
+            /*             sourceApi.rpc.chain
```
wirednkod, line 111: remove stray commented code

wirednkod (changes requested):

src/hooks/api/useApiCalls.ts
hbulgarini: thanks!

wirednkod (approved):

fig666 (commented):
what was the end result

PR #237, opened 2021-07-21
Moving action tab to GUI context

hbulgarini:
Moving the previous local state `items` &  `menuIdx` to GUI context.

The reason of this change is because when we evaluate the payload, estimatedFee we should consider some conditions met that really depends on the current action. For example, submit button should not be enabled until a receiver address is provided. In Remarks there is no receiver, so this is a conflict ( which actually is impacting the current master version ).

As all these conditions are being evaluated with reducers, having the current action in a local state makes things harder to handle ( we could set the action type as parameter on each reducer action, but is much easier to rely on the context)

### DIFF
--- a/src/components/MenuAction.tsx
+++ b/src/components/MenuAction.tsx
@@ -56,7 +56,7 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 interface MenuActionProps {
-  actions: Array<MenuActionItemsProps>;
+  actions: MenuActionItemsProps[];
   action: TransactionTypes;
   changeMenu: (type: TransactionTypes) => void;
 }
@@ -82,7 +82,7 @@ export const MenuAction = ({ actions, changeMenu, action }: MenuActionProps) => 
   return (
     <>
       <ButtonBase className={`${classes.item} current`} onClick={handleClick}>
-        {item!.title || '-'}
+        {item?.title || '-'}
         <ArrowDropDownIcon />
       </ButtonBase>
       <Popover
--- a/src/components/MenuAction.tsx
+++ b/src/components/MenuAction.tsx
@@ -17,7 +17,9 @@
 import { ButtonBase, Popover } from '@material-ui/core';
 import { makeStyles } from '@material-ui/core/styles';
 import ArrowDropDownIcon from '@material-ui/icons/ArrowDropDown';
-import React, { Dispatch, SetStateAction, useEffect } from 'react';
+import React, { useEffect } from 'react';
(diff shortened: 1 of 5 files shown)


src/components/MenuAction.tsx
```diff
@@ -80,10 +77,12 @@ export const MenuAction = ({ items, changeMenu, menuIdx }: MenuActionProps) => {
     setId(anchorEl ? 'simple-popover' : undefined);
   }, [anchorEl]);
 
+  const item = actions.find(({ type }) => type === action);
+
   return (
     <>
       <ButtonBase className={`${classes.item} current`} onClick={handleClick}>
-        {items[menuIdx]?.title || '-'}
+        {item!.title || '-'}
```
wirednkod, line 85: Why the exclamation mark? Why not question mark. Is `item = undefined` a case that should be allowed

src/components/MenuAction.tsx
```diff
@@ -53,18 +55,13 @@ const useStyles = makeStyles((theme) => ({
...
-interface MenuActionItemsProps {
-  idx: number;
-  title: string;
-  isEnabled: boolean;
-}
 interface MenuActionProps {
-  items: Array<MenuActionItemsProps>;
-  menuIdx: number;
-  changeMenu: Dispatch<SetStateAction<number>>;
+  actions: Array<MenuActionItemsProps>;
```
wirednkod, line 59: This is not wrong at all or a blocker. But in order to satisfy my OCD I believe this should be altered in order to be consistent with `src/contexts/GUIContextProvider.tsx` (line 27)

```suggestion
  actions: MenuActionItemsProps[];
```

src/hooks/transactions/useResetTransactionState.ts
```diff
@@ -19,15 +19,15 @@ import { useUpdateTransactionContext } from '../../contexts/TransactionContext';
 import { TransactionActionCreators } from '../../actions/transactionActions';
 import usePrevious from '../react/usePrevious';
 
-const useResetTransactionState = (type: string | number | undefined) => {
+const useResetTransactionState = (action: string | number | undefined) => {
```
wirednkod, line 22: I would rename this to something like `incAction` in order to avoid confusion with state `action`

hbulgarini: IMHO i think that it should remain as `action`, because this parameter called `action` is the `action` from the state and i think it makes sense to have the mental link about it.

wirednkod (approved):

Issue #205, opened 2021-06-29
Move isButtonDisable state to transaction reducer state.


hbulgarini:
The send message button is conducted by a `isButtonDisabled` that resides in the following code segment:

https://github.com/paritytech/parity-bridges-ui/blob/master/src/hooks/chain/useSendMessage.ts#L169

Since most of the state values resides in the transaction reducer , the state of this "ready" call should be moved to the reducer in order to keep consistency.

PR #238, opened 2021-07-22
Moving inputs for customCall and remark to transaction reducer.

hbulgarini:
Related #207, #205

This change is following the same convention applied for `Transaction` component related to debouncing the inputs and moving them to the reducer. This is needed in order to keep consistency across all components and general state management.

Also all the evaluations are happening now on the reducer having a smarter reducer.

This PR as it is, is not 100% working without the PR #239, so both PRs should be considered as one piece but for keeping PRs small and concise there were split.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -49,23 +49,23 @@ function App() {
       <KeyringContextProvider>
         <ApiCallsContextProvider>
           <AccountContextProvider>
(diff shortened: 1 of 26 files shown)


wirednkod (approved):
lgtm

hbulgarini:
won't do: to be implemented in #245

PR #239, opened 2021-07-22
Refactoring estimated Fee and payload process.

hbulgarini:
The current calculate payload and estimated fee hook/reducer model is a really pain and makes really difficult to manage the estimated fee component consistently. For example with the current model the calculation is triggered more than once in some specific scenarios. That's because of the async calls and the state management around payload and estimated fee calculations.

As the payload and estimated fee is a process quite related, the  solution proposed in this PR is  to unify the payload and estimate fee calculation in the same call, so we can : 

1) Avoid several checks about if payload is ready in order to calculate the estimated fee.
2) Avoid overlapped reducer actions for payload and estimated fee running and the same time and messing around with display components.
3) Duplicated code for both cases.

The downside is that if the estimated fee call fails for any reason, then the payload would not be updated to the reducer so the execution button would not be enabled. But in my opinion that's acceptable because is not right to let the user to run the transaction without being informed about the estimated fee. On the other side, if the payload process fails, then the estimated fee could not be calculated ( so in this case there is nothing to discuss )


hbulgarini:
@tomusdrw , this PR was opened as `Ready to review` because the PR #238 cannot be pushed to master without this PR , so both PR should be reviewed and approved. In case there is feedback on PR#238 , i will sync the changes .

Once both are approved the merge process will be:

1) Merge this PR into branch #238 
2) Merge PR #238 to master.

src/hooks/transactions/useEstimatedFeePayload.ts
```diff
@@ -51,23 +76,28 @@ export const useEstimateFee = () => {
...
-    dispatch
-  });
+      dispatchTransaction(TransactionActionCreators.setPayloadEstimatedFee(null, { estimatedFee, payload }, false));
+    };
 
-  return calculateEstimateFee;
+    try {
+      getPayloadEstimatedFee();
+    } catch (e) {
+      dispatchTransaction(
```
tomusdrw, line 85: incorrect because it's async.

tomusdrw (changes requested):
I don't really get the goal of the refactor and I feel we are going back in circles.

We've taken great care to implement `useDispatchGenericCall` and make sure that all possible async cases are handled correctly and now we revert that and introduce exactly the same bugs that we've been fighting with for such a long time.

tomusdrw:
> Once both are approved the merge process will be:
1. Merge this PR into branch Moving inputs for customCall and remark to transaction reducer. #238
2. Merge PR Moving inputs for customCall and remark to transaction reducer. #238 to master.

That's not correct order. PRs that are not against `master` should never be marked as ready to review. We should only merge PRs against master never branches-of-branches.
The reason is that after (1) from your list we would need to review yet again twice as large code containing both PRs and that's super hard.

hbulgarini:
@tomusdrw 
> I don't really get the goal of the refactor and I feel we are going back in circles.

> We've taken great care to implement useDispatchGenericCall and make sure that all possible async cases are handled correctly and now we revert that and introduce exactly the same bugs that we've been fighting with for such a long time.

To be honest i don't feel we are going back in circles. I saw a need for simplifying all the complexity that `useDispatchGenericCall` brought in terms of unnecessary rerenders , useCallback , returning functions etc. The purpose is well defined but at this point i'm not 100% convinced with the hook as it is.

Anyway i will keep your comments into account and see how can i change this process in order to accommodate it better and handle async calls properly.

hbulgarini:
> > Once both are approved the merge process will be:
> > 
> > 1. Merge this PR into branch Moving inputs for customCall and remark to transaction reducer. [Moving inputs for customCall and remark to transaction reducer. #238](https://github.com/paritytech/parity-bridges-ui/pull/238)
> > 2. Merge PR Moving inputs for customCall and remark to transaction reducer. [Moving inputs for customCall and remark to transaction reducer. #238](https://github.com/paritytech/parity-bridges-ui/pull/238) to master.
> 
> That's not correct order. PRs that are not against `master` should never be marked as ready to review. We should only merge PRs against master never branches-of-branches.
> The reason is that after (1) from your list we would need to review yet again twice as large code containing both PRs and that's super hard.

OK. I thought that being explicit in the PR description would solve the issue described but it seems it not. I will change this to draft until the other one gets merged.

hbulgarini:
won't do: to be implemented in #245

PR #240, opened 2021-07-23
Adding scheduler to update balances

hbulgarini:
Related #224 .

This PR schedules the balances update every certain amount of time.
This time can be configured by a `.env` variable

By the way, i didn't add any loader state variable on the state , because i think that balances updates should be transparent for the user ( similar as polkadot.js shows the balances update )


wirednkod (approved):
lgtm - just Add the information in the README.md on what that new variable is (and do)

hbulgarini:
> lgtm - just Add the information in the README.md on what that new variable is (and do)

THANKS!

.env
```diff
@@ -5,5 +5,6 @@ REACT_APP_CHAIN_2_CUSTOM_TYPES_URL=https://raw.githubusercontent.com/paritytech/
 REACT_APP_CHAIN_2_SUBSTRATE_PROVIDER=wss://wss.millau.brucke.link
 
 REACT_APP_LANE_ID=0x00000000
+REACT_APP_ACCOUNTS_INFORMATION_UPDATE_FREQUENCY=30000
```
tomusdrw, line 8: What's the point of making this configurable? This is just noise/boileplate imho. The UI should just work correctly without any need of fine-tuning.

src/hooks/context/useAccountsContextSetUp.ts
```diff
@@ -54,11 +56,15 @@ const useAccountsContextSetUp = () => {
...
     if (keyringPairsReady && keyringPairs.length) {
       dispatchAccount(AccountActionCreators.setAccounts(keyringPairs));
-      // This initial load might be temporary or it might change based on what it was disscussed in issue #224
       updateSenderAccountsInformation(dispatchAccount);
     }
   }, [keyringPairsReady, keyringPairs, dispatchAccount, updateSenderAccountsInformation]);
 
+  useEffect(() => {
+    const id = setInterval(() => updateSenderAccountsInformation(dispatchAccount), informationUpdateFrequency);
+    return () => clearInterval(id);
```
tomusdrw, line 65: This is not the best way to implement intervals cause it doesn't take into account delay caused by the function itself or handles errors. With low frequencies (or larger number of clients) that's basically a recipe to kill (ddos) your own server.

Imagine this:
1. frequency is low (say 3 seconds)
2. Response time is high (say 5 seconds)

This code will keep spamming node even if the node is not able to complete previous queries. With frequency higher than response rate it will just start eat up memory and eventually crash.

A better way to implement such refreshing is roughly like this:
```js
function updateMe(numberOfFailures) {
  updateSomethingAsync()
     .then(() => setTimeout(() => updateMe(0), DEFAULT_DELAY))
     .catch(e => {
       // since we failed, we are going to do exponential back-off to not overload the server.
       const delay = Math.min(MAX_DELAY, DEFAULT_DELAY * Math.pow(2, numberOfFailures));
       setTimeout(() => updateMe(numberOfFailures + 1), delay);
    });
}
```

tomusdrw (changes requested):
The frequency should not be configurable. Might be best to express it in number of blocks instead of absolute values too. For instance "update roughly every 2 blocks". The `setInterval` stuff could also be improved, but I can also live with the simpler version for now.

.env
hbulgarini: i think it is a nice feature to leave this configurable because hypothetically a certain team wants to implement their own pair UI, they will have the chance to customize this value as most of the other values we leave to configure from the .env.

src/hooks/context/useAccountsContextSetUp.ts
hbulgarini: i don't have anything to say about this.

wirednkod: I have - you need to make sure that you will clear the `setTimeout` during unmounting of the hook. 
Else you can have red screens and errors

hbulgarini: Yes, my submitted work handles the the `clearInterval` in the useEffect cleanup.

.env
tomusdrw: Honestly, I can't imagine any use case in which this would be useful to make this particular parameter configurable. It's basically like letting the "Submit" button to have customizable label via `.env` variables "because hypothetically...".
The choice to make this parameter customizable (and not others) is completely arbitrary and not based on any use cases we have considered. Either let's hardcode that, or if your intention is to go extra mile, let's make it based on the block time (i.e. refresh every four blocks or 30 seconds, whatever comes first).

hbulgarini: OK. We can definitely go for the four blocks approach if you like. But if we do this, my recommendation would be to remove the `setInterval` mechanism and just leave the blocks number count since i don't see much sense on keeping both. What do you think?

src/hooks/context/useAccountsContextSetUp.ts
hbulgarini: @tomusdrw if we implement the block number count + 30 seconds (whatever it comes first and after an update process, counters + timer gets reseted ), do we need to handle the catch error scenario? I understand your point and it is pretty valid, but i feel that is too much for what we are doing here, since if we need low frequency update we would just leave the balances update using subscriptions as it used to be before. The idea of this change is to update balances with  a larger amount of time ( around 4 blocks or 30 seconds ).

Btw ... do you want to establish these two numbers for now? 4 blocks  and 30 seconds?

.env
hbulgarini: >  (i.e. refresh every four blocks or 30 seconds, whatever comes first).

OK. Working on this.

src/hooks/context/useAccountsContextSetUp.ts
wirednkod: > Yes, my submitted work handles the the `clearInterval` in the useEffect cleanup.

Im not talking about the current code, but the code that Tomasz proposes.

src/hooks/context/useSendersBalancesContext.ts
```diff
@@ -0,0 +1,98 @@
...
+    dispatchAccount,
+    nextSourceBlockNumber,
+    nextTargetBlockNumber,
+    setNextSourceBlockNumber,
+    setNextTargetBlockNumber,
+    setTimerId,
+    timerId,
+    updateSenderAccountsInformation
+  ]);
+}
```
tomusdrw, line 98:
```suggestion  
  const updateAccounts = useCallback(() => {
        updateSenderAccountsInformation(dispatchAccount);
        setNextSourceBlockNumber(convertToBNAndIncrease(currentSourceBlock));
        setNextTargetBlockNumber(convertToBNAndIncrease(currentTargetBlock));
        // clear timer id if any
        if (timerId) {
          clearTimeout(timerId);
        }
        // set new timer 
        const newTimerId = setTimeout(updateAccounts, TIMER_DURATION);
        setTimerId(newTimerId);
  }, [
    dispatchAccount,
    currentSourceBlock,
    currentTargetBlock,
    setNextSourceBlockNumber,
    setNextTargetBlockNumber,
    setTimerId,
    timerId,
    updateSenderAccountsInformation
  ]);
  
  const blocksReached = getBlockAsBN(currentSourceBlock).gte(nextSourceBlockNumber)
    && getBlockAsBN(currentTargetBlock).gte(nextTargetBlockNumber);
    
  useEffect(() => {
    if (blocksReached) {
       updateAccounts();
    }
    // we need to make sure to cancel the timeout, this is enough to prevent setState after render,
    // so then we don't have to use `useMountedState` but rather can use the regular one.
    return () => {
      timerId && clearTimeout(timerId);
    };
  }, [blocksReached, updateAccounts, timerId]);
}
```

This is a bit shorter and imho much cleaner. By having only single state object for both `nextSource` and `nextTarget` block we could reduce the code (especially the deps) even more.

src/App.tsx
```diff
@@ -47,9 +47,9 @@ function App() {
   return (
     <SourceTargetContextProvider connections={connections}>
       <KeyringContextProvider>
-        <ApiCallsContextProvider>
-          <AccountContextProvider>
-            <SubscriptionsContextProvider>
+        <SubscriptionsContextProvider>
+          <ApiCallsContextProvider>
+            <AccountContextProvider>
               <MessageContextProvider>
```
tomusdrw, line 53: how is that related?

src/hooks/context/useSendersBalancesContext.ts
hbulgarini: The code looks very good, but it's not so accurate as my original code. i'm running this version and i'm logging every update with a timestamp , and i have verified that the updates are not happening with the expected frequency. In this case the updates should be every 24 seconds if there is no big delay ( block duration = 6 seconds * 4 ).

Also i did the test to reduce the timer to 10 seconds looking forward to get the update every 10 seconds and the update happens every 22 / 23 seconds.

src/App.tsx
hbulgarini: As we are using useSubscriptionsContext() inside the `AccountContextProvider` the prop chain inheritance needs to be changed moving the `SubscriptionsContextProvider` upper to the `AccountContextProvider`

src/hooks/context/useAccountsContextSetUp.ts
```diff
@@ -35,31 +35,23 @@ const useAccountsContextSetUp = () => {
...
   const companionBalance = useBalance(targetApi, accountState.companionAccount || '', true);
+  const prevAccountBalance = usePrevious(accountBalance);
+  const prevCompanionBalance = usePrevious(companionBalance);
 
   useEffect(() => {
-    dispatchAccount(AccountActionCreators.setSenderBalances(accountBalance, companionBalance));
-  }, [accountBalance, companionBalance, dispatchAccount]);
+    if (accountBalance !== prevAccountBalance || companionBalance !== prevCompanionBalance) {
+      dispatchAccount(AccountActionCreators.setSenderBalances(accountBalance, companionBalance));
+    }
```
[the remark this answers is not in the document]
hbulgarini: This is not necessary anymore. It was while i was working on the update process but after a couple of tweaks this is not needed anymore. Thanks for catching it.

src/hooks/context/useSendersBalancesContext.ts
tomusdrw: Aren't you curious WHY it does not work? Like if you agree that the code is "good" but you dismiss it because you can't figure out why it doesn't work as expected, then there is no point in merging this code.
Maybe try removing the cleanup function from `useEffect`?

The point of leaving code suggestions is always to challenge the existing solution and encourage you to go a bit of extra mile to figure out a better solution (cause you seem to agree that the current one is sub optimal). If you're not interested, I don't see much point of doing further reviews.

src/App.tsx
tomusdrw: Seems super fragile - I see basically every 2nd PR changing the order of these.

tomusdrw (approved):
🤷

src/hooks/context/useSendersBalancesContext.ts
hbulgarini: i was just saying pointing out that the code is not working as expected and i have never said that i'm not interested , actually i consider this as kind of work in progress to be honest, but i wanted to provide the feedback of what it is not working.

> Maybe try removing the cleanup function from useEffect?

I will try this out.

src/App.tsx
hbulgarini: As i'm using the hooks from the contexts and this is a refactor it is expected that this might change. At the moment is under control.

src/hooks/context/useSendersBalancesContext.ts
hbulgarini: Your code works great removing the cleanup function from useEffect as you proposed. My understanding is that if the timer is running and the component gets unmounted, the clean up will happen anyway because is part of the `updateAccounts` , so we would be covered on that. Anyway, the `useMountedState` will be still needed .

PR #241, opened 2021-07-26
Refactoring generic call

hbulgarini:
Based on previous discussions and a tendency of get rid of the hooks abuse, there is a need to simplify the generic call process that handles async calls properly covering all the cases and in case a dispatcher is provided, dispatches the corresponding actions under the `loading`,`error` & `data` values. 

Also i have added the `initalData` in case we need to set any specific value than `null`

### DIFF
--- a/src/util/apiUtlis.ts
+++ b/src/util/apiUtlis.ts
@@ -35,11 +35,12 @@ export const genericCall = async ({ call, emptyData = null, dispatch }: Dispatch
     executeDispatch(null, emptyData, true);
     data = await call();
     executeDispatch(null, data, false);
+    return { data, error };
   } catch (e) {
     error = e.message;
     executeDispatch(error, emptyData, false);
+    return { data: emptyData, error };
   }
-  return { data, error };
 };
 
 export default { genericCall };
--- a/src/util/apiUtlis.ts
+++ b/src/util/apiUtlis.ts
@@ -21,7 +21,7 @@ interface Result<T> {
   data: T | null;
 }
 interface DispatchGenericCallInput<T> {
-  call: (...args: any[]) => T;
+  call: (...args: any[]) => Promise<T>;
   dispatch?: (error: string | null, data: T | null, loading: boolean) => void;
   emptyData?: T | null;
 }
@@ -33,20 +33,19 @@ export const genericCall = async <T>({
 }: DispatchGenericCallInput<T>): Promise<Result<T>> => {
   let data = emptyData;
   let error = null;
-  const executeDispatch = (error: string | null, data: any, loading: boolean) =>
+  const executeDispatch = (error: string | null, data: T | null, loading: boolean) =>
     dispatch && dispatch(error, data, loading);
 
   try {
     executeDispatch(null, emptyData, true);
     data = await call();
-    executeDispatch(null, data, false);
-    return { data, error };
   } catch (e) {
     error = e.message;
     logger.error(error);
-    executeDispatch(error, emptyData, false);
-    return { data: emptyData, error };
+  } finally {
+    executeDispatch(error, data, false);
   }
+  return { data, error };
 };
 
 export default { genericCall };
--- a/src/util/apiUtlis.ts
+++ b/src/util/apiUtlis.ts
@@ -26,7 +26,7 @@ interface Result {
 }
 
 export const genericCall = async ({ call, emptyData = null, dispatch }: DispatchGenericCallInput): Promise<Result> => {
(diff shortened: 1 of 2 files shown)


src/util/apiUtlis.ts
```diff
@@ -0,0 +1,45 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+
+interface DispatchGenericCallInput {
+  call: Function;
+  dispatch?: (error: string | null, data: any, loading: boolean) => void;
+  emptyData?: unknown;
```
tomusdrw, line 20: This should be a generic parameter not `unknown`.

src/util/apiUtlis.ts
```diff
@@ -0,0 +1,45 @@
...
+// Parity Bridges UI is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
+
+interface DispatchGenericCallInput {
+  call: Function;
```
tomusdrw, line 18: I don't think it's that difficult to type that function properly.

src/util/apiUtlis.ts
```diff
@@ -0,0 +1,45 @@
...
+  let error = null;
+  const executeDispatch = (error: string | null, data: any, loading: boolean) =>
+    dispatch && dispatch(error, data, loading);
+
+  try {
+    executeDispatch(null, emptyData, true);
+    data = await call();
+    executeDispatch(null, data, false);
+  } catch (e) {
+    error = e.message;
```
tomusdrw, line 39: Might be worth to log the error?

src/util/apiUtlis.ts
```diff
@@ -0,0 +1,45 @@
...
+    executeDispatch(error, emptyData, false);
+  }
```
tomusdrw, line 41:
```suggestion
  try {
    executeDispatch(null, emptyData, true);
    data = await call();
  } catch (e) {
    error = e.message;
  } finally {
      executeDispatch(error, data, false);
  }
```

IMHO a bit cleaner, but maybe not as readable, so leaving here just as a suggestion.

src/util/apiUtlis.ts
```diff
@@ -0,0 +1,45 @@
...
+interface Result {
+  error: string | null;
+  data: unknown | null;
+}
+
+export const genericCall = async ({ call, emptyData = null, dispatch }: DispatchGenericCallInput): Promise<Result> => {
+  let data = emptyData;
+  let error = null;
+  const executeDispatch = (error: string | null, data: any, loading: boolean) =>
+    dispatch && dispatch(error, data, loading);
```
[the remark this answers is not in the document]
hbulgarini: exactly, the purpose of this function is to handle both cases: data that needs to be dispatched and data that should not. That's why the `dispatch` is optional.

src/util/apiUtlis.ts
hbulgarini: yes. let's do it... no cost and add value.

src/util/apiUtlis.ts
hbulgarini: i'll take it!

src/util/apiUtlis.ts
```diff
@@ -14,30 +14,36 @@
...
+import logger from '../util/logger';
 
-interface Result {
+interface Result<T> {
   error: string | null;
-  data: unknown | null;
+  data: T | null;
+}
+interface DispatchGenericCallInput<T> {
+  call: (...args: any[]) => Promise<T>;
```
tomusdrw, line 24:
```suggestion
  call: () => Promise<T>;
```

tomusdrw (approved):

src/util/apiUtlis.ts
hbulgarini: I wouldn't make the change ... If for whatever reason we want to provide a parameter to the useCallback function this will complain. My original code handles both cases.

tomusdrw: No it does not.

PR #248, opened 2021-07-30
Clear debounced text field once form is submitted

hbulgarini:
DebouncedTextField component is not getting its local value reseted once the reducer state gets reseted. This is expected because the value for the input should be independent  from the reducer value since the reducer one might contain different value as the one shown in the input component ( i.e `transferInput`).

This PR resets the input properly once the form gets executed cleaning everything. 
Also this PR solves some no desired validation messages on some inputs while the transaction is submited.

### DIFF
--- a/src/components/DebouncedTextField.tsx
+++ b/src/components/DebouncedTextField.tsx
@@ -33,7 +33,6 @@ interface Props {
   rows?: number;
   dispatchCallback?: (value: ValueType) => void;
   initialValue?: ValueType;
-  reset?: boolean;
 }
 
 export function DebouncedTextField({
@@ -48,10 +47,9 @@ export function DebouncedTextField({
   multiline,
   rows,
   dispatchCallback,
-  initialValue = '',
-  reset = false
+  initialValue = ''
 }: Props) {
-  const [value, setValue] = useDebounceState({ initialValue, dispatchCallback, reset });
+  const [value, setValue] = useDebounceState({ initialValue, dispatchCallback });
 
   return (
     <TextField
--- a/src/components/DebouncedTextField.tsx
+++ b/src/components/DebouncedTextField.tsx
@@ -33,6 +33,7 @@ interface Props {
   rows?: number;
   dispatchCallback?: (value: ValueType) => void;
   initialValue?: ValueType;
+  reset?: boolean;
 }
 
 export function DebouncedTextField({
@@ -47,9 +48,10 @@ export function DebouncedTextField({
   multiline,
   rows,
   dispatchCallback,
-  initialValue = ''
+  initialValue = '',
+  reset = false
 }: Props) {
-  const [value, setValue] = useDebounceState({ initialValue, dispatchCallback });
+  const [value, setValue] = useDebounceState({ initialValue, dispatchCallback, reset });
 
   return (
     <TextField
--- a/src/hooks/react/useDebounceState.ts
+++ b/src/hooks/react/useDebounceState.ts
@@ -59,7 +59,6 @@ export const useDebounceState = ({ initialValue, wait = 500, transformCallback, 
   }, [debounced, dispatchCallback, previousDebounced]);
 
   // Mechanism to reset local state input when the transaction is executed.
-  // In case no reset parameter is set to the hook, this process will not execute.
   useEffect(() => {
     if (transactions.length) {
       const latest = maxBy(transactions, 'id');
--- a/src/hooks/react/useDebounceState.ts
+++ b/src/hooks/react/useDebounceState.ts
@@ -26,18 +26,29 @@ interface Input {
   wait?: number;
   transformCallback?: (value: ValueType) => void;
   dispatchCallback?: (value: ValueType) => void;
+  reset?: boolean;
 }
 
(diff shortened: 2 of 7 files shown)


src/hooks/react/useDebounceState.ts
```diff
@@ -49,8 +60,28 @@ export const useDebounceState = ({ initialValue, wait = 500, transformCallback,
...
+    if (toReset && reset) {
+      setValue('');
+      setDebounced(null);
+      setToReset(false);
+    }
+
+    if (!shouldDispatch && !reset && previousReset !== reset) {
+      setShouldDispatch(true);
+    }
+  }, [reset, previousReset, setValue, value, debounced, toReset, shouldDispatch]);
```
wirednkod, line 84: I am not sure I understand the reasoning behind the existence of `toReset` and `previousReset`. I think the code can be much more simplified with the one below:
```suggestion
  useEffect(() => {
    if (reset) {
      if (value && debounced) {
        setShouldDispatch(false);
      }
      setValue('');
      setDebounced(null);
    }

    if (!shouldDispatch && !reset) {
      setShouldDispatch(true);
    }
  }, [reset, setValue, value, debounced, shouldDispatch]);
```

I am sure I am missing some reasoning here - but its a good conversation starter

wirednkod (changes requested):

src/hooks/react/useDebounceState.ts
tomusdrw: Yeah, I concur. The whole `setToReset` seems very unncessary. We should basically only check that `reset && !prevReset === true` and then run the clean up logic (i.e. the `reset` value changed from `false -> true`).

src/components/DebouncedTextField.tsx
```diff
@@ -33,6 +33,7 @@ interface Props {
   rows?: number;
   dispatchCallback?: (value: ValueType) => void;
   initialValue?: ValueType;
+  reset?: boolean;
```
tomusdrw, line 36: This smells of incorrect split of responsibilities. If we need to control internal state of the component from the outside it probably means that the state is not really that local.

I'd rather suggest to have `transactionId` being either passed as a prop or obtained from a reducer. Every time we send transaction `transactionId` is incremented, and all components are able to do any clean up they want if they notice that `transactionId` changes. On the first sight this looks much cleaner to me than the `reset/setToReset/toRest/previousReset` mumbojumbo happening internally now and is probably more future proof than passing `reset` to some specific components.

src/reducers/transactionReducer.ts
```diff
@@ -90,6 +97,15 @@ export default function transactionReducer(state: TransactionState, action: Tran
...
       const { customCallInput, createType, targetChain } = action.payload;
+      if (!customCallInput) {
+        return {
+          ...state,
+          customCallInput,
+          transactionReadyToExecute: false,
+          customCallError: null,
+          shouldEvaluatePayloadEstimatedFee: false
+        };
+      }
```
tomusdrw, line 108: How is this related?

src/reducers/transactionReducer.ts
```diff
@@ -64,6 +63,14 @@ export default function transactionReducer(state: TransactionState, action: Tran
...
     case TransactionActionTypes.SET_TRANSFER_AMOUNT: {
       const { transferAmount, chainDecimals } = action.payload;
+      if (!transferAmount) {
+        return {
+          ...state,
+          transferAmount,
+          transferAmountError: null,
+          transactionReadyToExecute: false
+        };
+      }
```
tomusdrw, line 73: How is this related?

src/hooks/react/useDebounceState.ts
hbulgarini: previous reset is not needed as we discussed. I have improved the code based on your suggestion.

src/components/DebouncedTextField.tsx
hbulgarini: shit, this is  such a great suggestion .... asking the transactionId to be different is such a greater event than transaction changed from false to true.

src/hooks/react/useDebounceState.ts
hbulgarini: guys i will resolve the conversation because i will base the process on listening for transaction.id changes, so this discussion gets outdated.

src/reducers/transactionReducer.ts
hbulgarini: When inputs are cleared ( because of this process or manual ) there is an error message which in my opinion is unnecessary.  The expected behavior is that when input gets cleared for whatever reason the execution button should not be available, the fee should not be calculated and no error message should appear on the input.

src/reducers/transactionReducer.ts
hbulgarini: https://github.com/paritytech/parity-bridges-ui/pull/248#discussion_r679939981

src/reducers/transactionReducer.ts
hbulgarini: I think it is related in the way that is part of the reset process which is the whole point of this PR, even if it's not 100% related to debounce text field component.

wirednkod (approved):

PR #249, opened 2021-08-02
Solving some conflicts with reset missing behaviors / chain switch.

hbulgarini:
This PR includes the following issues/ enhancements resolution:

1) Move the payload transaction display logic to the reducer so there are no leaks between renders.
2) Remove unnecessary reset in hooks.
3) Added a reset timestamp so components/hooks needed to reset as `useDebouncedState` know when to reset their local values ( this is a great and clear reset event )


src/reducers/transactionReducer.ts
```diff
@@ -15,8 +15,8 @@
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
 
 import type { InterfaceTypes } from '@polkadot/types/types';
+import moment from 'moment';
```
wirednkod, line 18: You should not use momentJs (https://momentjs.com/docs/#/-project-status/)

hbulgarini: how sad ... moment.js was so useful even though that mutable object stuff which tended to be a headache.

Anyway this can be easily replaced by `Date.now()` how i think about it.

wirednkod (approved):

PR #250, opened 2021-08-02
Adding validation to remark input

### LINKED ISSUE #247: Apply a validator for remark input.
At the moment the remark input does not have any value validator so for example something like 0x123 can be sent.

hbulgarini:
Closes #247


wirednkod (approved):
lgtm

PR #252, opened 2021-08-05
Fixing bug when switching chains

hbulgarini:
There was a bug where the account was not correctly set if the account selected was a target chain account.


wirednkod (approved):
lgtm

PR #257, opened 2021-08-12
Display message dispatch status in UI

### LINKED ISSUE #202: Display message dispatch status in UI
Currently after dispatching the message on target chain we deliver the confirmation with the result back to the source chain if it was dispatched successfully (see  https://github.com/paritytech/parity-bridges-common/issues/547) - for instance we send a transfer of TAR (target) tokens across the Source->Target bridge and we want to be sure that not only this message was delivered but also that this transfer actually succeeded on the target chain (for instance the account got created and not wiped out due to too low ED, etc). 

UI should display the dispatch result status as well. Currently we only display if the delivery was successful and there is no info about the actual **dispatch**.

UI can actually inspect the dispatch status directly on the target chain (i.e. it doesn't need to wait for that info to be delivered back to the source chain).
There are two possible ways how to do it:

1. Inspect the events at the exact block that the message was delivered and check this (could have been done even without #547)
2. Inspect the messages pallet storage to see the delivery status bit that was introduced in #547

CC @goldsteinsveta

hbulgarini:
Closes #202.

This new checks  inspects the delivery block on the target chain and looks for the event `MessageDispatched`. If found, then looks for the message nonce and the OK value returned.

### DIFF
--- a/src/util/transactions/index.ts
+++ b/src/util/transactions/index.ts
@@ -23,7 +23,8 @@ import {
   Payload,
   TransactionDisplayPayload,
   TransactionTypes,
-  TransactionState
+  TransactionState,
+  Step
 } from '../../types/transactionTypes';
 import shortenItem from '../shortenItem';
 import { Subscriptions } from '../../types/subscriptionsTypes';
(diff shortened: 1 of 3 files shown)


src/util/transactions/index.ts
```diff
@@ -161,6 +168,45 @@ interface InputTransactionUpdates {
...
+
+const checkMessageDispatchedEvent = async (
+  targetApi: ApiPromise,
+  blockNumber: string | null,
+  messageNonce: string | null
+) => {
+  if (!blockNumber || !messageNonce) {
+    return TransactionStatusEnum.IN_PROGRESS;
+  }
+  const blockHash = await targetApi.rpc.chain.getBlockHash(blockNumber);
```
wirednkod, line 193: Add try { } catch for errors

wirednkod (approved):

Issue #168, opened 2021-05-14
Refactor UI Select (Sender); replace MUI Select with lower level / custom components


goldsteinsveta:
MUI Select is not really suitable for intricately grouped options. That's why Sender component is a bit hacky now. IMO using a combination of lower / base components (ig [MUI Popover](https://material-ui.com/components/popover/)) should be the right way forward. New Select components could also be generic enough to also contain companion menus.

<img width="1053" alt="Screenshot 2021-05-14 at 18 22 54" src="https://user-images.githubusercontent.com/8782596/118299805-81028780-b4e1-11eb-8e0f-beabf05bde6d.png">

hbulgarini:
@goldsteinsveta , please take into account that PR #232 already sets the accounts with their companion and balances already in the context, avoiding the complex use of the `Account` / `AccountDisplay` components. With this change, we can use only one presentational component which can be `AccountDisplay`.

You can see the state variable here:

https://github.com/paritytech/parity-bridges-ui/pull/232/files#diff-c5386f21eabb4e08c56e9237bfb11f036a1beb47b363e2e78e9cd42359b42092R34

hbulgarini:
Hey @goldsteinsveta , i was wondering that instead of filtering chain by typing in the text field, we can add a tag functionality similar to this:
https://codesandbox.io/s/material-demo-forked-eejpw?file=/demo.js

what do you think?

tomusdrw:
Few comments on the existing component:
1. I think it should have a fixed size - currently, after filtering it might be reduced.
2. Network label is still displayed even if all accounts from that network are filtered out.
3. Would be cool if down/up arrows worked and it was possible to select an account by using down arrow + enter sequence.
4. The selector should probably be fixed on top, and only the account list scrollable (also would be good to display a tiny scrollbar)
5. Would be cool to add a bit of spacing between name and the balance box - it's only an issue for the longest account name on the list.
6. CC @goldsteinsveta I'd consider adding the identicon of companion account as well (maybe a tiny one near balance?) - cause sometimes it might be easier to correlate using this identicon.

hbulgarini:
Points 1 - 2- 5 - 6: OK. No comments from my side.

> 3. Would be cool if down/up arrows worked and it was possible to select an account by using down arrow + enter sequence.

[...]

PR #263, opened 2021-08-21
Refactor sender component based on popover component.

### LINKED ISSUE #103: Sender select component does not let to copy the source account from the identicon

hbulgarini:
According to the task described in the task #168, there is a need of refactoring the Sender component at moment mounted over the Material UI (MUI) `Select` component which brings a lot of limitations.

I tried first replacing the Select component of MUI by the Autocomplete, but even though a lot of hacky sections were solved, i still faced limitations in terms of all the customization and features we aim to implement in the dropdown body.
Based on this, i chose to build the component based on a `Popover` component ( as proposed in the #168 ) with low level features that replicate what it comes out of the box with `Select` or `Autocomplete` component.

Below is the component's hiercharchy in order to make the PR legibility more clear:

- Sender.tsx
    -- SenderAccount.tsx (shows the current account selected and hooks the anchor element for the popover)
    
    -- SenderDropdown.tsx (Contains the popover)
         ---  SenderFilters.tsx ( contains text field to filter, and switch components to show/hide )
         ---  SenderAccountsSection.tsx ( builds the section where all accounts per chain will be displayed )
               ---- SenderAccountsListOnChain.tsx ( loop the array of accounts)
                       ----- SenderDropdownItem.tsx ( equivalent to AccountDisplay but without so much logic )
    -- SenderCompanionAccount.tsx (shows the equivalent selected companion account)

Even it does not look exactly the same as the design, it is much closer to the original idea than the one it is now. Below some screenshots

Selected Account collapsed:
Screen Shot 2021-08-21 at 17 52 53

Dropdown opened:
Screen Shot 2021-08-21 at 17 53 09

With text filter:
Screen Shot 2021-08-21 at 17 53 32

With hide companion account filter on:
Screen Shot 2021-08-21 at 17 53 45

Also Closes #103


src/components/SenderDropdownItem.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+interface Props {
+  name: string;
+  balance: string;
+  companionBalance: string;
+  address: string;
+  showCompanion: boolean;
+}
+
+const useStyles = makeStyles((theme) => ({
```
hbulgarini, line 32: @goldsteinsveta i wonder if there is a better way to define styles and `Box` component hierarchy

hbulgarini:
CC: @tomusdrw

hbulgarini:
Update: Maybe really small CSS tweaks are needed to have pixel perfect implementation, but the component looks the almost the same as the original design.

PR #272, opened 2021-09-01
Fixing Dockerfile definition.

hbulgarini:
The docker image builder is failing with the following errors:

https://github.com/paritytech/parity-bridges-ui/runs/3485843336?check_suite_focus=true

In order to fix it i have change the order of the copy of the files and i added python package so errors are gone.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,9 +12,13 @@ ENV PYTHONUNBUFFERED=1
 RUN python3 -m ensurepip
 RUN pip3 install --no-cache --upgrade pip setuptools
 
-COPY . .
+# install app dependencies
+COPY package.json .
+COPY yarn.lock .
 RUN yarn
+COPY . .
 RUN yarn build
+
 
 # nginx state for serving content
 FROM nginx:alpine
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,6 @@ RUN pip3 install --no-cache --upgrade pip setuptools
 
 COPY . .
 RUN yarn
-RUN yarn downloadCustomTypes
 RUN yarn build
 
 # nginx state for serving content
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,11 @@ WORKDIR /parity-bridges-ui
 
 RUN apk update
 RUN apk add git
+# Install python/pip
+ENV PYTHONUNBUFFERED=1
+RUN apk add --update --no-cache python3 && ln -sf python3 /usr/bin/python
+RUN python3 -m ensurepip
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -6,12 +6,15 @@ WORKDIR /parity-bridges-ui
...
+ENV PYTHONUNBUFFERED=1
+RUN apk add --update --no-cache python3 && ln -sf python3 /usr/bin/python
+RUN python3 -m ensurepip
+RUN pip3 install --no-cache --upgrade pip setuptools
 
-# install app dependencies
-COPY package.json .
-COPY yarn.lock .
-RUN yarn
 COPY . .
```
tomusdrw, line 15: The `COPY` stuff should stay as-is, this is a common practice to utilize docker cache

Dockerfile
```diff
@@ -6,12 +6,15 @@ WORKDIR /parity-bridges-ui
 
 RUN apk update
 RUN apk add git
+# Install python/pip
+ENV PYTHONUNBUFFERED=1
+RUN apk add --update --no-cache python3 && ln -sf python3 /usr/bin/python
```
tomusdrw, line 11:
```suggestion
# Install git & python/pip
RUN apk add git python3
RUN ln -sf python3 /usr/bin/python
ENV PYTHONUNBUFFERED=1
```

Dockerfile
```diff
@@ -6,12 +6,15 @@ WORKDIR /parity-bridges-ui
...
+RUN python3 -m ensurepip
+RUN pip3 install --no-cache --upgrade pip setuptools
 
-# install app dependencies
-COPY package.json .
-COPY yarn.lock .
-RUN yarn
 COPY . .
+RUN yarn
+RUN yarn downloadCustomTypes
```
tomusdrw, line 17: Isn't this part of pre-build or post-install?

hbulgarini: pre build, it downloads the custom types files prior to build the image.

tomusdrw: right, so this line seems unneeded then :)

Dockerfile
hbulgarini: the problem i'm facing is that i need the path `src/configs` for making the `src/configs/chainsSetup/downloadCustomTypes.js` available, so i need more than the package.json and yarn.lock.

What's the best way to achieve this? Just copy the `src/configs` ? (i'm not an expert on this honestly)

hbulgarini: ok! reverted ... the prebuild script solved the issue i had.

tomusdrw (approved):

Issue #270, opened 2021-09-01
Apply behaviors and styles on Sender component.


hbulgarini:
- [ ] I think it should have a fixed size - currently, after filtering it might be reduced.

- [ ] Network label is still displayed even if all accounts from that network are filtered out.

- [ ] Would be cool if down/up arrows worked and it was possible to select an account by using down arrow + enter sequence.

- [ ] The selector should probably be fixed on top, and only the account list scrollable (also would be good to display a tiny scrollbar)

- [ ] Would be cool to add a bit of spacing between name and the balance box - it's only an issue for the longest account name on the list.

- [ ] Add the identicon of companion account as well - cause sometimes it might be easier to correlate using this identicon.

hbulgarini:
related #168

hbulgarini:
@tomusdrw : ` Add the identicon of companion account as well - cause sometimes it might be easier to correlate using this identicon.`

As far is understood we replaced this by the popover right? If so, can we set this as completed?

PR #274, opened 2021-09-07
Fixing some styles around Sender component.

hbulgarini:
Related with #270 

Items performed:

- have a fixed size - currently, after filtering it might be reduced.
- Network label is still displayed even if all accounts from that network are filtered out.
-The selector is now fixed on top, and only the account list scrollable (also would be good to display a tiny scrollbar)


hbulgarini:
cc: @tomusdrw

Issue #276, opened 2021-09-08
Few UI issues


tomusdrw:
1. [ ] Cancelling signing transaction in the extension doesn't enable the "Send" button back (same thing happens in case of any error, one need to change something to force re-calculation)
2. [ ] Race condition when typing amounts (can still press the button before the re-calculation kicks in).
4. [ ] Current transfer type can be de-selected (the UI goes into some weird state after that).
5. [ ] "Decoded" tab - copy icon jumps around (content is moving up/down - my suggestion would be to disable this completely and change the cursor to "arrow-plus").
6. [ ] Width of sender selector is changing (Marked as addressed in #270, but seems to still be there)
7. [ ] Internal transfer - receipt is incorrect (shows both networks - I guess it should be hidden anyway?).
8. [ ] After sending transaction, different receipt is displayed below and it initially seems to be missing `payload`.
9. [ ] When selecting the same account again - balance disappears.
10. [ ] Millau internal transfer doesn't work ("Transaction has a bad signature")

tomusdrw:
Just noticed this quirk as well:

![Screenshot from 2021-09-08 13-45-18](https://user-images.githubusercontent.com/581548/132503615-bf1da136-5365-468a-bac7-9ccb863f202c.png)

The `->` symbol is added and removed when switching between "Internal" and "Bridge" modes. It's visible after sending one internal transfer.

hbulgarini:
> Just noticed this quirk as well:
> 
> ![Screenshot from 2021-09-08 13-45-18](https://user-images.githubusercontent.com/581548/132503615-bf1da136-5365-468a-bac7-9ccb863f202c.png)
> 
> The `->` symbol is added and removed when switching between "Internal" and "Bridge" modes. It's visible after sending one internal transfer.

Niche catch. Addressed on #283

hbulgarini:
@tomusdrw  `Millau internal transfer doesn't work ("Transaction has a bad signature")`

we agreed to not work on this correct?

tomusdrw:
> we agreed to not work on this correct?

Correct, it's an issue with the Chrome Extension not being able to use a custom hasher, so it's not something we could easily address.

PR #282, opened 2021-09-12
Fixing race condition and amount evaluation

hbulgarini:
Related #276 .

> Race condition when typing amounts (can still press the button before the re-calculation kicks in).

Indeed. As the input is debounced , the disable event was triggered once the delay is met.
Also there was an undesired behavior that during a calculation, if the amount contained an incorrect character the submit button was still enabled.


hbulgarini:
cc @tomusdrw

PR #284, opened 2021-09-13
Add local image and don't crash when logo is missing.

tomusdrw:
I'm trying to finish up https://github.com/paritytech/parity-bridges-common/pull/1047 and currently the docker-compose networks are running with "Local Network" chain name. 
Currently this crashes the UI, cause the module with logo cannot be loaded.

This PR adds an icon for `local.png` and also uses that icon as a fallback. (currently just copy-pasted Rialto's icon).


hbulgarini (approved):
Great catch.

hbulgarini:
CC @tomusdrw : merging it.

PR #286, opened 2021-09-17
Fixing width for Sender Component.

hbulgarini:
Related to #276 #270 

This PR solves some design issues related with sender component size when account names are larger:

1) Balance has a margin left now.
2) The account name has an ellipsis that wraps the text and does not increase the size.
3) When the focus is on the address item, there is a tooltip that shows native/companion addresses:

screenshots below:

Screen Shot 2021-09-17 at 10 53 37

Screen Shot 2021-09-17 at 10 53 48

Screen Shot 2021-09-17 at 10 53 59


hbulgarini:
@tomusdrw , i set yourself as reviewer in case you want to review the design.

tomusdrw (approved):

PR #289, opened 2021-09-17
Removing jumping behavior on decoded tab.

hbulgarini:
Related #276 

> "Decoded" tab - copy icon jumps around (content is moving up/down - my suggestion would be to disable this completely and change the cursor to "arrow-plus").

This PR changes the clipboard on the payload tab. As this component relies completely on `ReactJson` library i'm afraid that manipulating the icon and the jumping behavior is not so direct.

I managed to disable the copy clipboard functionality so the jumping issue is gone. 

@tomusdrw i'm not sure if what you mean by "change the cursor to "arrow-plus" , but i have added the following cursor in case this is what you wanted (if not i can simply remove it)

Screen Shot 2021-09-17 at 17 19 51

### DIFF
--- a/src/components/TransactionPayload.tsx
+++ b/src/components/TransactionPayload.tsx
@@ -14,12 +14,14 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges UI.  If not, see <http://www.gnu.org/licenses/>.
 
-import React from 'react';
+import React, { useCallback } from 'react';
 import { Card, makeStyles, CircularProgress, Typography } from '@material-ui/core';
 import ReactJson from 'react-json-view';
 import TransactionHeader from './TransactionHeader';
+import { MessageActionsCreators } from '../actions/messageActions';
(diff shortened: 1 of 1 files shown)


tomusdrw:
Does it mean it's still possible to copy? If clicking does not copy, then the cursor should just stay as a regular arrow. I was thinking of making the behavior similar to copying the address from identicon (Hover -> Cursor changes -> Click to copy).

hbulgarini:
> Does it mean it's still possible to copy? If clicking does not copy, then the cursor should just stay as a regular arrow. I was thinking of making the behavior similar to copying the address from identicon (Hover -> Cursor changes -> Click to copy).

OK. It was not hard to add the custom function to manipulate the clipboard when the area is clicked. The only issue i was not able to manage is that the copy cursor is replaced by an arrow ( this is internal from the library ), but the copy functionality still works.  I think that this minor issue is something we can live . What do you think?

PR #296, opened 2021-10-20
Pay dispatch fee on the target chain

### LINKED ISSUE #266: Pay dispatch fees on target chain
Details: https://github.com/paritytech/parity-bridges-common/issues/561

hbulgarini:
Closes #266 .

This PR adds the code base to support paying the dispatch fee on the target chain. As the UX around this switch is not clear yet, the default still remains on the source chain until we define defaults + UX for switching.

Also this PR shows the fee involved in both chains and it solves a missing fee process on the source chain.

### DIFF
--- a/src/components/EstimatedFee.tsx
+++ b/src/components/EstimatedFee.tsx
@@ -20,9 +20,8 @@ import { fade, makeStyles } from '@material-ui/core/styles';
 import HelpOutlineIcon from '@material-ui/icons/HelpOutline';
 import { useSourceTarget } from '../contexts/SourceTargetContextProvider';
 import { useTransactionContext } from '../contexts/TransactionContext';
-import { transformToBaseUnit } from '../util/evalUnits';
-import round from 'lodash/round';
 import { Alert } from '.';
+import { formatBalance } from '@polkadot/util';
 
 const useStyles = makeStyles((theme) => ({
   container: {
@@ -39,8 +38,14 @@ const useStyles = makeStyles((theme) => ({
   }
 }));
 
-const getAmount = (fee: string | null, chainDecimals: number) =>
-  fee ? round(parseFloat(transformToBaseUnit(fee, chainDecimals)), 2) : null;
+const getFormattedAmount = (fee: string | null, chainDecimals: number, chainTokens: string) =>
+  fee
+    ? formatBalance(fee, {
+        decimals: chainDecimals,
+        withUnit: chainTokens,
+        withSi: true
+      })
+    : null;
 
 export const EstimatedFee = (): React.ReactElement => {
   const classes = useStyles();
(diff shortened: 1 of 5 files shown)


src/components/EstimatedFee.tsx
```diff
@@ -14,51 +14,104 @@
...
     transactionRunning,
     evaluateTransactionStatusError
   } = useTransactionContext();
   const srcChainDecimals = sourceChainDetails.apiConnection.api.registry.chainDecimals[0];
-  const { chainTokens } = sourceChainDetails.apiConnection.api.registry;
+  const tarChainDecimals = targetChainDetails.apiConnection.api.registry.chainDecimals[0];
+
+  const { chainTokens: srcChainTokens } = sourceChainDetails.apiConnection.api.registry;
+  const { chainTokens: tarChainTokens } = targetChainDetails.apiConnection.api.registry;
+  /*
```
tomusdrw, line 62: comments

src/reducers/tests/transactionReducer.test.ts
```diff
@@ -189,13 +195,19 @@ describe('transactionReducer', () => {
         payloadEstimatedFeeLoading,
         {} as SourceTargetState,
         () => 'type',
-        true
+        true,
+        null,
+        null,
+        9
```
tomusdrw, line 201: With so many parameters and some of them being optional I'd consider a single configuration object to be passed as the only argument to `setPayloadEstimatedFee` instead of these 9 positional arguments.

src/util/transactions/index.ts
```diff
@@ -111,16 +113,17 @@ export async function getTransactionCallWeight({
       case TransactionTypes.REMARK:
         call = (await targetApi.tx.system.remark(remarkInput)).toU8a();
         // TODO [#121] Figure out what the extra bytes are about
-        call = call.slice(2);
+        //  call = call.slice(2);
```
tomusdrw, line 116: It isn't needed?

src/components/EstimatedFee.tsx
```diff
@@ -14,51 +14,104 @@
...
+    marginTop: 2,
+    marginLeft: 2,
+    '&:not(:hover)': {
+      color: fade(theme.palette.text.hint, 0.75)
+    }
   }
 }));
 
+const getAmount = (fee: string | null, chainDecimals: number) =>
+  fee ? round(parseFloat(transformToBaseUnit(fee, chainDecimals)), 2) : null;
```
tomusdrw, line 43: Wouldn't it be better to have it formatted by `polkadot.js/api` function with all of the SI units?

tomusdrw (approved):
lgtm, modulo the comments

src/reducers/tests/transactionReducer.test.ts
hbulgarini: yeah, completely agree. I think it started with just a few and it kept growing.

src/util/transactions/index.ts
hbulgarini: btw, the last two byes that we slice exist because scale adds it. Not sure why, but we have detected that with other js developers. At the moment i will recommend to keep the `slice`

src/components/EstimatedFee.tsx
hbulgarini: I think it looks much better now:

Screen Shot 2021-10-21 at 10 58 01

PR #302, opened 2021-11-15
Fix bug with dispatch fee selector for local transfers.

hbulgarini:
Related #266 .

There was a bug introduced in the PR #300 where the fee for the local transaction were not being shown.

This PR solves that issue and shows the fee correctly on both cases now.
Also old references to `estimatedFee` variable were cleaned.


hbulgarini:
CC: @tomusdrw

Issue #299, opened 2021-11-03
Design for TransactionStatus cards, mock and submitted


goldsteinsveta:
_This is a lighter alternative to solution in  #108, closes #108_ 

Here's a desired flow:

1. Local storage menu and mock TransactionStatus are moved to the top right corner.
2. Transaction form – all necessary fields are filled, a valid transaction is composed, a mock TransactionStatus appears
3. Transaction form after hitting submit – is locked; signing extension – expecting signature
4. After a signed transaction is submitted:
    1. The form is cleared -> not enough data to construct a TransactionStatus mock -> "mock disappears"
    2. My Transaction menu opens. TransactionStatus of just submitted transaction (is not a mock anymore) is in the list of Ongoing Transactions. After the transaction is delivered it is in the Past Transactions list of My Transactions

<img width="1440" alt="01 Balance Transfer" src="https://user-images.githubusercontent.com/8782596/140059438-ca620716-4a50-4790-81b9-2b09bfb6c16c.png">
<img width="1440" alt="02 Balance Transfer" src="https://user-images.githubusercontent.com/8782596/140059451-4ee9d9aa-c9b0-401d-9968-ae81fe864b23.png">
<img width="1440" alt="03 Balance Transfer" src="https://user-images.githubusercontent.com/8782596/140059454-98f81637-599e-43b2-b245-9887f34c4a53.png">
<img width="1440" alt="04 Balance Transfer" src="https://user-images.githubusercontent.com/8782596/140059458-77b00bd5-3a6d-45de-bd22-d82722767f29.png">

hbulgarini:
Hey @goldsteinsveta ! I love the design and i think i will take this solution as the final one for this matter.

I have one question only: What happens after the `Transaction Submitted` confirmation? How will the user be able to come back to the form?

goldsteinsveta:
I think it should appear for only a second (I've edited the issue, added this steps)

PR #303, opened 2021-11-22
New TransactionStatus cards, mock and submitted behaviors.

hbulgarini:
Related #299 .

This PR implements most as possible the design and UX for the transactions card.

Also this PR brings a fix that when the extension confirmation pops up for signing the transaction, the form gets locked.


hbulgarini:
@goldsteinsveta @tomusdrw @dnlklmn merging this PR so you can check the new transaction cards implementation.

